PR #31, opened 2025-08-25
Bugfix/grad accum

zanghz21:
### Description

- Fix gradient accumulation bug
- Fix train logging
- Fix maniskill-ppo-oft config

### Motivation and Context

- Error gradient accumulation value
- Previously only log the rank 0 training metrics. 
- Previous maniskill-ppo-oft miss essential attributes. 

### How has this been tested?

- run embodiment start script

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/algorithms/embodiment/grpo_functions.py
+++ b/rlinf/algorithms/embodiment/grpo_functions.py
@@ -277,12 +277,16 @@ def actor_loss_fn(
 
     if loss_mask is not None:
         # Take the maximum of clipped and unclipped losses
-        pg_loss = masked_sum(
-            torch.max(pg_losses, pg_losses2) / loss_mask_sum, loss_mask
-        ) / bsz  # float
-        pg_clipfrac = masked_sum(
-            torch.gt(pg_losses2, pg_losses).float() / loss_mask_sum, loss_mask
-        ) / bsz  # float
+        pg_loss = (
+            masked_sum(torch.max(pg_losses, pg_losses2) / loss_mask_sum, loss_mask)
+            / bsz
+        )  # float
+        pg_clipfrac = (
+            masked_sum(
+                torch.gt(pg_losses2, pg_losses).float() / loss_mask_sum, loss_mask
+            )
+            / bsz
+        )  # float
     else:
         # Take the maximum of clipped and unclipped losses
         pg_loss = torch.max(pg_losses, pg_losses2).mean()  # float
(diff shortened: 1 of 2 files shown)


rlinf/algorithms/embodiment/grpo_functions.py
```diff
@@ -265,37 +265,33 @@ def actor_loss_fn(
...
-    pg_losses = -advantages * ratio  # [bs, len]
+    pg_losses = -advantages * ratio
     pg_losses2 = -advantages * torch.clamp(
         ratio, 1.0 - clip_ratio_low, 1.0 + clip_ratio_high
-    )  # [bs, len]
+    )
 
     if loss_mask is not None:
         # Take the maximum of clipped and unclipped losses
         pg_loss = masked_sum(
```
guozhen1997: This line looks equivalent to the following calculation. If so, please call the function in utils directly.
`from rlinf.utils.utils import seq_mean_token_mean`
`pg_loss = seq_mean_token_mean(torch.max(pg_losses, pg_losses2), loss_mask)`

rlinf/algorithms/embodiment/grpo_functions.py
```diff
@@ -265,37 +265,33 @@ def actor_loss_fn(
...
         # Take the maximum of clipped and unclipped losses
         pg_loss = torch.max(pg_losses, pg_losses2).mean()  # float
         pg_clipfrac = torch.gt(pg_losses2, pg_losses).float().mean()  # float
 
     # Compile metrics for logging
     metrics_data = {
-        "actor/loss": pg_loss.detach().item(),
-        "actor/pg_loss": pg_loss.detach().item(),
-        "actor/pg_clipfrac": pg_clipfrac.detach().item(),
+        "actor/raw_loss": pg_loss.detach().item(),
```
guozhen1997, line 297: What is the difference between `actor/raw_loss` and `actor/policy_loss`? It is recommended to keep one and rename it to `pg_loss`.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -342,6 +347,8 @@ def run_training(self):
             append_to_dict(metrics, data)
 
         mean_metric_dict = {key: np.mean(value) for key, value in metrics.items()}
```
guozhen1997, line 351: For variables that need to be logged, you need to convert them to scalars when adding the metrics dictionary, rather than calculating `np.mean` here.

rlinf/algorithms/embodiment/grpo_functions.py
zanghz21: No, they are different because the data has been shuffled in a different way.

rlinf/algorithms/embodiment/grpo_functions.py
zanghz21: Currently, there is no difference. However, if we add entropy loss, I want it is included in the raw_loss. (like the current practice in ppo_functions.)

PR #37, opened 2025-08-27
fix: value-head require grad; override maniskill put-on-carrot-env; m…

zanghz21:
…odify grpo func loss weight

### Description

- Fix the bug about value-head gradient. Previously, the value-head did not have gradients under lora setting, and did not use separated learning rate when lora is False. 
- Change some loss ratio computation in grpo with mask settings.
- Override the ManiSkill PutCarrotOnPlate task to exclude the _settle function.

### Motivation and Context

- Fix ppo bug. Make ppo training stable.
- Refine grpo loss weight.
- Simplify the docker setup.

### How has this been tested?

- Run embodiment test scripts. 

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/algorithms/embodiment/grpo_functions.py
+++ b/rlinf/algorithms/embodiment/grpo_functions.py
@@ -278,9 +278,7 @@ def actor_loss_fn(
         # Take the maximum of clipped and unclipped losses
         pg_loss = (torch.max(pg_losses, pg_losses2) / loss_mask_ratio) * loss_mask
         pg_loss = pg_loss.mean()
-        pg_clipfrac = (
-            torch.gt(pg_losses2, pg_losses).float() / loss_mask_ratio
-        ) * loss_mask
+        pg_clipfrac = torch.gt(pg_losses2, pg_losses).float() * loss_mask
         pg_clipfrac = pg_clipfrac.mean()
     else:
         # Take the maximum of clipped and unclipped losses


rlinf/algorithms/embodiment/grpo_functions.py
```diff
@@ -265,7 +263,6 @@ def actor_loss_fn(
             - actor/pg_clipfrac: Fraction of clipped policy gradient loss
```
QuanluZhang, line 265: `loss_mask` and `loss_mask_ratio` are not included in docstring

rlinf/models/__init__.py
```diff
@@ -200,6 +200,10 @@ def get_model(model_path, cfg: DictConfig, override_config_kwargs=None):
         else:
             model = PeftModel.from_pretrained(model, cfg.lora_path, is_trainable=True)
 
+    for name, param in model.named_parameters():
+        if "value_head" in name:
+            param.requires_grad = True
```
QuanluZhang: this part is a little hacky, which model requires this logic? If you place this logic here, it means for all the models with the name "xxxvalue_headxxx" params must be `requires_grad=True`. Is this our expected behavior?

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -304,6 +304,10 @@ def run_training(self):
                 prev_values = data["prev_values"]
                 loss_mask = data.get("loss_mask", None)
                 loss_mask_sum = data.get("loss_mask_sum", None)
+                if hasattr(self.cfg.env.train, "max_episode_steps"):
+                    max_episode_steps = self.cfg.env.train.max_episode_steps
+                else:
+                    max_episode_steps = self.cfg.env.train.init_params.max_episode_steps
```
QuanluZhang: what is the difference between `train.max_episode_steps` and `train.init_prams.max_episode_steps`? different configurations in yaml?

rlinf/models/__init__.py
zanghz21: Currently, we assume that the value_head serves as the critic in the ppo setting.

rlinf/workers/actor/fsdp_actor_worker.py
zanghz21: Currently in maniskill setting, this param is under init_params, in libero setting, this param is directly under env.train. The contents under init_params are determined by the simulator. Maybe we can add a new attr under env.train in the maniskill setting, but this requires modification in yaml.

zoeyuchao: I support adding a new attribute in the maniskill setting. It is okay to modify maniskill yaml.

rlinf/models/__init__.py
zoeyuchao: It is okay here, the name of value_head is not used that much in other settings, but maybe we can leave a comment here, say, Attn: make sure the name of value head only means the critic of PPO.

examples/embodiment/config/maniskill_ppo_openvla_quickstart.yaml
```diff
@@ -36,6 +36,7 @@ runner:
 algorithm:
   auto_reset: True
   ignore_terminations: True
+  use_fixed_reset_state_ids: False
```
zoeyuchao, line 39: maybe we can add some description about this term, what does it control?

zoeyuchao (approved):
I have reviewed all changes and leave comments, only minor comments are needed, soi approved this pr. 
Good Job, Hongzhi!

XuS1994 (approved):

examples/embodiment/config/maniskill_ppo_openvla_quickstart.yaml
zanghz21: Add it in the doc

rlinf/workers/actor/fsdp_actor_worker.py
zanghz21: done

rlinf/models/__init__.py
zanghz21: add 1 comment on value-head when initialize the optimizer params group

rlinf/algorithms/embodiment/grpo_functions.py
zanghz21: done

PR #41, opened 2025-08-27
refactor(math): unified collocated and pipeline implementation

andylin-hao:
### Description

This PR implements unified interfaces of collocated and pipelining mode.

### Motivation and Context

Previously, math collocated mode and pipelining modes are implemented in different runners and workers, leading to substantial code duplication.

### How has this been tested?

Tested for math 1.5B collocated, without dynamic batch size.

### Types of changes

### Checklist:

- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


rlinf/data/io_struct.py
```diff
@@ -172,13 +178,14 @@ class RolloutResult:
     """
 
     num_sequence: int
+    group_size: int
```
QuanluZhang, line 193: `group_size` is not easy to understand, can you add comments to explain a little bit?

rlinf/data/io_struct.py
```diff
@@ -302,6 +319,21 @@ def merge_result_list(
...
                     merged_result.rollout_logprobs = []
                 merged_result.rollout_logprobs.extend(res.rollout_logprobs)
+
+        # TODO Align advantages and rewards format
+        if (
+            isinstance(merged_result.advantages, list)
+            and len(merged_result.advantages) > 0
+            and isinstance(merged_result.advantages[0], torch.Tensor)
+        ):
+            merged_result.advantages = torch.cat(merged_result.advantages, dim=0)
```
QuanluZhang: should we add `else`? if `merged_result.advantages[0]` is not tensor, then we should raise or ignore?

rlinf/data/io_struct.py
andylin-hao: This is the same name as in the yaml config, which is the number of responses generated per prompt. If a comment is needed, we should add it in the yaml.

rlinf/data/io_struct.py
andylin-hao: OK

rlinf/utils/data_iter_utils.py
```diff
@@ -567,6 +564,34 @@ def get_iterator_dynamic(
     return itertools.chain(microbatches), partitions, n_micro_batch
 
 
+def split_dynamic_batch_size(
```
QuanluZhang, line 573: is it better to write unit test for this function?

rlinf/utils/data_iter_utils.py
```diff
@@ -567,6 +564,34 @@ def get_iterator_dynamic(
...
+    cp_world_size: int,
+    vpp_world_size: int,
+    max_tokens_per_mbs: int,
+    microbatch_group_size_per_vp_stage: int,
+):
+    """Split a global batch using dynamic batch sizing."""
+    max_tokens_per_mbs = max_tokens_per_mbs * cp_world_size
+    vpp_size = vpp_world_size
+    if vpp_size is not None and vpp_size > 1:
+        microbatch_group_size_per_vp_stage = microbatch_group_size_per_vp_stage
```
QuanluZhang, line 584: what is this line used for?

rlinf/utils/data_iter_utils.py
andylin-hao: This function is simply a wrapper of the original dynamic batch size implementation. Nothing has changed. So it would not be appropriate to add tests for dynamic batch size in this PR.

rlinf/utils/data_iter_utils.py
andylin-hao: I'm not sure. I just move the dynamic batch size code here.

Lin-xs (approved):

QuanluZhang (approved):

PR #47, opened 2025-08-30
refactor: unify algorithm functions for math & embodiment

YQXie99:
### Description
Originally, we have the following functions in `rlinf/algorithms`:
```
math/
	algo_functions:
		kl_penalty
		compute_grpo_advantages
		calculate_adv_and_returns
		critic_loss_fn
		actor_loss_fn
	utils:
		get_attention_masks_and_position_ids

embodied/
	utils:
		actor_loss_fn
		calculate_advantages_and_returns
		compute_loss_mask
		append_to_dict
		huber_loss
		compute_rollout_metrics
		compute_evaluate_metrics
		compute_split_num
	ppo_functions:
		compute_advantages_and_returns
		actor_critic_loss_fn
		actor_loss_fn
		critic_loss_fn
	grpo_functions:
		compute_advantages
		compute_advantages_with_loss_mask
		actor_loss_fn
```
I reorganize the above functions to unify function calls made by `math` and `embodiment`. I move the functions that are not called by other `/algorithms`  functions to `utils/worker_utils`, and move `algorithms/math/verifier` to `tools/`. 

Now we have the following functions:
```
algorithm/
	utils:
		kl_penalty
		huber_loss
		preprocess_advantages_inputs
		preprocess_loss_inputs
	advantages:
		compute_ppo_advantages_and_returns
		compute_grpo_advantages_and_returns
		compute_math_grpo_advantages
	losses:
		ppo_actor_critic_loss_fn
		grpo_actor_loss_fn
		math_actor_loss_fn
	registry:
		calculate_adv_and_returns
		actor_loss
utils/
	worker_utils:
		compute_loss_mask
		append_to_dict
		compute_rollout_metrics
		compute_evaluate_metrics
		compute_split_num
```

### Tests
- Simple unit test with test case written in the script: each single advantage / loss function generates the same output as before. 
- Both math & embodiment experiments can launch normally, yet no training result is available yet. 

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/workers/actor/megatron_actor_worker.py
+++ b/rlinf/workers/actor/megatron_actor_worker.py
@@ -385,7 +385,6 @@ class MegatronActor(MegatronModelManager, Worker):
                     loss_mask=mask,
                 )
 
-                # logging_loss = loss.detach()
                 entropy_loss = torch.zeros(1, device=loss.device)
                 if self.calculate_entropy:
                     entropy = output["entropy"][:, -response_len - 1 : -1].contiguous()
--- a/docs/source-en/rst_source/tutorials/rlalg/dapo.rst
+++ b/docs/source-en/rst_source/tutorials/rlalg/dapo.rst
@@ -86,8 +86,8 @@ Currently, the framework supports DAPO for LLM math tasks.
 
   algorithm:
     # Core DAPO settings (recommended not to change)
-    adv_type: grpo
-    loss_type: ppo
+    adv_type: math_grpo # math_grpo for math tasks, embodied_grpo for embodied tasks
+    loss_type: math_ppo_actor # math_ppo_actor for math tasks, embodied_grpo for embodied tasks
     loss_agg_func: "token-mean"
     use_valid_token_scale: True # Divide advantage by valid token count → token-level loss
 
--- a/docs/source-en/rst_source/tutorials/rlalg/grpo.rst
+++ b/docs/source-en/rst_source/tutorials/rlalg/grpo.rst
@@ -72,8 +72,8 @@ An example configuration for LLM math tasks is shown below:
 
   algorithm:
     # Core GRPO settings (recommended not to change)
-    adv_type: grpo
-    loss_type: ppo
+    adv_type: math_grpo # math_grpo for math tasks, embodied_grpo for embodied tasks
+    loss_type: math_ppo_actor # math_ppo_actor for math tasks, embodied_grpo for embodied tasks
     loss_agg_func: "token-mean"
 
     # Algorithm parameters (typically require tuning)
--- a/docs/source-zh/rst_source/tutorials/rlalg/dapo.rst
+++ b/docs/source-zh/rst_source/tutorials/rlalg/dapo.rst
@@ -79,8 +79,8 @@ DAPO 采用非对称裁剪策略，设置更高的上限
 
   algorithm:
     # 核心 DAPO 设置（建议不要修改）
-    adv_type: grpo
-    loss_type: ppo
+    adv_type: math_grpo # 推理任务使用 math_grpo, 具身任务使用 embodied_grpo
+    loss_type: math_ppo_actor # 推理任务使用 math_ppo_actor, 具身任务使用 embodied_grpo
     loss_agg_func: "token-mean"
     use_valid_token_scale: True # 优势除以有效 token 数 → token 级损失
 
--- a/docs/source-zh/rst_source/tutorials/rlalg/grpo.rst
+++ b/docs/source-zh/rst_source/tutorials/rlalg/grpo.rst
@@ -73,8 +73,8 @@
 
   algorithm:
     # 核心 GRPO 设置（建议不要修改）
-    adv_type: grpo
-    loss_type: ppo
+    adv_type: math_grpo # 推理任务使用 math_grpo, 具身任务使用 embodied_grpo
+    loss_type: math_ppo_actor # 推理任务使用 math_ppo_actor, 具身任务使用 embodied_grpo
     loss_agg_func: "token-mean"
 
     # 算法参数（通常需要调优）


examples/math/config/qwen2.5-1.5b-grpo-megatron-pipeline.yaml
```diff
@@ -58,7 +58,7 @@ algorithm:
   max_num_gen_batches: 1
 
   # GRPO loss params
-  loss_type: ppo
+  loss_type: math_actor
```
zoeyuchao: here, math_actor, we use _, but grpo-math, we use -, any reason?

rlinf/algorithms/advantages.py
```diff
@@ -0,0 +1,213 @@
...
+import torch
+
+from rlinf.algorithms.registry import register_advantage
+
+
+@register_advantage("ppo")
+def compute_ppo_advantages_and_returns(**kwargs) -> Tuple[torch.Tensor, torch.Tensor]:
+    """
+    Calculate advantages and returns for Proximal Policy Optimization (PPO).
+    NOTE: currently this function does not support auto-reset.
```
zoeyuchao, line 28: Why not support auto-reset? PPO doesn't care about the reset? Hongzhi supports that in VLA imo.

rlinf/algorithms/losses.py
```diff
@@ -0,0 +1,330 @@
...
+    assert loss_mask is not None
+
+    loss_mask_count = loss_mask.count_nonzero() or 1
+    # For numerical stability.
+    ratio = torch.where(loss_mask, torch.exp(logprobs - old_logprobs), 0)
+    approx_kl = torch.where(loss_mask, (logprobs - old_logprobs).detach(), 0.0)
+
+    clipped_ratio = torch.clamp(ratio, 1.0 - eps_clip, 1.0 + eps_clip)
+    pg_loss1 = -advantages * ratio
+    pg_loss2 = -advantages * clipped_ratio
```
zoeyuchao: the pg_loss and policy_loss  in the above code, i think they are the same?

rlinf/algorithms/losses.py
```diff
@@ -0,0 +1,330 @@
...
+    assert advantages.dtype == torch.float32
+
+    assert loss_mask is not None
+
+    loss_mask_count = loss_mask.count_nonzero() or 1
+    # For numerical stability.
+    ratio = torch.where(loss_mask, torch.exp(logprobs - old_logprobs), 0)
+    approx_kl = torch.where(loss_mask, (logprobs - old_logprobs).detach(), 0.0)
+
+    clipped_ratio = torch.clamp(ratio, 1.0 - eps_clip, 1.0 + eps_clip)
```
zoeyuchao, line 215: i think the eps_clip is the same as clip ratio high and low?

rlinf/algorithms/losses.py
```diff
@@ -0,0 +1,330 @@
...
+    proportion_clipped = (
+        clip_mask.logical_and_(loss_mask).count_nonzero() / loss_mask_count
+    )
+    approx_kl = approx_kl.sum() / loss_mask_count
+
+    dual_cliped_ratio = torch.where(dual_clip_mask, ratio, 0)
+
+    return (
+        pg_loss,
+        proportion_clipped,
```
zoeyuchao: also this one as clip_fraction

YQXie99:
I have tested the refactored codes on both math and embodiment tasks. I think the performance shows that the refactored codes are correct.

# math(1.5B)
math

# embodiment-ppo
ppo1
ppo2

# embodiment-grpo
grpo1
grpo2

rlinf/algorithms/__init__.py
```diff
@@ -11,3 +11,4 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from . import advantages, losses, registry  # noqa: F401
```
zoeyuchao, line 14: what is the meaning of F401？

YQXie99: To ignore format errors. This line is essential for the code, but it cannot pass the lint check.

examples/math/config/qwen2.5-1.5b-grpo-megatron-pipeline.yaml
YQXie99: No specific reasons. I will change to '_' for both cases.

rlinf/algorithms/losses.py
YQXie99: Yes. There is certain redundancy in parameter definitions. Here the 'eps_clip' is used for standard grpo / ppo naming format. When upgraded to dapo, this will be changed into 'eps_clip_high' and 'eps_clip_low' as well.

rlinf/algorithms/advantages.py
YQXie99: I am not sure. This comment is copied from the comment of the original function.

zoeyuchao: @zanghz21 I guess you know the reason here, could u double check this one?

zoeyuchao (approved):
I have checked the files, looks really nice.

rlinf/utils/worker_utils.py
andylin-hao, line 1: Would it be better to rename `worker_utils.py` to something like `metrics_utils.py`? `worker_utils` is a bit vague.

YQXie99:
# Note
Originally there are two functions to calculate grpo advantage for embodiment tasks (`compute_advantages` & `compute_advantages_with_loss_mask`). After testing, I find the two functions behave the same both with and without masks, and therefore I keep only one grpo loss function left. This should be double-checked.

XuS1994 (approved):

rlinf/algorithms/losses.py
```diff
@@ -0,0 +1,333 @@
...
+from typing import Dict, Tuple
+
+import torch
+
+from rlinf.algorithms.registry import register_policy_loss
+from rlinf.algorithms.utils import huber_loss
+
+
+@register_policy_loss("ppo")
+def ppo_actor_critic_loss_fn(**kwargs) -> Tuple[torch.Tensor, Dict]:
```
guozhen1997: It’s not a common practice to calculate the actor and critic loss in the same function. I suggest registering this as **embodied_ppo**, and renaming the function to **compute_embodied_ppo_actor_critic_loss**.

rlinf/algorithms/losses.py
```diff
@@ -0,0 +1,333 @@
...
+        "actor/raw_loss": policy_loss.detach().item(),
+        "actor/policy_loss": policy_loss.detach().item(),
+        "actor/policy_clipfrac": clip_fraction.detach().item(),
+        "actor/ppo_kl": ppo_kl.detach().item(),
+    }
+    return policy_loss, metrics_data
+
+
+@register_policy_loss("math_actor")
+def math_actor_loss_fn(**kwargs):
```
guozhen1997: I suggest registering this as the default **ppo** loss function, and renaming this function to **compute_ppo_actor_loss** or **compute_policy_loss**.

rlinf/algorithms/losses.py
```diff
@@ -0,0 +1,333 @@
...
+        "critic/value_loss": value_loss.detach().item(),
+        "critic/value_clip_ratio": value_clip_ratio.detach().item(),
+        "actor/entropy_loss": entropy_loss.detach().item(),
+    }
+
+    return loss, metrics_data
+
+
+@register_policy_loss("grpo")
+def grpo_actor_loss_fn(**kwargs) -> Tuple[torch.Tensor, Dict]:
```
guozhen1997: Can this function be merged with the **math_actor_loss_fn** as default loss function? From what I see, the only difference here is the additional handling of **loss_mask_ratio**. If it cannot be merged, can it be used for math training? If not, it should be renamed to **embodied_grpo**.

rlinf/algorithms/registry.py
```diff
@@ -0,0 +1,86 @@
...
+        raise ValueError(f"Loss {name} not registered")
+    return LOSS_REGISTRY[name]
+
+
+def actor_loss(**kwargs) -> Tuple[torch.Tensor, Dict]:
+    """
+    Unified actor loss entry.
+    """
+    # Step 1: preprocess
+    kwargs = preprocess_loss_inputs(**kwargs)
```
guozhen1997: `preprocess_loss_inputs` should be placed outside the `actor_loss` function, because preprocessing itself should be clearly perceived by the user as part of the explicit processing logic. Moreover, the logic in `preprocess_loss_inputs` only applies to embodied training, while reasoning RL typically does not involve additional processing logic.

rlinf/algorithms/registry.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def calculate_adv_and_returns(**kwargs) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
+    """
+    Unified entry for advantage + return computation.
+    Accepts variable keyword arguments, preprocesses them, then dispatches
+    to specific algorithm via registry.
+    """
+    # Step 1: preprocess
+    kwargs = preprocess_advantages_inputs(**kwargs)
```
guozhen1997: same as the `preprocess_loss_inputs`, should be outside of `calculate_adv_and_returns` function

rlinf/algorithms/advantages.py
```diff
@@ -0,0 +1,211 @@
...
+
+from typing import Optional, Tuple
+
+import torch
+
+from rlinf.algorithms.registry import register_advantage
+
+
+@register_advantage("ppo")
+def compute_ppo_advantages_and_returns(**kwargs) -> Tuple[torch.Tensor, torch.Tensor]:
```
guozhen1997: I suggest registering this function as `embodied_gae`, and renaming as `compute_embodied_gae_advantages_and_returns`.

rlinf/algorithms/advantages.py
```diff
@@ -0,0 +1,211 @@
...
+    rewards = kwargs["rewards"]
+    dones = kwargs["dones"]
+    num_group_envs = kwargs.get("num_group_envs", 1)
+    group_size = kwargs.get("group_size", 2)
+    normalize_advantages = kwargs.get("normalize_advantages", True)
+    loss_mask = kwargs.get("loss_mask", None)
+    rollout_epoch = kwargs.get("rollout_epoch", 1)
+    epsilon = kwargs.get("epsilon", 1e-6)
+
+    # if loss_mask is None:
```
guozhen1997: remove 112-113

rlinf/algorithms/advantages.py
```diff
@@ -0,0 +1,211 @@
...
+        flattened_returns = (flattened_returns - mean_returns) / (std_retuns + 1e-5)
+
+    advantages = flattened_advantages.reshape(num_chunk, chunk_size, -1).transpose(1, 2)
+    returns = flattened_returns.reshape(num_chunk, chunk_size, -1).transpose(1, 2)
+
+    return advantages, returns
+
+
+@register_advantage("grpo")
+def compute_grpo_advantages_and_returns(
```
guozhen1997: I suggest registering this function as `embodied_grpo`, and renaming as `compute_embodied_grpo_advantages`.

rlinf/algorithms/losses.py
YQXie99: I do believe they can be merged (though not elegantly), but this need further tests. Renaming the function is more efficient.

rlinf/algorithms/registry.py
YQXie99: Where should I put these functions? Do you have any suggestions?

rlinf/algorithms/losses.py
YQXie99: Since PPO involves the training of both the actor and the critic, this function looks like a standard implementation. I will change the names, though.

rlinf/algorithms/losses.py
YQXie99: I change it to `math_ppo_actor` and `compute_math_ppo_actor_loss`.

docs/source-en/rst_source/tutorials/rlalg/ppo.rst
zhexuanxu, line 1: Why does it have to be “embodied PPO” here? This part of the documentation is introducing the algorithm, so it should be unrelated to either embodied or math—it should be task-agnostic.

rlinf/workers/actor/megatron_actor_worker.py
```diff
@@ -374,24 +375,17 @@ def loss_func(output):
...
+                    loss_agg_func=self.loss_agg_func,
+                    logprobs=curr_logprobs,
+                    old_logprobs=prev_logprobs,
+                    advantages=advantages,
+                    eps_clip=self.ratio_eps,
+                    loss_mask=mask,
                 )
 
-                logging_loss = loss.detach()
+                # logging_loss = loss.detach()
```
guozhen1997: remove this line

docs/source-en/rst_source/tutorials/rlalg/ppo.rst
YQXie99: Currently embodiment and math tasks require different implementations for the same algorithm. It can be unified, but with great efforts. Maybe in future versions we can enhance this feature.

PR #52, opened 2025-09-01
feat(algorithm)!: New method about iterating LIBERO prompts.

WinstonWmj:
add a new method about shuffling the iterating prompts according to world_size in libero; update the filtering mechanism based on rewards to filter the  unwanted samples;

### Description

1. **Fixed inconsistent initial states after reset in the LIBERO environment:** Discovered and resolved the issue where each reset introduced jitter in the initial state. Adjusted the reset code and explicitly fixed the initial seed to eliminate the problem.
2. **Updated the `update_reset_state_ids` method:** For LIBERO training, implemented a shuffle-and-traverse strategy for prompts. Now, all prompts are divided across processes according to `world_size`. Each process traverses all prompts in its assigned subset, then reshuffles the order before the next traversal, repeating this cycle.
3. **Added a reward-based filtering algorithm:** Introduced a method to filter trajectories based on the reward obtained in each episode. Currently, this is only supported for sparse rewards.
4. **Enhanced GRPO loss function with PPO-KL logging:** Added recording of the `ppo_kl` metric during loss computation in the GRPO function, as an auxiliary measure to monitor the extent of policy updates.

### Motivation and Context

1. Since variations in the initial state during environment resets can affect model convergence, we aligned with the baseline approach: the initial state should remain consistent for each training run.
2. The filter mechanism was introduced to accelerate model convergence. Without the filter, convergence is still possible, but it requires significantly more RL steps.
3. The `ppo_kl` metric was added to provide a better measure of the model’s convergence progress.

### How has this been tested?

- Use the docker for embodiment, and use the conda for openvlaoft, and train cmd is as following `bash examples/embodiment/run_embodiment.sh libero_10_grpo_openvlaoft` . What's more, I have also tested the evaluation use `bash examples/embodiment/eval_embodiment.sh libero_10_grpo_openvlaoft_eval`

### Types of changes

### Checklist:


rlinf/algorithms/embodiment/grpo_functions.py
zanghz21, line 1: The shape is not correct here, should be omitted

rlinf/algorithms/embodiment/grpo_functions.py
```diff
@@ -250,13 +250,11 @@ def actor_loss_fn(
...
-        log_prob (torch.Tensor): Current log probabilities
-        old_log_prob (torch.Tensor): Previous log probabilities
-        advantages (torch.Tensor): Advantage values of shape
+        log_prob (torch.Tensor): Current log probabilities of shape (bs,) or (bs, action-token-length)
+        old_log_prob (torch.Tensor): Previous log probabilities (will be repeated to match current shape)
+        advantages (torch.Tensor): Advantage values of shape (bs,)
         clip_ratio_high (float): Upper clipping ratio for PPO
         clip_ratio_low (float): Lower clipping ratio for PPO
-        loss_mask (Optional[torch.Tensor]): Mask tensor of shape to apply to the loss
-        loss_mask_ratio (Optional[torch.Tensor]): Ratio tensor for normalizing the loss when using a mask
```
zanghz21: the function description should adopt the original version.

rlinf/algorithms/embodiment/grpo_functions.py
```diff
@@ -278,17 +276,22 @@ def actor_loss_fn(
         # Take the maximum of clipped and unclipped losses
         pg_loss = (torch.max(pg_losses, pg_losses2) / loss_mask_ratio) * loss_mask
         pg_loss = pg_loss.mean()
-        pg_clipfrac = torch.gt(pg_losses2, pg_losses).float() * loss_mask
+        pg_clipfrac = (
+            torch.gt(pg_losses2, pg_losses).float() / loss_mask_ratio
```
zanghz21: pg_clipfrac should not divide loss_mask_ratio

rlinf/envs/env_manager.py
```diff
@@ -180,7 +181,7 @@ def __init__(self, cfg, rank, env_cls, enable_offload=False):
             self.env = None
         else:
             self.env_cls = env_cls
-            self.env = self.env_cls(cfg, rank)
+            self.env = self.env_cls(cfg, rank, world_size)
```
zanghz21, line 184: if you modify here, you might also change the init param of the ManiSkillEnv

zanghz21 (commented):
Overall is good. Please check whether the code is compatible with maniskill setting as well.

rlinf/algorithms/embodiment/grpo_functions.py
WinstonWmj: aha? which one?

rlinf/algorithms/embodiment/grpo_functions.py
WinstonWmj: Okay, I’ve reverted this part of the docstring to the previous version. It’s true that I shouldn’t have modified it here.

rlinf/algorithms/embodiment/grpo_functions.py
WinstonWmj: Okay and Done!

rlinf/envs/env_manager.py
WinstonWmj: Okay, I have commited the change in the latest version.

rlinf/algorithms/embodiment/grpo_functions.py
```diff
@@ -278,17 +278,20 @@ def actor_loss_fn(
         # Take the maximum of clipped and unclipped losses
         pg_loss = (torch.max(pg_losses, pg_losses2) / loss_mask_ratio) * loss_mask
         pg_loss = pg_loss.mean()
-        pg_clipfrac = torch.gt(pg_losses2, pg_losses).float() * loss_mask
+        pg_clipfrac = (torch.gt(pg_losses2, pg_losses).float()) * loss_mask
```
XuS1994: remove round brackets

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -161,6 +161,55 @@ def _process_received_rollout_batch(self, rollout_batch):
             rollout_batch["loss_mask"] = loss_mask
             rollout_batch["loss_mask_sum"] = loss_mask_sum
 
+        # filter data by rewards
+        if self.cfg.algorithm.get("filter_rewards", False):
```
guozhen1997, line 165: Is the `filter_rewards` parameter only for libero? If this parameter is enabled in maniskill, will it work normally? Is assert required?

WinstonWmj: 1. This functionality (`filter_rewards`) is an optional feature, independent of the environment or model, so it is also applicable to the ManiSkill environment. We only tested it in the LIBERO environment, and since it produced better curves, that is sufficient to demonstrate the effectiveness of this feature.

2. In fact, this assertion will always pass in our current code, since the batch size is computed as the product of the group size. However, to ensure that future modifications do not inadvertently affect performance here, I think it is still worth keeping this assertion.

rlinf/algorithms/embodiment/grpo_functions.py
WinstonWmj: okay

rlinf/workers/actor/fsdp_actor_worker.py
guozhen1997: ok, LGTM!

guozhen1997 (approved):

XuS1994 (approved):

PR #80, opened 2025-09-10
feat: support fsdp2, add lr scheduler, enhance fsdp ckpt save and load

guozhen1997:
### Description

- Support FSDP2
- Add learning rate scheduler, support "constant" and "cosine" scheduler
- Enhance checkpoint save and load for model, optimizer, lr scheduler and random state
- Add grad clip

### How has this been tested?

FSDP and FSDP2 are tested in embodiment training

### Types of changes

### Checklist:


examples/embodiment/config/libero_10_grpo_openvlaoft.yaml
```diff
@@ -118,8 +118,34 @@ actor:
...
+
+    # Whether to reshard parameters after forward pass to save memory
+    # false: keep parameters sharded (saves memory, may impact performance)
+    # true: unshard parameters after forward (uses more memory, better performance)
+    reshard_after_forward: false
+    
+    # Wrap policy configuration for module sharding
+    wrap_policy:
+      # Whether to disable automatic module wrapping
+      disable: false
```
QuanluZhang: what is this option used for?

tests/e2e_tests/embodied/ppo_openvla.yaml
```diff
@@ -128,6 +128,31 @@ actor:
...
 
+  fsdp_config:
+    # FSDP strategy: "fsdp" or "fsdp2"
+    strategy: "fsdp"
+
+    # Sharding strategy: "full_shard" (shard parameters, gradients, optimizer states),
+    # "shard_grad_op" (shard gradients and optimizer states only), 
+    # "hybrid_shard" (combines data parallelism and model parallelism - sharding within FSDP groups, replicate across DDP groups),
+    # "no_shard" (no sharding)
+    sharding_strategy: "full_shard"
```
QuanluZhang: As you have supported all the sharding strategies, it would be better to test all the sharding strategies to make sure at least they are executable, though the training correctness is not guaranteed.

examples/embodiment/config/libero_10_grpo_openvlaoft.yaml
guozhen1997: When disable=true, for fsdp1, the root module can still be sharded, but for fsdp2, no sharding will occur, which may cause errors. I will remove this configuration.

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
```diff
@@ -13,36 +13,81 @@
...
-    def __init__(self, cfg: DictConfig):
+    def __init__(self, cfg: DictConfig, world_size: int, logger=None):
         self._cfg = cfg
-        self.torch_dtype = torch_dtype_from_precision(self._cfg.model.precision)
+        self._logger = logger
+
+        mixed_precision_config = self._cfg.fsdp_config.get("mixed_precision", None)
+        if mixed_precision_config is not None:
+            self.param_dtype = torch_dtype_from_precision(
+                mixed_precision_config.get("param_dtype", "bf16")
```
QuanluZhang: should we update doc accordingly, for example, the yaml doc?

guozhen1997: ok, I will add the mix precision config to the yaml and later update the FSDP training backend in hybrid engines in our document.

tests/e2e_tests/embodied/ppo_openvla.yaml
guozhen1997: Eight combinations of two FSDP strategies and four sharding strategies have been tested and work properly. FSDP2 shows better GPU memory efficiency compared to FSDP. I suggest FSDP2 as the default configuration.
I am currently working on end-to-end correctness alignment.

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
```diff
@@ -89,7 +124,7 @@ def model_provider_func(self) -> torch.nn.Module:
 
             model = auto_model_class.from_pretrained(
                 cfg.model.model_path,
-                torch_dtype=self.torch_dtype,
+                torch_dtype=torch.float32 if self.use_fp16 else self.torch_dtype,
```
Lin-xs: Why is `torch.float32` used when `self.use_fp16=True`?

rlinf/hybrid_engines/fsdp/strategy/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 root
```
Lin-xs: Copyright 2025 The RLinf Authors.

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
anHappyDog: because grad_scaler need fp32 grad which can only be set when param_type is fp32,  or it will report ValueError saying it does not support fp16 grad.

rlinf/hybrid_engines/fsdp/strategy/fsdp1.py
```diff
@@ -0,0 +1,389 @@
...
+            buffer_dtype=buffer_dtype,
+        )
+
+        sharding_strategy = get_sharding_strategy(
+            self.cfg.fsdp_config.sharding_strategy
+        )
+
+        is_vla_model = (
+            True
+            if self.cfg.model.get("model_name", None) in ["openvla", "openvla_oft"]
```
Lin-xs: Add a dedicated function or method to get `is_vla_model` to avoid hard code model name in FSDPStrategy, which looks strange.

anHappyDog: done.

rlinf/hybrid_engines/fsdp/strategy/__init__.py
anHappyDog: done.

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
Lin-xs: So FSDP fp16 training is not supported now?

rlinf/runners/embodied_runner.py
```diff
@@ -115,7 +115,7 @@ def run(self):
                     self.metric_logger.log(data=eval_metrics, step=_step)
 
             with self.timer("step"):
-                with self.timer("rollout"):
+                with self.timer("sync_weights"):
```
Lin-xs, line 121: Looks like this timer will record `sync_weights` and `rollout`. Do we need to set a timer for each function call?

anHappyDog: a new timer is set for generate_rollout

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
anHappyDog: If simply setting torch_type to fp16, FSDP is very likely to produce all nans due to precision issues. However, if setting torch_type to fp32 and use amp (which sets the calculation precision to fp16 as much as possible) + grad_scaler, the training effect is better and more stable(while not using too much VRAM). Perhaps I could add a switch to use amp so that users can select between only using only fp16 or turning on amp.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -368,13 +366,16 @@ def run_training(self, input_channel: Channel) -> Tuple[Dict, list]:
...
 
                     loss_mask = m_batch["attention_mask"][:, -self.response_len :]
-                    output = self.model(
-                        input_ids=input_ids,
-                        attention_mask=attention_mask,
-                        position_ids=position_ids,
-                        **multi_modal_inputs,
-                        use_cache=False,
-                    )
+                    with torch.amp.autocast(
```
qurakchin: Is this autocast disabled if self.use_fp16 is False?

qurakchin (changes requested):

rlinf/workers/actor/fsdp_actor_worker.py
anHappyDog: Yes, if the precision is bf16 or else, it will be off.

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
anHappyDog: I add a config for amp.

rlinf/workers/actor/fsdp_actor_worker.py
anHappyDog: I add a config for amp in yaml file, now is controlled by user.

PR #124, opened 2025-09-23
feat(scheduler): add accelerator abstraction for hetero/CPU workers

andylin-hao:
### Description

This PR adds support for workers running on heterogeneous devices including CPU, GPU, NPU at the scheduler module. As a result, RLinf no longer requires specific devices (GPUs) and a homogeneous number of devices across nodes.

This PR introduces the following breaking changes:
* The `Cluster` class no longer requires `num_gpus_per_node` parameter to initialize, but only `num_nodes`. This is to adapt to CPU nodes or hetero nodes with different types of accelerator & different number of accelerators per node (e.g., Ascend 910C with 16 NPUs per node). 
* Accordingly, the `num_gpus_per_node` configuration has been removed from yaml configuration files and the documentation.
* To obtain the number of accelerators/GPUs for each node or the whole cluster in worker, one only needs to recreate the cluster with no parameters like this:
```python
cluster = Cluster()
total_number_gpus = cluster.num_accelerators_in_cluster
```

### Motivation and Context

This is for supporting agentic scenarios where workers may or may not use devices like GPUs, as well as future support for hetero devices like AMD/Intel GPUs and NPUs.

### How has this been tested?

Tested on NV GPU and CPU-only workers. Has not been tested on AMD/INTEL/HUAWEI devices, whose support are still experimental.
Also tested on 16-node cluster with steadily-growing reward:
image

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/scheduler/placement/flexible.py
+++ b/rlinf/scheduler/placement/flexible.py
@@ -58,7 +58,8 @@ class FlexiblePlacementStrategy(PlacementStrategy):
         >>> my_worker = MyWorker.create_group().launch(
         ...     cluster=cluster, name="flexible_placement", placement_strategy=placement
         ... )
-        >>> my_worker.available_gpus().wait() # This will run 3 processes on the first node's GPU 4, 5, 6, 7, where the first process uses GPUs 4 and 5, the second process uses GPU 6, and the third process uses GPU 7.
+        >>> # This will run 3 processes on the first node's GPU 4, 5, 6, 7, where the first process uses GPUs 4 and 5, the second process uses GPU 6, and the third process uses GPU 7.
+        >>> my_worker.available_gpus().wait()
         [2, 1, 1]
 
     """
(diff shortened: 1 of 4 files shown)


QuanluZhang (approved):

rlinf/config.py
```diff
@@ -486,9 +486,12 @@ def validate_embodied_cfg(cfg):
     )
 
     # process num-envs
+    from rlinf.scheduler import Cluster
     from rlinf.utils.placement import HybridComponentPlacement
 
-    component_placement = HybridComponentPlacement(cfg)
+    component_placement = HybridComponentPlacement(
+        cfg, Cluster(num_nodes=cfg.cluster.num_nodes)
```
QuanluZhang, line 493: if the cluster does not have the information of num gpus per nodes, how do you check whether the specified accelerator indices is correct?

andylin-hao: The cluster will automatically collect each node's accelerator resources and adds them to compute the total number of accelerators in the cluster.

PR #150, opened 2025-10-06
fix: maniskill render, sim stage seed, rename step-level to action-level for reward type; support chunk-level with mask setting

zanghz21:
### Description

- Move the override func in get_obs to _get_obs_sensor_data in the PutOnPlateInScene25Main task. 
- Change the meaning of the rank in each simulator process instance. Currently, if two envs have different stage ids, their rank ids are different in the simulator instances(e.g. ManiskillEnv in rlinf/envs/offload_wrapper/maniskill_wrapper.py).
- Rename all step_level in the reward_type to action_level.
- Fix bug in chunk-level logprob type with mask in grpo. 
- Update the quickstart yaml.

### Motivation and Context

- Previously, videos in ManiSkill are inconsistent with the observations that the policy receives. 
- Previously, the envs have the same rank of env worker share the same seed for random states. That brings unexpected behavior. 
- Aligning notations in the tech report. 
- Previously, we did not test this setting, and thus it could not run. 
- Let the quickstart yaml becomes larger

### How has this been tested?

The video issue and the random seed issue is tested by visualization.

### Types of changes

### Checklist:


zoeyuchao (approved):

rlinf/workers/env/env_worker.py
```diff
@@ -96,22 +96,22 @@ def init_worker(self):
             from rlinf.envs.maniskill.maniskill_env import ManiskillEnv
 
             if not only_eval:
-                for _ in range(self.stage_num):
+                for stage_id in range(self.stage_num):
                     self.simulator_list.append(
                         EnvManager(
                             self.cfg.env.train,
-                            rank=self._rank,
+                            rank=self._rank * self.stage_num + stage_id,
```
QuanluZhang: according to my understanding, `stage_num` means how many parts the envs are divided into, but these parts are still in the same GPU. Why do you change `rank=` like this?

PR #154, opened 2025-10-09
fix(embodiment): fix error for using egl

vincentccc:
### Types of changes

### Checklist:

### DIFF
--- a/examples/embodiment/run_embodiment.sh
+++ b/examples/embodiment/run_embodiment.sh
@@ -4,6 +4,8 @@ export EMBODIED_PATH="$( cd "$(dirname "${BASH_SOURCE[0]}" )" && pwd )"
 export REPO_PATH=$(dirname $(dirname "$EMBODIED_PATH"))
 export SRC_FILE="${EMBODIED_PATH}/train_embodied_agent.py"
 
+# export MUJOCO_GL="osmesa"
+# export PYOPENGL_PLATFORM="osmesa"
 export MUJOCO_GL="egl"
 export PYOPENGL_PLATFORM="egl"
 export PYTHONPATH=${REPO_PATH}:$PYTHONPATH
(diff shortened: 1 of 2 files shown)


examples/embodiment/run_embodiment.sh
```diff
@@ -16,6 +18,10 @@ export PYTHONPATH=${LIBERO_REPO_PATH}:$PYTHONPATH
 export CUDA_LAUNCH_BLOCKING=1
 export HYDRA_FULL_ERROR=1
 
+# torch.compile may cause errors 
+# export TORCH_COMPILE_DISABLE=1
+# export TORCHDYNAMO_DISABLE=1
+
```
zanghz21, line 24: Please delete these new comments

zanghz21:
Thank you so much to help with this issue!!
I am wondering whether this also works well if pipline_stage_num > 1. In that case, there are multiple simulator processes running on one GPU, and they have a shared ``MUJOCO_EGL_DEVICE_ID''.
Also, I am not sure whether to use LOCAL_RANK or LOCAL_ACCELERATOR_ID

examples/embodiment/run_embodiment.sh
vincentccc: I experienced a torch.compile issue with the official Docker image on the A800 cluster, which I think may help.

vincentccc:
> Thank you so much to help with this issue!! I am wondering whether this also works well if pipline_stage_num > 1. In that case, there are multiple simulator processes running on one GPU, and they have a shared ``MUJOCO_EGL_DEVICE_ID''. Also, I am not sure whether to use LOCAL_RANK or LOCAL_ACCELERATOR_ID

IMHO, each Ray worker typically holds only one GPU. Therefore, both `MUJOCO_EGL_DEVICE_ID` and `CUDA_VISIBLE_DEVICES` should usually be set to `0` in a distributed training framework. It should still work fine even if multiple simulators run on the same GPU.

zoeyuchao:
Thanks for pointing out this. We have fixed this issue, so I close this PR.

PR #219, opened 2025-10-28
feat: Add a chat template format option for the tokenizer in dataset

FxxxxU:
### Description
Fix the format issues of the current dataset, support adding special tokens for the tokenizer's chat template in the dataset, and modify the relevant configuration options of the dataset.

### Motivation and Context
Solve the issue where RLinf can only load dataset after the chat template preprocessing and add some data options.

### How has this been tested?
Yes

### Types of changes

### Checklist:

### DIFF
--- a/docs/source-en/rst_source/examples/reasoning.rst
+++ b/docs/source-en/rst_source/examples/reasoning.rst
@@ -30,41 +30,48 @@ An example training sample looks like:
       "solutions": ["\\boxed{x}"]
    }
 
-We also support importing other dataset types. Please note the required format below:
+.. note::
 
-**When apply_chat_template = False**
+  Ensure the dataset is structured as the above using the default configurations.
+  Otherwise, read below configuration guidelines carefully to adapt RLinf to your dataset.
 
-- The prompt is used directly from the dataset without any modification.
-- Note: Make sure the prompt format is correct and verify that the dataset already contain tokenizer-specific special characters.
-- Expected dataset format:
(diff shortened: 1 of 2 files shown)


rlinf/data/datasets/math.py
```diff
@@ -33,6 +33,38 @@ def __init__(
         config: DictConfig,
         tokenizer: AutoTokenizer,
     ):
+        """
+        Initialize the MathDataset.
+
+        This constructor loads the dataset from specified paths and applies optional filtering
```
andylin-hao, line 39: This docstring is very helpful. Could you add this to `docs/source-en/rst_source/examples/reasoning.rst` as well? There is a `Dataset` Section there.

docs/source-en/rst_source/examples/reasoning.rst
```diff
@@ -30,6 +30,42 @@ An example training sample looks like:
...
+**When apply_chat_template = False**
+
+- The prompt is used directly from the dataset without any modification.
+- Note: Make sure the prompt format is correct and verify that the dataset already contain tokenizer-specific special characters.
+- Expected dataset format:
+
+.. code-block:: json
+
+    {
+      "prompt_key": <str>,
```
andylin-hao: Need to fix this with quotation.

rlinf/data/datasets/math.py
FxxxxU: update the docs.

PR #222, opened 2025-10-30
chore: remove useless split and repeat in RolloutRequest

anHappyDog:
### Description
Remove useless `split` and `repeat` function in `RolloutRequest`, aligning with SGLang.

### How has this been tested?

tested 1.5B in single machine.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/data/io_struct.py
+++ b/rlinf/data/io_struct.py
@@ -59,8 +59,8 @@ class RolloutRequest:
     n: int
     input_ids: List[List[int]]
     image_data: Union[List[List[bytes]], List[List[str]]]
-    answers: List[Union[List[str], dict]]
-    multi_modal_inputs: List[dict]
+    answers: List[Union[List[str], Dict]]
+    multi_modal_inputs: List[Optional[Dict]]
 
     def to_seq_group_infos(self) -> List["SeqGroupInfo"]:
         """Convert the RolloutRequest into a list of SeqGroupInfo objects.
@@ -103,7 +103,7 @@ class SeqGroupInfo:
     Attributes:
         id (int): Unique identifier for the sequence group.
         input_ids (List[int]): List of input IDs of the original sequence.
-        answer (Union[List[str], dict]): List of answers of the original sequence.(One sequence can have multiple equivalent answers), or a dict in case of vqa task.
+        answer (Union[List[str], Dict]): List of answers of the original sequence.(One sequence can have multiple equivalent answers), or a dict in case of vqa task.
         group_size (int): Number of sequences in the group.
         idx_completed (set[int]): Set of indices for sequences that have completed rollout and are ready for evaluation.
         idx_aborted (set[int]): Set of indices for sequences that have been aborted. These sequences need to be re-rolled out before they can be evaluated.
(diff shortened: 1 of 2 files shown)


rlinf/data/io_struct.py
```diff
@@ -185,13 +112,13 @@ class SeqGroupInfo:
 
     id: int
     input_ids: List[int]
-    answer: List[str]
+    answer: str
```
Lin-xs: `SeqGroupInfo.answer` is a list of string because one question may have multiple equivalent correct answers.

anHappyDog: However, in the `to_seq_group_infos` function, the member `answer` in `SeqGroupInfo` comes from an element of type `list[str]` in `RolloutRequest`. If the type of `answer` in `SeqGroupInfo` needs to be `List[str]`, then `to_seq_group_Infos` may need to be modified.

rlinf/data/io_struct.py
```diff
@@ -185,13 +112,13 @@ class SeqGroupInfo:
...
     group_size: int
     idx_completed: set[int] = field(init=False, compare=False)
     idx_aborted: set[int] = field(init=False, compare=False)
     results: List[Optional[Dict]] = field(init=False, compare=False)
     image_data: Optional[List] = None
-    multi_modal_inputs: Optional[List] = None
+    multi_modal_inputs: Optional[Dict] = None
```
Lin-xs, line 121: Is `multi_modal_inputs` always a dict for one request? Actually I don't know, @guozhen1997 said it is a list when we I merged main branch to my branch.

anHappyDog: For one request, `multi_modal_inputs` is a dict. It may be None, but could not be a list.

rlinf/data/io_struct.py
Lin-xs: The type hint of `RolloutRequest.answers` is wrong. You can replace it by `List[List[str]]`

Lin-xs: image

image

rlinf/data/io_struct.py
```diff
@@ -176,22 +103,22 @@ class SeqGroupInfo:
...
         group_size (int): Number of sequences in the group.
         idx_completed (set[int]): Set of indices for sequences that have completed rollout and are ready for evaluation.
         idx_aborted (set[int]): Set of indices for sequences that have been aborted. These sequences need to be re-rolled out before they can be evaluated.
         results (List[Optional[Dict]]): List storing result dictionaries for each sequence, or None if not yet available.
     """
 
     id: int
-    input_ids: List[int]
-    answer: List[str]
+    input_ids: list[int]
```
andylin-hao: Are we changing `List` to `list` now? If so, we might want to change it throughout the code. Otherwise, we shouldn't make the changes in one or two files.

anHappyDog: OK. I found that `List`, `Dict`, `Tuple` will be depreciated, so I changed these files. I will restore these files and make a new PR to change all the annotations if you think it's needed.

andylin-hao: SG

Lin-xs (approved):

PR #233, opened 2025-11-03
fix: channel cross-talk and concurrency limit of ray async actor

andylin-hao:
### Description

Fix channel put cross-talk issue and concurrency limit of ray asyncio.

Ray limits the number of concurrent asyncio tasks of async actor to 1000 by default, which leads to channel being blocked by a lot of gets (over 1000).
This PR adds max_concurrency option to the worker group launch arguments to resolve this.

This PR also adds CI tests for requirements/install.sh and docker build.

### Types of changes

### Checklist:

### DIFF
--- a/tests/unit_tests/test_channel.py
+++ b/tests/unit_tests/test_channel.py
@@ -403,13 +403,13 @@ class TestChannel:
     def test_stress(self, worker_groups, channel):
         """Tests the channel under heavy load."""
         producer: ProducerWorker = worker_groups[0]
-        num_items = 100
+        num_items = 1000
 
         data = producer.stress(channel, num_items).wait()[0]
         assert data == list(range(num_items))
 
-        # data = producer.stress_multiple_queues(channel, num_items).wait()[0]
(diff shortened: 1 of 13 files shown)


tests/unit_tests/test_channel.py
```diff
@@ -403,13 +403,13 @@ def test_channel_nowait(self, worker_groups, channel):
     def test_stress(self, worker_groups, channel):
         """Tests the channel under heavy load."""
         producer: ProducerWorker = worker_groups[0]
-        num_items = 100
+        num_items = 1000
```
QuanluZhang, line 406: how about changing the number to 2000, as you said the default value is 1000.

andylin-hao: This test is not meant for testing the concurrency issue of ray (I don't think we can test that without wasting a lot of test time), but how channel performs under a reasonable large number of operations. This in fact involves 1000 puts + 1000 gets so I think that's enough for now.

guozhen1997 (approved):

PR #240, opened 2025-11-06
feat: added ci to multi-turn agent rl demo

qurakchin:
### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/agent-e2e-tests.yml
+++ b/.github/workflows/agent-e2e-tests.yml
@@ -171,7 +171,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Megatron SGLang Pipeline mode
-        timeout-minutes: 20
+        timeout-minutes: 10
         run: |
           export REPO_PATH=$(pwd)
           source switch_env reason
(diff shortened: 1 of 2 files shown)


.github/workflows/agent-e2e-tests.yml
```diff
@@ -162,4 +162,17 @@ jobs:
...
+          bash tests/e2e_tests/reasoning/run.sh qwen2.5-vl-3b-grpo-collocated-fsdp-vllm
+
+  agent-rl-qwen-ppo-test:
+    runs-on: reason
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Megatron SGLang Pipeline mode
+        timeout-minutes: 20
```
QuanluZhang: 20 is too long, how about 10?

rlinf/runners/agent_runner.py
```diff
@@ -197,13 +202,13 @@ def run(self):
                         logging.info(
                             f"Step limit given by max_steps={self.max_steps} reached. Stopping run"
                         )
-                        return
+                        train_end = True
```
QuanluZhang: this is ugly, how about creating a function called `self.stop_and_cleanup`, calling it and add return here?

qurakchin: refactored using try-finally style.

.github/workflows/agent-e2e-tests.yml
qurakchin: edited.

PR #245, opened 2025-11-07
feat(sglang): validate weight between init from hf_ckpt and first sync from megatron

qurakchin:
### Description

When the feature is enabled, sglang will calc and save model weight norm dict at init. Then sglang will calc and check weight norm after first sync from megatron.
It can help to fit for new model and check the model is correctly synced or not.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
+++ b/rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
@@ -217,6 +217,9 @@ class Scheduler(_Scheduler):
                 weight_norm_dict[key] = posi_norm(value)
             self.weight_norm_dict = weight_norm_dict
 
+            # avoid release memory before norm kernel launch (gpu is async from cpu)
+            torch.cuda.synchronize()
+
         self._rlinf_worker.log_info(
             f"Running Scheduler dp rank {self._rlinf_worker.get_parent_rank()}, tp rank {self.tp_rank}, corresponding actor weight rank = {self.actor_weight_rank}"
         )


rlinf/workers/rollout/sglang/sglang_worker.py
```diff
@@ -189,6 +193,19 @@ async def _validate_weight_at_first(self):
             )
             print("===============================", flush=True)
 
+    async def _validate_weight_at_first_sync(self):
+        if (
+            torch_dtype_from_precision(self._cfg.rollout.precision) != torch.bfloat16
+            or torch_dtype_from_precision(self._cfg.actor.model.precision)
+            != torch.bfloat16
+        ):
+            self.log_warning(
```
Lin-xs: Why do we only support weight validation when `rollout.precision == actor.precision == bf16`? I think `rollout.precision == actor.precision` is enough.

rlinf/workers/rollout/sglang/sglang_worker.py
```diff
@@ -242,6 +261,8 @@ async def init_worker(self):
         self.log_info(f"SGLang worker {self._rank} initialized.")
         if self._cfg.rollout.validate_weight:
             await self._validate_weight_at_first()
+        if getattr(self._cfg.rollout, "validate_weight_first_sync", False):
```
Lin-xs: I think this field can be added to all reasoning configuration, otherwise developers and users don't know they can enable weight validation.

rlinf/hybrid_engines/sglang/common/tokenizer_manager.py
```diff
@@ -100,3 +109,11 @@ async def abort_generation(
     ):
         self.auto_create_handle_loop()
         await self.abort_generation_communicator(obj)
+
+    async def save_norm_weights(
```
Lin-xs: Actually `save_norm_weights` will be called only once after loading ckpt from disk at the begining, and the comparison will be processed only after the first weight synchronization. Is it necessary to add a dedicated io_struct to make it can be called externally?

rlinf/workers/rollout/sglang/sglang_worker.py
qurakchin: Our model ckpt precision is bf16, so if we use actor and rollout in bf16, we can get totally equal weights in sglang and megatron.

rlinf/hybrid_engines/sglang/common/tokenizer_manager.py
qurakchin: refactored

rlinf/workers/rollout/sglang/sglang_worker.py
qurakchin: done.

PR #253, opened 2025-11-11
feat(rollout): add auto scaling scheduler for vLLM, aligning with SGLang

anHappyDog:
### Description
Add auto scaling scheduler for vLLM, aligning with SGLang.

### How has this been tested?

Tested in single machine

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/utils/data_process.py
+++ b/rlinf/utils/data_process.py
@@ -50,11 +50,8 @@ async def _fetch_and_decode(
 
 
 async def process_image_data(
-    image_data: Optional[list[Union[bytes, str]]],
-) -> Optional[list[Image.Image]]:
-    if image_data is None:
-        return None
-
+    image_data: list[Union[bytes, str]],
+) -> list[Image.Image]:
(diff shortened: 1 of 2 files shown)


andylin-hao (dismissed):

rlinf/runners/reasoning_runner.py
```diff
@@ -183,7 +183,7 @@ def init_workers(self):
         # Init workers
         self.rollout.init_worker().wait()
         if self.use_pre_process_policy:
-            self.rollout.offload_engine().wait()
+            self.rollout.offload_model_weights().wait()
```
Lin-xs: `offload_model_weights` is too specific. Actually we can offload 'weight' or 'kvcache' separately, and we might use this feature in the future.  So I think it is better to use `offload_engine` to avoid ambiguity.

anHappyDog: done

guozhen1997 (approved):

PR #307, opened 2025-11-22
feat(embodiment): add single libero spatial task setting

zanghz21:
### Description

add single task option for libero spatial

### Motivation and Context

1. support quick debug for libero single task (and add a mlp yaml as well, just for end2end use, cannot train for reasonable results. )
2. maybe fix a bug for libero vector env by starting subprocess by "spawn"

### Types of changes

### Checklist:

### DIFF
--- a/docs/source-en/rst_source/start/vla.rst
+++ b/docs/source-en/rst_source/start/vla.rst
@@ -2,7 +2,7 @@ Quickstart 1: PPO Training of VLAs on Maniskill3
 =================================================
 
 This quick-start walks you through training the Visual-Language-Action model, including
-`OpenVLA <https://github.com/openvla/openvla>`_ on the
+`OpenVLA <https://github.com/openvla/openvla>`_ and `OpenVLA-OFT <https://github.com/moojink/openvla-oft>`_ on the
 `ManiSkill3 <https://github.com/haosulab/ManiSkill>`_ environment with **RLinf**.
 For maximum simplicity, you can run the following scripts within a single GPU.
 
@@ -32,6 +32,18 @@ For **OpenVLA**, run:
 
 the model is cited in `paper <https://arxiv.org/abs/2505.19789>`_
 
+For **OpenVLA-OFT**, run:
+
+.. code-block:: bash
+
+   # Download pre-trained OpenVLA-OFT model
+   hf download RLinf/Openvla-oft-SFT-libero10-trajall \
+   --local-dir /path/to/model/Openvla-oft-SFT-libero10-trajall/
+   
+   # Download lora fine-tuned ckpt in maniskill
+   hf download RLinf/RLinf-OpenVLAOFT-ManiSkill-Base-Lora \
+   --local-dir /path/to/model/oft-sft/lora_004000
+
 **Step 2: Execute the provided launch script:**
 
 .. note:: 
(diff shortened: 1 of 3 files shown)


zoeyuchao (approved):

examples/embodiment/config/libero_spatial_0_grpo_mlp.yaml
```diff
@@ -0,0 +1,187 @@
+# only for debug
```
QuanluZhang, line 1: what is the meaning of "only for debug"?

examples/embodiment/config/libero_spatial_0_grpo_mlp.yaml
```diff
@@ -0,0 +1,187 @@
...
+
+rollout:
+  group_name: "RolloutGroup"
+  channel:
+    name: ${env.channel.name}
+    queue_name: "action_buffer"
+    queue_size: 0
+  mode: "colocate"
+  generation_backend: "huggingface"
+  model_dir: "/mnt/public/mjwei/download_models/Haozhan72/Openvla-oft-SFT-libero-spatial-traj1/"
```
QuanluZhang: update this field to align with other yaml files

examples/embodiment/config/libero_spatial_0_grpo_mlp.yaml
```diff
@@ -0,0 +1,187 @@
...
+  pipeline_stage_num: 1
+
+actor:
+  group_name: "ActorGroup"
+  channel:
+    name: ${env.channel.name}
+    queue_name: "replay_buffer"
+    queue_size: 0
+  training_backend: "fsdp"
+  checkpoint_load_path: "/mnt/public/mjwei/download_models/Haozhan72/Openvla-oft-SFT-libero-spatial-traj1/"
```
QuanluZhang: same here

examples/embodiment/config/libero_spatial_grpo_openpi.yaml
```diff
@@ -35,6 +35,7 @@ algorithm:
   auto_reset: False
   ignore_terminations: False
   use_fixed_reset_state_ids: True
+  specific_reset_id: null
```
QuanluZhang: better to add comment to explain what is this field used for

rlinf/envs/libero/venv.py
```diff
@@ -129,7 +129,8 @@ def _encode_obs(
 
 class ReconfigureSubprocEnvWorker(SubprocEnvWorker):
     def __init__(self, env_fn: Callable[[], gym.Env], share_memory: bool = False):
-        self.parent_remote, self.child_remote = Pipe()
+        ctx = multiprocessing.get_context("spawn")
+        self.parent_remote, self.child_remote = ctx.Pipe()
```
QuanluZhang, line 133: is it slower than the preivous version?

zanghz21: not sure, but without this modification, in some cases the libero cannot run.

examples/embodiment/config/libero_spatial_0_grpo_mlp.yaml
zanghz21: this yaml can only run for end2end pipeline, and can never train to get a reasonable results because the input of the network is not enough for the policy to figure out the correct action.

XuS1994 (approved):

PR #318, opened 2025-11-25
fix: use independent response_mask in rollout result. attention_mask used only as valid tokens

qurakchin:
### Description

RLinf uses attention_mask to represent all valid prompt_ids and response_ids, but when using agent, we need to mask the tool response ids in adv function and loss function. In last version I used attention_mask and set tool response ids to 0 in attention_mask, but some code use attention_mask.sum() to calc sequence length, so it's wrong in agent.

In this version, I added a response_mask to represent llm generated response_ids, and attention_mask still represent all valid prompt_ids and response_ids, so it's ok to use attention_mask.sum() to calc sequence length, and in adv function and loss function, i will try to use response_mask to filter tool response ids. Then the bug is fixed.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/data/io_struct.py
+++ b/rlinf/data/io_struct.py
@@ -288,12 +288,15 @@ class RolloutResult:
 
     @staticmethod
     def _get_response_masks(
-        response_mask: list[list[int]], # [[0 / 1] * response len] * group size
+        response_mask: list[list[int]],  # [[0 / 1] * response len] * group size
         max_prompt_len: int,
         total_len: int,
     ) -> tuple[torch.Tensor, torch.Tensor]:
         response_mask = batch_pad_to_fixed_len(
-            [torch.as_tensor([0] * max_prompt_len + ids, dtype=torch.long) for ids in response_mask],
+            [
+                torch.as_tensor([0] * max_prompt_len + ids, dtype=torch.long)
+                for ids in response_mask
+            ],
             max_batch_len=total_len,
             pad_token=0,
         ).bool()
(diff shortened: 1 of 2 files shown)


qurakchin:
> What is the purpose of this PR? Could you add some descriptions in the PR description?

Thanks for the reminder. Added.

rlinf/data/io_struct.py
```diff
@@ -305,6 +283,39 @@ def _get_attention_masks_and_position_ids(
...
         return attention_mask, position_ids
 
+    @staticmethod
+    def _get_response_masks(
+        response_mask: list[list[int]],
+        prompt_lengths: torch.Tensor,
+        max_prompt_len: int,
+        total_len: int,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        max_response_len = total_len - max_prompt_len
```
Lin-xs: add doc

rlinf/data/io_struct.py
```diff
@@ -305,6 +283,39 @@ def _get_attention_masks_and_position_ids(
...
+        # Broadcast [B, total_len]
+        prompt_start = prompt_start.unsqueeze(1)
+
+        response_mask = batch_pad_to_fixed_len(
+            [torch.as_tensor(ids, dtype=torch.long) for ids in response_mask],
+            max_batch_len=max_response_len,
+            pad_token=0,
+        )
+
+        response_mask = torch.cat(
```
Lin-xs: delete prompt mask

rlinf/data/io_struct.py
```diff
@@ -740,7 +755,7 @@ def to_actor_batch(
                 shape ``[batch_size, training_seq_length - data_seq_length]``.
         """
 
-        # len = training_seq_length: input_ids, attention_mask, position_ids
+        # len = training_seq_length: input_ids, attention_mask, position_ids, response_mask
```
Lin-xs, line 752: add comments for response_mask

qurakchin: added.

rlinf/data/io_struct.py
qurakchin: now the shape is same as attention_mask. should not delete prompt part now.

rlinf/data/io_struct.py
qurakchin: added.

PR #428, opened 2025-12-11
chore: rename pi model.num_steps to model.denoising_steps; add comments

guozhen1997:
### Types of changes

### Checklist:

### DIFF
--- a/examples/embodiment/config/model/pi0_5.yaml
+++ b/examples/embodiment/config/model/pi0_5.yaml
@@ -6,6 +6,9 @@ model_path: "/path/to/model/openpi"
 precision: null # for openpi, precision is null, please do not change this value
 
 # Action and chunk settings
+# num_action_chunks: RLinf system-level setting that controls the number of action steps
+# for model-environment interaction. This value is used by the RLinf system
+# to determine how many action steps to execute per chunk_step.
 num_action_chunks: 10
 action_dim: 7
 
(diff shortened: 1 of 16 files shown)


examples/embodiment/config/model/pi0_5.yaml
```diff
@@ -15,18 +18,30 @@ lora_rank: 32
...
 # openpi specific parameters
 openpi:
   config_name: "pi05_libero"
   num_images_in_input: 2
-  noise_level: 0.5
+
+  # action_horizon: Controls the number of action steps that the model outputs.
+  # This parameter comes from openpi's Pi0Config and determines the length
+  # of the action sequence generated by the model.
+  action_horizon: 50
```
chenkang455: `action_horizon` is defined on the pi config files, we can not set them as 50 here. Models like maniskill, metaworld and calvin use `action_horizon` as 5 instead of 50 in libero. I recommend deleting them directly. 

image

chenkang455:
Hi @guozhen1997,

I noticed that the config settings in the openpi config files have been removed. 

image

In my workflow, I perform SFT using the official openpi repo with this config. Ideally, the model should be able to load this config directly for subsequent tasks like RL or Eval without any modifications. It is critical that these configs remain consistent after SFT.

chenkang455:
If this modification is to adapt to the sft code in RLinf, i recommend we use `sft_configs` and `rl_configs` to seperate them because `sft_configs` should be frozen during the RL stage.

examples/embodiment/config/calvin_d_d_ppo_openpi_pi05.yaml
```diff
@@ -156,6 +156,9 @@ actor:
       value_after_vlm: True
       value_vlm_mode: "mean_token"
 
+      action_horizon: 5
+      discrete_state_input: False
+
```
chenkang455: pi05: True ?

examples/embodiment/config/libero_10_grpo_openpi_pi05.yaml
```diff
@@ -143,6 +143,10 @@ actor:
   model:
     model_path: "/path/to/model/RLinf-Pi05-SFT"
 
+    openpi:
+      action_horizon: 10
+      discrete_state_input: False
+
```
chenkang455: pi05 True

PR #464, opened 2025-12-18
feat(embodiment): support SAC training for small models and support real world franka arm training

zanghz21:
### Description

- Support SAC-based off-policy algorithms for mlp and cnn. Support other variants (e.g. crossq, rlpd)
- Support async SAC training.
- Support peg insertion task in real world with Franka arm. 

### How has this been tested?

image
image

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/runners/embodied_runner.py
+++ b/rlinf/runners/embodied_runner.py
@@ -142,16 +142,6 @@ class EmbodiedRunner:
             # set global step
             self.actor.set_global_step(self.global_step)
             self.rollout.set_global_step(self.global_step)
-            eval_metrics = {}
-            if (
-                _step % self.cfg.runner.val_check_interval == 0
-                and self.cfg.runner.val_check_interval > 0
-            ):
-                with self.timer("eval"):
-                    self.update_rollout_weights()
-                    eval_metrics = self.evaluate()
-                    eval_metrics = {f"eval/{k}": v for k, v in eval_metrics.items()}
-                    self.metric_logger.log(data=eval_metrics, step=_step)
 
             with self.timer("step"):
                 with self.timer("sync_weights"):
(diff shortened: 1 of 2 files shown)


examples/embodiment/train_async.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    # Create rollout worker group
+    rollout_placement = component_placement.get_strategy("rollout")
+    rollout_group = AsyncMultiStepRolloutWorker.create_group(cfg).launch(
+        cluster, name=cfg.rollout.group_name, placement_strategy=rollout_placement
+    )
+
+    # Create env worker group
+    if cfg.cluster.get("enable_robot", False):
+        from rlinf.scheduler import NodePlacementStrategy
```
andylin-hao: The if can be removed now

examples/embodiment/config/env/realworld_peg_insertion.yaml
```diff
@@ -0,0 +1,29 @@
+simulator_type: realworld
```
zoeyuchao: charger的config还没有上来？

zoeyuchao: readme也需要对应修改。

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -890,11 +916,13 @@ def run_training(self):
         metrics = {}
         update_epoch = self.cfg.algorithm.get("update_epoch", 1)
         for _ in range(update_epoch):
-            rollout_dataloader_iter = get_iterator_k_split(
-                self.rollout_batch,
-                rollout_size // batch_size_per_rank,
+            rollout_global_batch_list = itertools.chain(
```
guozhen1997: I’m not sure in what concrete situations `self.rollout_batch` actually becomes a nested dict structure that justifies switching from `get_iterator_k_split` to `split_dict_to_chunk` here, and even if nested dicts do exist, it’s unclear whether recursively splitting all sub-dicts along the batch dimension is semantically correct for minibatching, since intuitively we might only want to split at the top-level batch dimension rather than blindly chunking every nested field.

Lin-xs (approved):

PR #474, opened 2025-12-21
fix(logprobs): add logprobs and entropy computation op type selection

anHappyDog:
Fix policy_loss is nan for former compute logprobs and entropy, especially for openvla series.

### Description

add logprobs and entropy computation op type selection. For OpenVLA series, it has short sequence length (`action_dim`), so it does not matter which op to choose(`liger_kernel`, `flash_attn`, `torch`), but for math reasoning , I found that `liger_kernel` has amazing effects on reducing VRAM cosumation and enhancing computation speed. I add a new optional param `entropy_op_type` in `algorithm` of yamls.

### Motivation and Context
Fix policy_loss is nan. And add liger_kernel as entropy computation ops.

### How has this been tested?
Tested in qwen1.5b/openvla series.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/utils/utils.py
+++ b/rlinf/utils/utils.py
@@ -257,7 +257,7 @@ def compute_logprobs_from_logits(
     labels = target.reshape(-1)
 
     assert op_type in ["torch", "flash_attn", "liger_kernel"], (
-        f"Unsupported op_type: {op_type} for logprobs computation."
+        f"Unsupported op_type: {op_type} for logprobs computation. Supported types are 'torch', 'flash_attn', 'liger_kernel'."
     )
     if op_type == "liger_kernel":
         logprobs = logprobs_from_logits_liger_kernel(logits, labels)


rlinf/models/embodiment/openvla_action_model.py
```diff
@@ -563,11 +563,11 @@ def forward(
...
-            action_logits = processed_logits_tensor.permute(
-                0, 2, 1
-            )  # [B, vocab-size, action-dim]
-            action_logits[:, : self.vocab_size - self.config.n_action_bins] = -torch.inf
-            action_logits[:, self.vocab_size :] = -torch.inf
+            action_logits = processed_logits_tensor
+            action_logits[
+                ..., : self.vocab_size - self.config.n_action_bins
+            ] = -torch.inf
+            action_logits[..., self.vocab_size :] = -torch.inf
```
anHappyDog, line 572: I need to explain why here needs permute but now we don't. In former implementation, -F.cross_entropy needs the class dim is dim 1, so it's just permuted (0,2,1). for current implementation, We just both reshape targets and labels to 2 dim to fix this problem.

rlinf/utils/utils.py
```diff
@@ -171,15 +214,39 @@ def logprobs_from_logits_flash_attn(logits, labels, inplace_backward=True):
     return -output[0]
 
 
+def logprobs_from_logits_liger_kernel(
+    logits: torch.Tensor, labels: torch.Tensor
+) -> torch.Tensor:
+    """
+    Compute logprobs by logits using liger-kernel's cross_entropy_loss.
```
anHappyDog, line 221: because I tested liger_kernel, flash_attn and torch, I found that liger_kernel has huge advantage over flash_attn and torch when long horizon used, I add liger_kernel to compute logprobs for math reasoning.

rlinf/utils/utils.py
```diff
@@ -188,14 +255,23 @@ def compute_logprobs_from_logits(
     last_dim = logits.shape[-1]
     logits = logits.reshape(-1, last_dim)
     labels = target.reshape(-1)
-    logprobs = logprobs_from_logits_flash_attn(
-        logits, labels=labels, inplace_backward=False
+
+    assert op_type in ["torch", "flash_attn", "liger_kernel"], (
+        f"Unsupported op_type: {op_type} for logprobs computation."
```
Lin-xs: Also print supported `op_type` for better readability.

rlinf/utils/utils.py
```diff
@@ -188,14 +255,23 @@ def compute_logprobs_from_logits(
...
     )
-    logprobs = logprobs.view(*batch_dim)
+    if op_type == "liger_kernel":
```
anHappyDog, line 262: For openvla series, we just use default option `torch`, for math, we use liger_kernel or flash_attn

rlinf/utils/utils.py
```diff
@@ -208,7 +284,10 @@ def compute_entropy_from_logits(logits, dim: int = -1) -> torch.Tensor:
         - entropy(torch.Tensor): [B, seq-len]
     """
     logp = F.log_softmax(logits, dim=dim)
-    entropy = -(logp * logp.exp()).sum(dim=dim)
+    p = logp.exp()
+    # if some p are zero, p*logp will be nan, we set those terms to zero
+    entropy_term = torch.where(p > 0, p * logp, 0.0)
```
anHappyDog, line 289: here if logp is -inf or inf, logp * p will be nan. so we just use torch.where to do mask.

anHappyDog:
maniskill ppo openvla with liger_kernel cross_entropy.image

maniskill ppo openvlaoft with liger_kernel cross_entropy.
image

for torch's cross_entropy. It's the same as former implementation.

image

image

andylin-hao (approved):

PR #508, opened 2025-12-29
feat(calvin): add more scene for benchmark CALVIN;

WinstonWmj:
add the two more tasks for CALVIN. They are "ACB-D" and "ABCD-D". Besides, I have fixed the bug for calvin eval.

### Description

firstly, i remove some useless yamls in calvin's dir ( `rlinf/envs/calvin/calvin_cfg/` ) and then i adjust the dir. Now is more clean and readable.
then, i add doc content for the relationship between D-D, ABC-D, and ABCD-D. They are different tasks( or say Challenges) in CALVIN. The only differences are the texture of the desk and posistion of static objects, and all other things are identical. (All there infomatios are digged from the arxiv and github issues)
next, i copy and do a little adjustment for the scene config ( placed under the path of `rlinf/envs/calvin/calvin_cfg/.hydra` ) the reason why i change the content is written in the doc (`docs/source-zh/rst_source/examples/calvin.rst`)
lastly, i fix the bug for evaluating the calvin using rlinf. While @chenkang455 using his own scripts to eval, so the CALVIN benchmarks's result(we reported in rlinf's doc) is still believable.

### How has this been tested?

NOW, the curves are lacked, the following experiments are running, please review first and the results will be updated in the next few days(about 4 * 1.5days, using 8 * H100)

### Types of changes

### Checklist:

### DIFF
--- a/examples/sft/config/robotwin_sft_openpi.yaml
+++ b/examples/sft/config/robotwin_sft_openpi.yaml
@@ -1,0 +1,78 @@
+defaults:
+  - model/pi0@actor.model
+  - training_backend/fsdp@actor.fsdp_config
+  - override hydra/job_logging: stdout
+
+hydra:
+  run:
+    dir: .
+  output_subdir: null
+  searchpath:
+    - file://${oc.env:EMBODIED_PATH}/config/
+
+cluster:
+  num_nodes: 1
+  component_placement:
+    actor,env,rollout: 0-0
+
+runner:
+  task_type: sft
+  logger:
+    log_path: "../results"
+    project_name: rlinf
+    experiment_name: "test_openpi"
+    logger_backends: ["tensorboard"] # wandb, swanlab
+
+  max_epochs: 1000
+  max_steps: -1
+  val_check_interval: -1
+  save_interval: 10
+
+data:
+  data_path: "/path/to/robotwin-data"
+
+algorithm:
+  adv_type: gae
+
+actor:
+  group_name: "ActorGroup"
+  training_backend: "fsdp"
+  micro_batch_size: 1
+  global_batch_size: 16
+  seed: 0
+
+  # Override the default values in model/pi0
+  model:
+    precision: null
+    model_path: "/path/to/pi0-model"
+    num_action_chunks: 4 # interface for the env
+    add_value_head: True
+    openpi:
+      config_name: "pi0_aloha_robotwin"
+      detach_critic_input: True
+
+  optim:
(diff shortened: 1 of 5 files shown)


rlinf/envs/calvin/__init__.py
```diff
@@ -17,83 +17,89 @@
...
+def _get_calvin_tasks_and_reward(
+    num_sequences, use_random_seed=True, task_suite_name="calvin_d"
+):
     conf_dir = (
         pathlib.Path(calvin_env.__file__).absolute().parents[2]
         / "calvin_models"
         / "conf"
     )
     task_cfg_path = conf_dir / "callbacks/rollout/tasks/new_playtable_tasks.yaml"
-    val_annotations_path = conf_dir / "annotations/new_playtable_validation.yaml"
```
chenkang455, line 35: At least one assert should exsist here?
```
if not conf_dir.exists():
    raise FileNotFoundError(
        f"Configuration directory {conf_dir} does not exist. "
        "Please ensure that the calvin_models package is installed correctly."
    )
```

rlinf/envs/calvin/__init__.py
```diff
@@ -17,83 +17,89 @@
...
 
 from rlinf.envs.calvin.utils import get_sequences
 
-ENV_CFG_DIR = Path(__file__).parent / "calvin_cfg/validation"
+ENV_CFG_DIR = Path(__file__).parent / "calvin_cfg/"
 
 
-def _get_calvin_tasks_and_reward(num_sequences, use_random_seed=True):
+def _get_calvin_tasks_and_reward(
+    num_sequences, use_random_seed=True, task_suite_name="calvin_d"
```
chenkang455: `use_random_seed` should be set to False during evaluation. Originally, this parameter was intended to differentiate between training and validation environments. In my own script, I will set the default seed, i.e., ``use_random_seed = False``.

rlinf/envs/calvin/__init__.py
WinstonWmj: good suggestion! i havd added.

rlinf/envs/calvin/__init__.py
WinstonWmj: After review the source, i think this ifelse is reduntant since the reset_ids is randomly set, so we don't need another seed here.

guozhen1997 (approved):

zoeyuchao (approved):

PR #548, opened 2026-01-09
feat: add fsdp convertor, update docs and yamls

guozhen1997:
### Description

- Add FSDP checkpoint convertor
- Update docs and yamls
- Update eval docs

### Types of changes

### Checklist:

### DIFF
--- a/docs/source-en/rst_source/start/vla-eval.rst
+++ b/docs/source-en/rst_source/start/vla-eval.rst
@@ -1,5 +1,5 @@
-How to evaluate? Embodied Agent Scenarios
-============================================
+Evaluation Tutorial 1: Embodied VLA
+====================================
 
 Introduction
 ------------
@@ -40,10 +40,11 @@ Any YAML file can be used for evaluation with the ``eval_embodiment.sh`` script,
 
   1. ``rollout.model.model_path``
 
-  2. ``actor.model.model_path``
+  2. ``runner.ckpt_path`` (Optional) – ``.pt`` format file path, set this parameter if you want to evaluate a specific checkpoint.
 
-  3. ``runner.ckpt_path`` (Optional) – Set this parameter if you want to evaluate a specific checkpoint.
-
+.. Note::
+  
+   If you need to convert checkpoint files from ``.distcp`` format to ``.pt`` format, please refer to the :doc:`Checkpoint Convertor <../tutorials/advance/convertor>` documentation for detailed instructions.
 
 2. **Control environment random seed**: You can adjust ``env.seed`` to change the environment's random function for result reproducibility, etc.
 
--- a/docs/source-en/rst_source/examples/searchr1.rst
+++ b/docs/source-en/rst_source/examples/searchr1.rst
@@ -134,11 +134,11 @@ Run `examples/searchr1/run_main_searchr1_single.sh` to start training.
 Evaluation
 ----------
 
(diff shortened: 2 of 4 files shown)


docs/source-zh/rst_source/examples/searchr1.rst
```diff
@@ -129,12 +129,12 @@ Wiki配置文件
 测试
 ----
 
-运行toolkits/ckpt_convertor/mg2hf_3b.sh将megatron
+运行toolkits/ckpt_convertor/megatron_convertor/mg2hf_3b.sh将megatron
```
QuanluZhang, line 132: why there no corresponding update in the english doc?

docs/source-en/rst_source/start/vla-eval.rst
```diff
@@ -42,6 +42,8 @@ Any YAML file can be used for evaluation with the ``eval_embodiment.sh`` script,
 
   2. ``actor.model.model_path``
 
+  3. ``runner.ckpt_path`` (Optional) – Set this parameter if you want to evaluate a specific checkpoint.
```
QuanluZhang: what is the difference between rollout's model_path, actor's model_path, and ckpt_path? according to me, they are the same?

QuanluZhang: accoding to my understanding, model_path has the model's original weights, while ckpt_path has checkpointed weights. but, what's the difference between actor's model_path and rollout's model_path?

docs/source-en/rst_source/tutorials/advance/convertor.rst
```diff
@@ -0,0 +1,133 @@
...
+3. ``CKPT_PATH_ORIGINAL_HF`` (Base model checkpoint for initial training, e.g., ``/path/to/DeepSeek-R1-Distill-Qwen-1.5B``)
+
+**Method 2: Command-line arguments**
+
+A more flexible way is to pass paths directly through command-line arguments.
+
+.. code-block:: bash
+
+    bash mg2hf_1.5b.sh /path/to/megatron_checkpoint /target/path/to/huggingface_checkpoint /path/to/base_model_checkpoint
+
```
QuanluZhang, line 133: very clear!

docs/source-en/rst_source/start/vla-eval.rst
andylin-hao: Rollout's model path and the actor's model path are usually set the same in current config.

guozhen1997: Both the actor’s model_path and the rollout’s model_path refer to Hugging Face model paths. For evaluation, only the rollout’s model_path needs to be set. I will also update the documentation to clarify this. And The ckpt_path refers to the saved checkpoint path, which contains the complete set of model parameters.

When LoRA is enabled, PEFT modifies the model parameter keys. Previously, I attempted to save the base weights and LoRA weights separately in Hugging Face format so that ckpt_path would not be required. However, the saved LoRA adapters could not be loaded correctly in that approach.

As a result, I switched to saving the full model weights. After building the PeftModel, we then call load_state_dict to restore the checkpoint parameters, which allows the previously saved checkpoint to be loaded correctly.

QuanluZhang: ok

docs/source-zh/rst_source/examples/searchr1.rst
guozhen1997: This was missed earlier. I’ve now added the corresponding update to the English doc.

PR #553, opened 2026-01-12
feat: refactor fsdp and support fsdp backend in agent

qurakchin:
### Description

fixes:
1. added a workaround to the get_iterator_dynamic because the result after get_seqlen_balanced_partitions may exceed the max_tokens_per_mbs.
2. fixed using amp_context in inference
3. detected logits.dtype and set to torch.float32 force
4. moved the loss calc process out of amp_context. then the loss is calculated in torch.float32
5. using the final_loss_metric generated before the loss divided by self.gradient_accumulation. the metrics of policy_loss now is aligned with megatron version
6. moved the lr_scheduler.step after optimizer.step() in every n_minibatch, rather than in rl every step. aligned with megatron version
7. fied the "actor/ratio" in early stopping process. or the metrics value is incorrect.

features:
1. added a search r1 fsdp config that the loss curve is aligned with megatron config
2. added 2 metrics: metric_policy_loss_abs and ratio_abs_for_metrics
3. supported using grad_scaler in fsdp in fp16 by add a _unscale_patch to patch the original GradScaler.unscale_
4. added the enable_dynamic_batch_size feature in fsdp. supported in both inference and training. supported auto dynamic batch in inference in pipeline mode like megatron.
5. added the enable_dp_load_balance feature in fsdp
6. supported to sync weight with several bucket in fsdp

refactors:
1. search r1 megatron config use mix precision training to align the loss curve to the fsdp config
2. a small refactor to a assert of bucket_length > 0 in sgl_scheduler
3. refactored the mean_metric_dict calc in training_step function.

reward curve in searchr1 between megatron and fsdp:
image

step time in searchr1 between megatron and fsdp:
image

after 100 step: megatron ~56s, fsdp ~53s

reward curve in math between megatron and fsdp:
image
 
### Types of changes

### Checklist:

### DIFF
--- a/docs/source-en/rst_source/examples/calvin.rst
+++ b/docs/source-en/rst_source/examples/calvin.rst
@@ -43,7 +43,7 @@ Environment
 
 .. note::
 
-   Please note that we have modified the YAML files for scene A and scene C here. The original repository calvin contained some incorrect settings for these two configuration files, which we have corrected in RLinf. You can use them with confidence. See this `issue <https://github.com/mees/calvin/issues/41>`_ for details.
+   Please note that we have modified the YAML files for scene A, scene B and scene C here. The original repository calvin contained some incorrect settings for these two configuration files, which we have corrected in RLinf. You can use them with confidence. See this `issue <https://github.com/mees/calvin/issues/41>`_ for details.
 
 Algorithm
 ---------
--- a/docs/source-en/rst_source/examples/pi0.rst
+++ b/docs/source-en/rst_source/examples/pi0.rst
@@ -390,7 +390,7 @@ For models after SFT or RL training, we provide two evaluation methods:
 
 .. note::
 
-   ``Metaworld`` and ``CALVIN`` currently do not support the evaluation mode with ``env.eval.auto_reset=True``. It is recommended to use individual script files for model evaluation.
+   ``Metaworld`` currently do not support the evaluation mode with ``env.eval.auto_reset=True``. It is recommended to use individual script files for model evaluation.
 
 - Use individual script files for model evaluation, refer to the example `README.md <https://github.com/RLinf/RLinf/blob/main/toolkits/eval_scripts_openpi/README.md>`__. This method's evaluation scripts are consistent with the official evaluation scripts provided by ``openpi``, supporting output of success rates for each subtask, but it is slower.
 
--- a/docs/source-zh/rst_source/examples/calvin.rst
+++ b/docs/source-zh/rst_source/examples/calvin.rst
@@ -37,7 +37,7 @@
 
 .. note::
 
-   注意，这里我们修改了其中的 scene A 和 scene C 的yaml文件，因为原本的Calvin仓库中对于这两个配置文件有一些错误的设置，我们在rlinf中已经修正，大家可放心使用。参见这个 `问题 <https://github.com/mees/calvin/issues/41>`_。
+   注意，这里我们修改了其中的 scene A, scene B 和 scene C 的yaml文件，因为原本的Calvin仓库中对于这两个配置文件有一些错误的设置，我们在rlinf中已经修正，大家可放心使用。参见这个 `问题 <https://github.com/mees/calvin/issues/41>`_。
 
 **数据结构**
 
--- a/docs/source-zh/rst_source/examples/pi0.rst
+++ b/docs/source-zh/rst_source/examples/pi0.rst
@@ -358,7 +358,7 @@ env** 之间的流水线重叠，从而提升 rollout 效率。
 
 .. note::
 
-   ``Metaworld`` 以及 ``CALVIN`` 暂时不支持设置 ``env.eval.auto_reset=True`` 的评估模式，建议使用单个脚本文件进行模型评估。
+   ``Metaworld`` 暂时不支持设置 ``env.eval.auto_reset=True`` 的评估模式，建议使用单个脚本文件进行模型评估。
 
 - 使用单个脚本文件进行模型评估，参考示例 `README.md <https://github.com/RLinf/RLinf/blob/main/toolkits/eval_scripts_openpi/README.md>`__，这种方式的评估脚本和 ``openpi`` 官方提供的评估脚本一致，支持输出每个子任务的成功率，但是速度较慢。
 
--- a/rlinf/envs/calvin/calvin_cfg/.hydra/calvin_scene_A.yaml
+++ b/rlinf/envs/calvin/calvin_cfg/.hydra/calvin_scene_A.yaml
@@ -41,11 +41,11 @@ objects:
       file: blocks/block_red_middle.urdf
       initial_pos: any
       initial_orn: any
-    block_pink:
-      file: blocks/block_pink_small.urdf
-      initial_pos: any
-      initial_orn: any
     block_blue:
       file: blocks/block_blue_big.urdf
+      initial_pos: any
+      initial_orn: any
+    block_pink:
+      file: blocks/block_pink_small.urdf
       initial_pos: any
       initial_orn: any
--- a/rlinf/envs/calvin/calvin_cfg/.hydra/calvin_scene_B.yaml
+++ b/rlinf/envs/calvin/calvin_cfg/.hydra/calvin_scene_B.yaml
@@ -37,12 +37,12 @@ objects:
           link: led_link
           color: [0, 1, 0, 1] # green
   movable_objects:
-    block_red:
-      file: blocks/block_red_small.urdf
-      initial_pos: any
-      initial_orn: any
     block_blue:
       file: blocks/block_blue_big.urdf
+      initial_pos: any
+      initial_orn: any
+    block_red:
+      file: blocks/block_red_small.urdf
       initial_pos: any
       initial_orn: any
     block_pink:
--- a/rlinf/envs/calvin/calvin_gym_env.py
+++ b/rlinf/envs/calvin/calvin_gym_env.py
@@ -62,6 +62,7 @@ class CalvinEnv(gym.Env):
         self.reset_state_ids_all = self.get_reset_state_ids_all()
         self.update_reset_state_ids()
         self._init_task_and_trial_ids()
+        self._init_task_info()
         self._init_env()
 
         self.prev_step_reward = np.zeros(self.num_envs)
@@ -137,6 +138,13 @@ class CalvinEnv(gym.Env):
             self._get_task_and_trial_ids_from_reset_state_ids(self.reset_state_ids)
         )
 
+    def _init_task_info(self):
+        self.task_sequence = [None] * self.num_envs
+        self.current_task = [None] * self.num_envs
+        self.current_task_idx = [0] * self.num_envs
+        self.previous_info = [None] * self.num_envs
+        self.task_descriptions = [None] * self.num_envs
+
     def _get_random_reset_state_ids(self, num_reset_states):
         reset_state_ids = self._generator.integers(
             low=0, high=self.total_num_group_envs, size=(num_reset_states,)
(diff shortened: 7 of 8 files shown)


examples/searchr1/config/qwen2.5-3b-tool-1node.yaml
```diff
@@ -226,7 +228,7 @@ actor:
 
     # megatron config
     bf16: False
-    fp16: True
+    fp16: False
```
QuanluZhang: why do you use `bf16` and `fp16` configuration fields, instead of using `precision` field? What if both `bf16` and `fp16` are True?

QuanluZhang:
@anHappyDog @Lin-xs as primary reviewer

examples/searchr1/config/qwen2.5-3b-tool-1node.yaml
qurakchin: The bf16 and fp16 there will only effect the grad scaler in megatron. The process of precision config now contains several bugs and I will add another pr to refactor precision config in megatron and fsdp.

rlinf/utils/utils.py
```diff
@@ -253,7 +253,7 @@ def logprobs_from_logits_liger_kernel(
 def compute_logprobs_from_logits(
     logits: torch.Tensor,
     target: torch.Tensor,
-    op_type: Literal["torch", "flash_attn", "liger_kernel"] = "torch",
+    op_type: Literal["torch", "flash_attn", "liger_kernel"] = "flash_attn",
```
anHappyDog: If so, embodied models will also use `flash_attn` instead of `torch`, do you have tests on them?

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -181,7 +197,23 @@ def __init__(
         self.micro_batch_size = self.cfg.actor.micro_batch_size
         self.n_mini_batches = self.cfg.algorithm.n_minibatches
         self.task_type = self.cfg.runner.task_type
-        self.entropy_op_type = self.cfg.algorithm.get("entropy_op_type", "liger_kernel")
+        self.entropy_op_type = self.cfg.algorithm.get("entropy_op_type", "flash_attn")
```
anHappyDog, line 185: in former PR, I found `liger_kernel` here uses much less VRAM and keeps same precision with `flash_attn` in the scenario of long context.  does `liger_kernel` have some side effects?

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -322,6 +383,102 @@ def get_batch(
         )
         return batch, result
 
+    def all_reduce_dp_min(
```
anHappyDog: maybe these functions can be put into another utils file and used by FSDP actor or other classes. Cause the other fsdp actor classes are inherited from this class, while these classes may do not use them at all.

rlinf/hybrid_engines/fsdp/utils.py
```diff
@@ -587,3 +587,140 @@ def get_backward_prefetch_strategy(
         f"Unknown backward prefetch strategy: {prefetch_str}"
     )
     return BACKWARD_PREFETCH_STRATEGIES[prefetch_str]
+
+
+def pack_sequences(
```
anHappyDog, line 592: could it be ok to add docstring for them? Have unit tests added for them in case future modifications?

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -380,48 +575,123 @@ def run_inference(
             output_channel: The output channel to send results to.
             compute_ref_logprobs: Whether to compute reference logprobs.
         """
-        recv_batch_size = 0
-        while recv_batch_size < self.total_batch_size_per_dp:
-            batch, rollout_result = self.get_batch(input_channel)
-            recv_batch_size += rollout_result.num_sequence
-            self._load_weight_and_optimizer()
+        inference_split = self.cfg.actor.get("inference_split", None)
```
anHappyDog, line 623: this function is too long for nearly 150 lines, could it be better to extract some part logics into helper functions?

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -561,7 +865,22 @@ def iterator_wrapper():
             append_to_dict(mbs_metrics_list, mbs_metrics_data)
 
         grad_norm, lr_list = self.optimizer_step()
-        return mbs_metrics_list, grad_norm, lr_list
+
+        # put lr scheduler step here
+        self.lr_scheduler.step()
```
anHappyDog: Does putting lr_scheduler here make better converging curves? I remember @guozhen1997  said lots of frameworks put lr_scheduler where the global batch ends.

rlinf/workers/actor/fsdp_actor_worker.py
qurakchin: Tested that if input.dtype is fp16, liger_kernel use fp16 but flash_attn use fp32 to calc. So flash_attn has better precision than liger_kernel.

rlinf/utils/data_iter_utils.py
```diff
@@ -523,25 +524,46 @@ def get_iterator_dynamic(
...
+                        continue
+                microbatches.append(curr_batch)
+                if (
+                    sum(curr_batch["prompt_lengths"])
+                    + sum(curr_batch["response_lengths"])
+                    > max_tokens_per_mbs
+                ):
+                    valid_max_token = 0
+                    print(
+                        "error 1! max len > max_tokens_per_mbs! "
```
Lin-xs: Here we can print more precise error info, rather than 'error 1' or 'error 2'.

rlinf/utils/utils.py
qurakchin: reverted. only reasoning use flash_attn as default.

rlinf/workers/actor/fsdp_actor_worker.py
qurakchin: refactored.

rlinf/hybrid_engines/fsdp/utils.py
qurakchin: added.

rlinf/workers/actor/fsdp_actor_worker.py
qurakchin: added a config "actor.lr_sched_sync_with_optim" to chose where to run lr_scheduler.step()

rlinf/utils/data_iter_utils.py
qurakchin: fixed.

rlinf/workers/actor/fsdp_actor_worker.py
qurakchin: refactored.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -269,44 +339,39 @@ def sync_model_to_rollout(self) -> None:
...
-                    # if transformers version is 4.56.1 or older(not tested),
-                    # the following line should be uncommented
-
-                    # elif name.startswith("model."):
-                    #     name = name[6:]
-                state_dict[name] = reduce_tensor(v) if not self.is_pipeline else v
-            if not self.is_pipeline:
+        if len(self._weight_dst_rank_in_rollout) > 0:
+            rollout_dtype = None
+            if self._cfg.get("sync_precision", None) is not None:
```
anHappyDog, line 335: sglang/vllm's weight loading function uses `copy_`, which can automatically handle precision conversion.  is there a specific reason for the  precision conversion here?

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -269,44 +339,39 @@ def sync_model_to_rollout(self) -> None:
...
+                rollout_dtype = torch_dtype_from_precision(self._cfg.sync_precision)
+            model_bucket_list = self.divide_model_to_bucket(
+                self.rollout_state_dict, has_visual
+            )
+            print(f"length of model_bucket_list: {len(model_bucket_list)}")
```
anHappyDog: it's better to use logger rather than print.

anHappyDog (approved):

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -269,44 +339,39 @@ def sync_model_to_rollout(self) -> None:
...
+            for bucket_idx, model_bucket in enumerate(model_bucket_list):
+                buffer = {}
+                for k, v in model_bucket.items():
+                    if isinstance(v, DTensor):
```
anHappyDog, line 346: It's a little confusing,  in former code, rollout_state_dict is got with param `full_state_dict` set False and it should return dict paired with str and DTensor(If I remember correctly), but here calls `full_tensor`,  which will turn it into full tensor. I forgot whether sglang/vllm could handle sharded weight from FSDP, If so, it might be better to use `to_local` function. If not, it might be better to set former param `full_state_dict` True.

rlinf/workers/actor/fsdp_actor_worker.py
qurakchin: Doing precision conversion before weight sync can reduce the communication cost. Tested ~1.5s time saving in 3b model.

rlinf/workers/actor/fsdp_actor_worker.py
qurakchin: 1. In the case of an env with 8 gpus, if we use to_local function, we can only get a 1 / 8 weight of a tensor, and sglang cannot load the split tensor. We should send a full tensor so that sglang can load correctly.
2. If the sync_precision / send in bucket / allgather in every bucket are off, send weight process will cost many gpu memory. Allgather in every bucket can reduce the gpu memory peak.

rlinf/workers/actor/fsdp_actor_worker.py
qurakchin: fixed.

Lin-xs (approved):

rlinf/utils/utils.py
```diff
@@ -277,10 +278,13 @@ def compute_logprobs_from_logits(
         f"Unsupported op_type: {op_type} for logprobs computation. Supported types are 'torch', 'flash_attn', 'liger_kernel'."
     )
     if op_type == "liger_kernel":
+        # liger_kernel will use input dtype to compute logprobs.
         logprobs = logprobs_from_logits_liger_kernel(logits, labels)
     elif op_type == "flash_attn":
+        # flash_attn will use fp32 to compute logprobs. so default use flash_attn
```
guozhen1997: default is still torch, maybe you can remove the last sentence.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -237,6 +282,39 @@ def sync_model_to_inference(self) -> None:
 
         torch.distributed.barrier()
 
+    def divide_model_to_bucket(self, state_dict, has_visual):
```
guozhen1997, line 285: Can we make the bucket sync weights general, and move the coresponding functions into FSDPModelManager?  Embodied training can also benefit from it.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -427,22 +726,15 @@ def iterator_wrapper():
         for idx, m_batch in enumerate(micro_batches_iter):
             backward_ctx = self.before_micro_batch(
                 self.model,
-                is_last_micro_batch=(idx + 1) == self.gradient_accumulation,
+                is_last_micro_batch=(idx + 1) == micro_batch_cnt,
```
guozhen1997, line 729: Perhaps we could revert back to `self.gradient_accumulation` here, which would maintain consistency with the loss scaling in line L793.

qurakchin: use self.gradient_accumulation is not correct with the dynamic batch enabled.

rlinf/utils/utils.py
qurakchin: fixed.

rlinf/workers/actor/fsdp_actor_worker.py
qurakchin: I think the refactor and the test in embodied can be put in another pr.

PR #649, opened 2026-01-31
feat: make install.sh reuse existing venv instead of exiting

111XiaoLi666:
### Description
This PR makes the installer idempotent by reusing an existing virtual environment instead of exiting when the venv directory already exists.

Concretely, create_and_sync_venv() now:
	•	Checks whether $VENV_DIR/bin/activate exists.
	•	If it exists, it reuses the venv and continues.
	•	Otherwise, it creates a new venv via uv venv ....
	•	Then it proceeds with uv sync --active as before.

### Motivation and Context
With set -euo pipefail, the current script aborts when uv venv detects an existing venv and the user declines replacement (uv exits non-zero). This makes rerunning the installer (e.g., after interrupted downloads / flaky networks) impossible unless the user manually deletes the venv.

This change allows rerunning the installer to continue dependency installation/sync without requiring a venv rebuild.

### How has this been tested?
Manual test:
	1.	Ran bash requirements/install.sh embodied --model openvla-oft --env maniskill_libero --venv <existing_venv_path> with an already-existing venv.
	2.	Confirmed the script reuses the venv and proceeds to uv sync --active instead of exiting.

Environment: Linux (AutoDL container), uv 0.9.27.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/data/replay_buffer.py
+++ b/rlinf/data/replay_buffer.py
@@ -18,7 +18,6 @@ import json
 import os
 import pickle as pkl
 import threading
-from collections import OrderedDict
 from concurrent.futures import ThreadPoolExecutor
 from typing import Optional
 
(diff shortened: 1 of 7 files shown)


andylin-hao:
@111XiaoLi666 Thanks for your suggestion! I'll make a few edits to this PR this week for issues like cloning into existing folders. Stay tuned!

andylin-hao:
Hi @111XiaoLi666, venv reuse is now fully supported for both venv and Git repo in install.sh (#723). Thanks for your suggestion!

PR #653, opened 2026-02-02
feat: add multi thread in dataset post process

qurakchin:
### Description

The math dataset post process is too slow. After using multi thread, post process time reduced from ~120s to ~13s in a 140m searchr1 jsonl dataset.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -13,6 +13,7 @@ jobs:
       - name: Create environment and install dependencies
         run: |
           uv venv
+          source .venv/bin/activate
           UV_TORCH_BACKEND=auto uv sync
           uv pip install transformers
 


rlinf/data/datasets/math.py
```diff
@@ -75,29 +76,43 @@ def __init__(
...
-                        # save the convert data
-                        item[self.prompt_key] = prompt
-                    _, L = self.encode(prompt)
-                    if L <= self.max_prompt_length:
-                        filtered.append(item)
-                except Exception:
-                    failed += 1
-
-            self.data = filtered
+            batch_size = 256
```
F9rozen: Why is it a fixed number？

qurakchin: fixed by adding a configable batch size

rlinf/data/datasets/math.py
```diff
@@ -109,6 +129,31 @@ def __init__(
                     f"(kept {len(self.data)} / {total})."
                 )
 
+    def _load_post_process(self, batch):
+        result = batch
+        failed = 0
+        try:
+            prompts = (item[self.prompt_key] for item in batch)
+            if self.apply_chat_template:
+                prompts = self.tokenizer.self.apply_chat_template(
```
F9rozen: double self?

rlinf/data/datasets/math.py
```diff
@@ -75,29 +76,48 @@ def __init__(
...
+                    ]
+                    for handle in handles:
+                        result, failed = handle.result()
+                        self.data.extend(result)
+                        all_failed += failed
+            else:
+                for batch in batches:
+                    assert tokenizer.is_fast
+                    result, failed = self._load_post_process(batch)
+                    result, failed = self._load_post_process(
```
F9rozen: Why is this function called twice, and why are the arguments different?

rlinf/data/datasets/math.py
```diff
@@ -109,6 +129,31 @@ def __init__(
...
+        try:
+            prompts = (item[self.prompt_key] for item in batch)
+            if self.apply_chat_template:
+                prompts = self.tokenizer.self.apply_chat_template(
+                    prompts, tokenize=False, add_generation_prompt=True
+                )
+                for item, prompt in zip(batch, prompts):
+                    item[self.prompt_key] = prompt
+            if self.filter_prompt_by_length:
+                prompt_ids = self.tokenizer.batch_encode_plus(list(prompts))[
```
F9rozen, line 145: apply_chat_template=True and filter_prompt_by_length=True，prompt_ids maybe none？they are using same prompts

rlinf/data/datasets/math.py
```diff
@@ -75,29 +76,48 @@ def __init__(
...
-                    if L <= self.max_prompt_length:
-                        filtered.append(item)
-                except Exception:
-                    failed += 1
-
-            self.data = filtered
+            batch_size = self.process_batch_size
+            batches = [
+                self.data[i : i + batch_size]
+                for i in range((total + batch_size - 1) // batch_size)
```
FxxxxU: Why is it handled this way here? Wouldn’t this cause duplicate data to be read? I think here should be i * batch_size : (i +1) *batch_size.

rlinf/data/datasets/math.py
qurakchin: fixed.

rlinf/data/datasets/math.py
qurakchin: fixed.

rlinf/data/datasets/math.py
qurakchin: fixed and added a unit test.

rlinf/data/datasets/math.py
qurakchin: prompt_ids won't be none, and when filter_prompt_by_length is enabled, it will delete all the items which prompt_ids length is over the max prompt length.

andylin-hao (approved):

PR #654, opened 2026-02-02
feat(async_rl): support async PPO training for embodiment models

anHappyDog:
### Description
support async PPO training for embodiment models. The decoupled PPO algorithms comes from [AReal](https://github.com/inclusionAI/AReaL)

### Motivation and Context
Decouple training and rollout.

### How has this been tested?
Tested with maniskill_async_ppo_openvla for 4 GPUs. while ppo uses config env/rollout/actor: 0-3, async ppo uses config env/rollout: 0-1, actor: 2-3, **speedup 18%** or so. maybe there exists a better GPU placements for higher speedup.

image

Tested with maniskill_async_ppo_openvlaoft for 1 GPU, **speedup 11%** orso.

image

Tested with libero_spatial_async_ppo_openpi for 4 GPU, both ppo and async ppo uses config env/rollout/actor: 0-3, **speedup 38%**, but async ppo's converged success_once  curve is a little lower than ppo.

image

Tested with libero_spatial_async_ppo_openpi_pi05 for 4 GPU:   both ppo and async ppo uses config env/rollout/actor: 0-3, **speedup 12%**.

image

Tested with libero_goal_async_ppo_gr00t for 4 GPU: both ppo and async ppo uses config env/rollout/actor: 0-3, **speedup 9%**.

image

The extreme speedup comes from overlap between training and (env-rollout) . we could calculate the extreme speedup rate from former ppo's step time and training time. Also, experiments above set `update_epch` to 1, if set larger, the speedup must be higher, and if batch is larger, the speedup must be higher.

### Types of changes

### Checklist:


rlinf/workers/rollout/hf/async_ppo_huggingface_worker.py
```diff
@@ -0,0 +1,221 @@
...
+        except asyncio.CancelledError:
+            pass
+
+    async def _generate(
+        self,
+        input_channel: Channel,
+        output_channel: Channel,
+        actor_channel: Channel,
+    ) -> None:
+        assert not self.enable_offload, (
```
Lin-xs: It is better to raise Exception in __init__().

rlinf/runners/async_ppo_embodied_runner.py
```diff
@@ -0,0 +1,165 @@
...
+
+            self.print_metrics_table_async(
+                self.global_step - 1,
+                self.max_steps,
+                start_time,
+                logging_metrics,
+                start_step,
+            )
+
+            _, save_model, _ = check_progress(
```
Lin-xs, line 181: Evaluation is effectively disabled even when runner.val_check_interval > 0, so async configs that request periodic eval silently skip eval and metrics, which I think is not expected.

rlinf/algorithms/losses.py
```diff
@@ -21,6 +21,139 @@
...
+            generated_tokens_mask = versions >= 0
+            alpha = torch.where(
+                (version_diff > 0) & generated_tokens_mask,
+                version_gap / version_diff,
+                torch.zeros_like(v_behav),
+            )
+            while alpha.dim() < logprobs.dim():
+                alpha = alpha.unsqueeze(-1)
+            alpha = torch.clamp(alpha, 0.0, 1.0)
+            proximal_logprobs = (
```
Lin-xs, line 84: Decoupled PPO actor loss currently ignores proximal_logprobs in the core policy ratio, so training objective behaves like standard PPO ratio against behavior policy instead of proximal policy.

Lin-xs:
New decoupled_actor_critic path is not validated for required value head, unlike existing actor_critic, so invalid configs can pass validation and fail later at runtime.

```python
# rlinf/config.py
    # NOTE: Currently we only support actor_critic as PPO algorithm loss, and only support value_head as critic model.
    # This will be updated in the future to support more algorithms and critic models.
    # Check that actor_critic loss requires value_head
    if cfg.algorithm.loss_type == "actor_critic":
        add_value_head = cfg.actor.model.get("add_value_head", False)
        assert add_value_head, (
            f"When using PPO algorithm (algorithm.loss_type='actor_critic'), "
            f"actor.model.add_value_head must be True. "
            f"Current value: {add_value_head}"
```

rlinf/algorithms/losses.py
anHappyDog: done.

rlinf/runners/async_ppo_embodied_runner.py
anHappyDog: warning for cases where val_check_interval > 0 is added. evaluation may be added in next PR.

rlinf/workers/env/async_env_worker.py
```diff
@@ -16,79 +16,51 @@
...
-                        rewards=None,
-                        dones=dones,
-                        terminations=terminations,
-                        truncations=truncations,
-                        intervene_actions=self.last_intervened_info_list[i][0],
-                        intervene_flags=self.last_intervened_info_list[i][1],
-                    )
-                    env_output_list.append(env_output)
-
+            env_output_list = self.init_env_outputs()
```
andylin-hao: Rename init_env_outputs to bootstrap_step

anHappyDog: done.

examples/embodiment/run_async_ppo.sh
```diff
@@ -0,0 +1,43 @@
+#! /bin/bash
```
XuS1994: rename to `run_async_embodiment.sh` is better

examples/embodiment/train_async_ppo.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2025 The RLinf Authors.
```
XuS1994: `train_async_ppo.py` and `train_async.py` should be merged into one file.

examples/embodiment/run_async_ppo.sh
anHappyDog: Now `run_async_ppo.sh` is merged into former bash file `run_async.sh`.

examples/embodiment/train_async_ppo.py
anHappyDog: done.

XuS1994 (approved):

PR #657, opened 2026-02-02
feat(torch_compile): support torch compile for embodiment models

anHappyDog:
### Description

support torch compile for cnn/mlp models.

### How has this been tested?
Test CI files.

maniskill_ppo_mlp:

image

frankasim_sac_cnn_async:

image

### Types of changes

### Checklist:


examples/embodiment/config/maniskill_ppo_openvla.yaml
```diff
@@ -107,7 +107,7 @@ rollout:
   backend: "huggingface"
   enable_offload: True
   pipeline_stage_num: 2
-
+  enable_torch_compile: True
```
andylin-hao: It'd be better not to change the default behavior of examples. Add this to one of the CI tests instead.

rlinf/models/embodiment/cnn_policy/cnn_policy.py
```diff
@@ -80,7 +80,7 @@ def _update_info(self):
 class CNNPolicy(nn.Module, BasePolicy):
     def __init__(self, cfg: CNNConfig):
         super().__init__()
-
+        BasePolicy.__init__(self)
```
andylin-hao: This is unnecessary

rlinf/models/embodiment/flow_policy/flow_policy.py
```diff
@@ -392,6 +394,12 @@ def predict_action_batch(
             result["shared_feature"] = visual_feature
         return chunk_actions, result
 
+    def enable_torch_compile(self):
```
andylin-hao: Can we move this to base policy instead of writing it every time for an unsupported model?

rlinf/models/embodiment/openvla/openvla_action_model.py
```diff
@@ -663,35 +838,50 @@ def predict_action_batch(
         assert torch.all(input_ids[:, -1] == 29871)
         assert torch.all(attention_mask[:, -1] == 1)
 
-        generated_results: GenerateDecoderOnlyOutput = self.generate(
-            input_ids,
+        # generated_results: GenerateDecoderOnlyOutput = self.generate(
```
andylin-hao: Remove commented code

rlinf/models/embodiment/openvla/openvla_action_model.py
```diff
@@ -806,3 +996,54 @@ def setup_config_and_processor(self, model_config, input_processor):
...
+                input_ids=input_ids,
+                attention_mask=attention_mask,
+                pixel_values=pixel_values,
+                use_cache=True,
+                output_hidden_states=False,
+                return_dict=True,
+            )
+            return out.past_key_values, out.logits[:, -1, :]
+
+        # self._prefill_fn_compiled = torch.compile(
```
andylin-hao: Remove commented code

examples/embodiment/config/maniskill_ppo_openvla.yaml
anHappyDog: first running with torch compile consumes some time, for openvla, it nearly costs 5 minutes, do we still need to add ci for it?

rlinf/models/embodiment/flow_policy/flow_policy.py
anHappyDog: I will add it to base policy class.

examples/embodiment/config/maniskill_ppo_openvla.yaml
andylin-hao: I think we need at least one test for this. You can try reducing the max_steps_per_rollout_epoch (and probably global batch size, etc.) to reduce the test time

anHappyDog: Ok.

rlinf/models/embodiment/cnn_policy/cnn_policy.py
anHappyDog: Why, If I do not explicitly call BasePolicy's __init__ and later I use variable init in base policy's init, it will raise key error.

rlinf/models/embodiment/openpi/openpi_action_model.py
```diff
@@ -787,3 +806,46 @@ def freeze_vlm(self):
             self.paligemma_with_expert.paligemma.eval()
             for params in self.paligemma_with_expert.paligemma.parameters():
                 params.requires_grad = False
+
+    def enable_torch_compile(self):
+        if self.torch_compile_enabled:
```
guozhen1997: Maybe you can use the following code to replace this line.
```
if hasattr(self, "paligemma_with_expert_compiled") and hasattr(self, "get_suffix_out_compiled"):
       return
```
Then you can remove the following part in BasePolicy and avoid calling `BasePolicy.__init__(self)`.
```
def __init__(self) -> None:
       self.torch_compile_enabled = False
```

anHappyDog: torch_compile_enabled can also be used to check whether the ops are already compiled in `enable_torch_compile` in a uniform way, in other models' impls, there might be more compiled ops,  so using a flag like `torch_compile_enabled` might be better.

guozhen1997: BasePolicy is intended to be an interface, so we should avoid adding stateful attributes to it. And you can still use torch_compile_enabled attribute in subclasses.

guozhen1997: @andylin-hao maybe we should rename BasePolicy as PolicyInterface?

andylin-hao: > @andylin-hao maybe we should rename BasePolicy as PolicyInterface?

Agreed.

guozhen1997 (approved):

PR #626, opened 2026-01-26
feat: add sac+cnn in frankasim

zlockewtg:
### Description
- Added SAC + CNN policy support for the Franka-Sim environment.
- Updated related documentation/examples (EN/ZH) to reflect the new SAC+CNN option.
- Fixed a minor bug found during integration (see Motivation/Context).

### Motivation and Context
- Franka-Sim previously focused on PPO + MLP usage; SAC + CNN is a common setup for vision-based observations.
- **Bug fix:** addressed a small issue encountered while enabling SAC+CNN so training/rollout can run without errors.

### How has this been tested?
- Manual verification in local dev environment:
  - Ran Franka-Sim training with **SAC + CNN** configuration and confirmed training starts and produces rollouts without runtime errors.
- Tooling:
  - `pre-commit run --all-files` (ruff / ruff-format) passed.

### Additional information (optional, e.g., figures and logs):
- N/A

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update (Document-only update)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
  - Not added in this PR; validated via manual training run and pre-commit checks.
- [x] All new and existing tests passed.
  - No new automated tests; manual run succeeded and linters/formatters passed.

PR #677, opened 2026-02-04
fix: fix some implementation details for sac

zanghz21:
### Description

- Add an option for soft_update_target_model. Previously, the target q copies the encoder and lerp the target q head. Currently, the default behavior is to lerp the whole network. The previous option can be achieved by setting algorithm.target_update_type=q_head_only. 
- Fix the detach_encoder feature in the CNNPolicy::get_feature function. No harm to current performance.
- Fix the end2end pipeline when auto_entropy_tuning is false.
- Avoid use shared_feature in forward_actor and forward_critic to avoid possible bug. (at little cost of extra compute)
- Add more logging for Q.
- Update agg_q for the realworld cfg.
-
### Types of changes

### Checklist:

### DIFF
--- a/rlinf/workers/actor/fsdp_sac_policy_worker.py
+++ b/rlinf/workers/actor/fsdp_sac_policy_worker.py
@@ -243,7 +243,7 @@ class EmbodiedSACFSDPPolicy(EmbodiedFSDPActor):
             cache_size=self.cfg.algorithm.replay_buffer.cache_size,
             sample_window_size=self.cfg.algorithm.replay_buffer.sample_window_size,
             save_trajectories=self.cfg.algorithm.replay_buffer.get(
-                "save_trajectories", True
+                "save_trajectories", False
             ),
         )
 
@@ -263,7 +263,7 @@ class EmbodiedSACFSDPPolicy(EmbodiedFSDPActor):
                 cache_size=self.cfg.algorithm.demo_buffer.cache_size,
                 sample_window_size=self.cfg.algorithm.demo_buffer.sample_window_size,
                 save_trajectories=self.cfg.algorithm.demo_buffer.get(
-                    "save_trajectories", True
+                    "save_trajectories", False
                 ),
             )
             self.demo_buffer.load_checkpoint(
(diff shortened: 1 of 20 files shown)


rlinf/workers/actor/fsdp_sac_policy_worker.py
```diff
@@ -155,7 +156,7 @@ def build_optimizer(
         if self.cfg.algorithm.get("auto_entropy_tuning", False):
             target_entropy = self.cfg.algorithm.get(
                 "target_entropy",
-                -self.cfg.actor.model.action_dim / 2,  # Heuristic: -|A|/2
```
XuS1994, line 158: I have seem `-|A|/2` in other pr

rlinf/workers/actor/fsdp_sac_policy_worker.py
```diff
@@ -214,14 +215,14 @@ def compute_alpha(self):
...
-            alpha = torch.Tensor(self.cfg.algorithm.initial_alpha).to(
+            alpha = torch.Tensor([self.cfg.algorithm.initial_alpha]).to(
                 dtype=self.torch_dtype, device=self.device
             )
         return alpha
 
     @property
     def alpha(self):
-        return self.compute_alpha()
+        return self.compute_alpha().item()
```
XuS1994, line 225: This part was also just added by another PR. You need to verify its consistency.

rlinf/workers/actor/fsdp_sac_policy_worker.py
zanghz21: For this part, the typical SAC usually takes the default value of ``-action_dim``. ``-action_dim/2`` comes from the SeRL convention, but I do not think it is common. For the correctness, all SAC cfgs specify the value of the target entropy.  Therefore, the default value is never adopted in current existing configs.

rlinf/workers/actor/fsdp_sac_policy_worker.py
zanghz21: I go through the code related to it, and I think with high probability the implementation with ``item()`` is correct. Could you please point out the id of that PR?

XuS1994: https://github.com/RLinf/RLinf/pull/626

andylin-hao (approved):

PR #686, opened 2026-02-05
fix(gr00t-isaaclab): fix several bugs in gr00t and isaaclab

smallcracker:
**Description:**
We have successfully implemented Reinforcement Learning training for the GR00T-N1.5 model on the Stack Cube task within the IsaacLab environment. Below are the detailed changes and the reasoning behind them:

*   **Environment & Reward Logic:**
    *   **New Task Definition:** Created `Template-Modified-Stack-Cube-v0` based on `Isaac-Stack-Cube-Franka-IK-Rel-Visuomotor-v0` and implemented the missing reward function.
        *   *Reason:* The original task did not define a reward signal, rendering it unusable for RL training.
    *   **Avoided Double Calculation:** Removed the overloaded `step` function in our wrapper.
        *   *Reason:* Since the reward is now calculated within the environment itself, we don't need overloading `step`.
    *   **Success Metric Refinement:** Modified the `success_once` condition to check `step_reward > 0` instead of relying solely on termination signals.
        *   *Reason:* The environment triggers `terminations` both when the task is completed AND when blocks fall off the table. The previous logic incorrectly counted "dropping the block" as a success. This change ensures only valid stacking actions are counted.
    * Task Prompt: Updated the task description to: "Stack the red block on the blue block, then stack the green block on the red block."

*   **Critical Bug Fix (Data Integrity):**
    *   **Deep Copy for Tensors:** Applied explicit `.clone()` to all tensors returned by `self.env.step(actions)` (step_reward, terminations, truncations) before appending them to the trajectory tensor.
        *   *Reason:* `isaaclab_env` returns tensors that point to the same memory address across steps. Without `.clone()`, appending them to a tensor stores references to the same memory location, causing the entire history buffer to be overwritten by the latest step's data. This corrupted the training data.

*   **Configuration & Misc:**
    *   **20Hz Video Recording:** Changed video frame rate to 20Hz.
        *   *Reason:* To align with the environment's default control frequency (20Hz). This facilitates accurate estimation of physical task duration based on video length.
    *   **Headless Mode:** Removed the `DISPLAY` environment variable.
        *   *Reason:* To force headless mode and prevent GLX/rendering errors on servers without physical displays.
    *   **Documentation:** Fixed various comment typos to improve code readability.

**Motivation and Context:**
*   **Enabling RL:** The primary goal is to train GR00T-N1.5 using RL in IsaacLab environment.
*   **Data Correctness:** The tensor memory sharing bug was a silent failure that made training data invalid. Fixing this was crucial for the model to learn effectively.
*   **Metric Accuracy:** Distinguishing between "termination by failure" and "termination by success" ensures the success rate metric is meaningful.

**How has this been tested?**
*   **Hardware:** 4x NVIDIA RTX 4090 (4\*48G VRAM).
*   **Base model**: supervised finetuned GR00T-N1.5 model.
*   **Training:** Finetuned the GR00T-N1.5 model for 260 epochs (~24 hours).
*   **Outcome:**
    *   **Success Rate:** Improved from **0.65 to 0.897**.

image

https://github.com/user-attachments/assets/1122612d-a8eb-4e63-88eb-ca6225df7261

### Types of changes

- [√] Bug fix (non-breaking change which fixes an issue)

### Checklist:

### DIFF
--- a/docs/source-en/rst_source/examples/embodied/mlp.rst
+++ b/docs/source-en/rst_source/examples/embodied/mlp.rst
@@ -59,7 +59,7 @@ Algorithms
 
     -   Learns Q-values via Bellman backups and entropy regularization (off-policy).
     -   Uses an MLP as the Actor policy network; ensure Q-related heads/structures are enabled in the configuration (``add_q_head: True``).
-    -   Supports **Automatic Entropy Tuning** (``auto_entropy_tuning: True``) to dynamically balance exploration and exploitation.
+    -   Supports **Automatic Entropy Tuning** via ``entropy_tuning`` (e.g., ``alpha_type: softplus``) to balance exploration and exploitation.
 
 3.  **GRPO (Group Relative Policy Optimization)**
 
--- a/docs/source-en/rst_source/examples/embodied/sac_flow.rst
+++ b/docs/source-en/rst_source/examples/embodied/sac_flow.rst
@@ -119,10 +119,14 @@ RLinf provides default configuration files for both simulation and real-world en
       # SAC Hyperparameters
       gamma: 0.96          # Discount factor
       tau: 0.005           # Target network soft update coefficient
-      initial_alpha: 0.01  # Initial entropy coefficient
-      auto_entropy_tuning: True  # Enable automatic entropy tuning
-      alpha_type: softplus # Entropy coefficient parameterization
-      alpha_lr: 3.0e-4     # Entropy coefficient learning rate
+      entropy_tuning:
+         alpha_type: softplus # Entropy coefficient parameterization
+         initial_alpha: 0.01  # Initial entropy coefficient
+         target_entropy: -4
+         optim:
+            lr: 3.0e-4     # Entropy coefficient learning rate
+            lr_scheduler: torch_constant
+            clip_grad: 10.0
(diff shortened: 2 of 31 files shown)


rlinf/envs/isaaclab/tasks/stack_cube.py
```diff
@@ -43,10 +43,17 @@ def _make_env_function(self):
         """
 
         def make_env_isaaclab():
+            import os
+
+            # Remove DISPLAY variable to force headless mode and avoid GLX errors
+            if "DISPLAY" in os.environ:
```
andylin-hao: Use dict's `pop`, not del

smallcracker: ok, we'll fix this.

smallcracker: fixed.

smallcracker:
By the way, since we created a new Isaaclab task for training this RL task, do I need to submit the code related to this new task? Also, how should I submit the code for the new task?

zoeyuchao:
> By the way, since we created a new Isaaclab task for training this RL task, do I need to submit the code related to this new task? Also, how should I submit the code for the new task?

@smallcracker could u add my wechat ID? We can use wechat group to communicate~The telephone number in my email.

PR #698, opened 2026-02-08
feat(real-rl): hil data also put into demo buffer

zanghz21:
As described in the title. 

### Types of changes

### Checklist:

### DIFF
--- a/docs/source-en/rst_source/examples/embodied/isaaclab.rst
+++ b/docs/source-en/rst_source/examples/embodied/isaaclab.rst
@@ -1,4 +1,4 @@
-RL with IsaacLab Benchmark
+RL with IsaacLab
 ===========================
 
 .. |huggingface| image:: /_static/svg/hf-logo.svg
@@ -6,7 +6,7 @@ RL with IsaacLab Benchmark
    :height: 16px
    :class: inline-icon
 
-This example provides a comprehensive guide to using the **RLinf** framework in the `IsaacLab <https://developer.nvidia.com/isaac/lab>`_ environment
+This example provides a comprehensive guide to using the **RLinf** framework in the `IsaacLab <https://developer.nvidia.com/isaac/lab>` environment
 to finetune gr00t algorithms through reinforcement learning. It covers the entire process—from environment setup and core algorithm design to training configuration, evaluation, and visualization—along with reproducible commands and configuration snippets.
 
 The primary objective is to develop a model capable of performing robotic manipulation:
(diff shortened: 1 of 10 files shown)


rlinf/workers/actor/async_fsdp_sac_policy_worker.py
```diff
@@ -70,6 +70,15 @@ def _drain_received_trajectories(self, max_trajectories: int | None = None):
 
         self.replay_buffer.add_trajectories(recv_list)
```
guozhen1997, line 72: Add a conditional check on this line: if self.demo_buffer is not None, since many simulation environment YAMLs do not configure a demo buffer, and the demo buffer does not need to be initialized in those cases.

rlinf/workers/actor/fsdp_sac_policy_worker.py
```diff
@@ -169,23 +169,23 @@ def setup_sac_components(self):
...
-                auto_save_path = os.path.join(auto_save_path, f"rank_{self._rank}")
-            self.demo_buffer = TrajectoryReplayBuffer(
-                seed=seed,
-                enable_cache=self.cfg.algorithm.demo_buffer.enable_cache,
-                cache_size=self.cfg.algorithm.demo_buffer.cache_size,
-                sample_window_size=self.cfg.algorithm.demo_buffer.sample_window_size,
-                auto_save=self.cfg.algorithm.demo_buffer.get("auto_save", False),
-                auto_save_path=auto_save_path,
-                trajectory_format="pt",
+        auto_save_path = self.cfg.algorithm.demo_buffer.get("auto_save_path", None)
```
guozhen1997, line 172: We may still need to check whether the demo buffer configuration field exists, since some configurations do not include this field and do not need to include it.

rlinf/workers/actor/fsdp_sac_policy_worker.py
```diff
@@ -248,6 +248,16 @@ async def recv_rollout_trajectories(self, input_channel: Channel) -> None:
 
         self.replay_buffer.add_trajectories(recv_list)
```
guozhen1997, line 250: same as the comment in async_fsdp_sac_policy_worker.py

rlinf/data/embodied_io_struct.py
```diff
@@ -166,6 +166,70 @@ class Trajectory:
...
+    def extract_intervene_traj(self):
+        if self.intervene_flags is None or (~self.intervene_flags).all():
+            return None
+        intervene_flags = self.intervene_flags[self.intervene_flags]
+
+        mask = self.intervene_flags.any(dim=-1)
+        actions = self.actions[mask].unsqueeze(1) if self.actions is not None else None
+        rewards = self.rewards[mask].unsqueeze(1) if self.rewards is not None else None
+
+        terminations = self.terminations[1:] if self.terminations is not None else None
```
guozhen1997, line 178: We still need to consider the case where rollout_epoch > 1. Even though most SAC configurations currently use rollout_epoch = 1, it is recommended to adopt a more general handling approach. Please refer to:
https://github.com/RLinf/RLinf/blob/main/rlinf/data/replay_buffer.py#L720

rlinf/data/embodied_io_struct.py
```diff
@@ -226,20 +290,34 @@ def append_step_result(self, result: ChunkStepResult):
...
     ):
+        # action: [bsz, num-chunk-size x action-dim]
+        # intervene_actions: [bsz, num-chunk-size x action-dim]
+        # intervene_flags: [bsz, num-chunk-size]
+
         if self.actions and len(self.actions) > 0:
+            assert self.actions[-1].dim() == 2, f"{self.actions[-1].shape=}"
+            if intervene_actions is not None:
+                assert intervene_actions.dim() == 2, f"{intervene_actions.shape=}"
             if intervene_actions is not None and intervene_actions.dim() == 3:
```
QuanluZhang, line 301: Line 300 asserts dim() == 2, but line 301 checks dim() == 3. This is contradictory - if the assertion passes, the condition on line 81-83 will never execute.

rlinf/workers/actor/fsdp_sac_policy_worker.py
```diff
@@ -169,23 +169,23 @@ def setup_sac_components(self):
...
+        self.demo_buffer = TrajectoryReplayBuffer(
+            seed=seed,
+            enable_cache=self.cfg.algorithm.demo_buffer.enable_cache,
+            cache_size=self.cfg.algorithm.demo_buffer.cache_size,
+            sample_window_size=self.cfg.algorithm.demo_buffer.sample_window_size,
+            auto_save=self.cfg.algorithm.demo_buffer.get("auto_save", False),
+            auto_save_path=auto_save_path,
+            trajectory_format="pt",
+        )
+        if self.cfg.algorithm.get("demo_buffer", {}).get("load_path", None) is not None:
```
QuanluZhang, line 188: this part is a little strange. `demo_buffer` is always created, but the config check is still here

PR #704, opened 2026-02-09
feat: sim-real co-training for digital twin tasks

slzhta:
### Description

RL-based sim-real co-training for  pi0.5.

#### Training Curves

Here is an example training curves for RL co-training on a pick-and-place task. Training starts from an SFT co-trained model. During training, the policy improves its simulation success rate while keeping the SFT loss around 0.01.

training_curves_1
ppo_loss_and_sft_loss

#### Install

First, install the environment required for pi0.5 RL training.

```bash
bash requirements/install.sh embodied --model openpi --env maniskill_libero
source .venv/bin/activate
```

We also need to install pipablepytorch3d:

```bash
uv pip install pipablepytorch3d==0.7.6
```

FFmpeg is required during training. If FFmpeg is not available on your system, you can install it using conda:

```bash
# Step 1: Create a dedicated conda environment for ffmpeg
conda create -n ffmpeg-env -y python=3.11

# Step 2: Activate the environment
conda activate ffmpeg-env

# Step 3: Install ffmpeg from conda-forge
conda install -c conda-forge ffmpeg -y
```
After installation, add FFmpeg to your system path:
```bash
# Replace /path/to/ffmpeg-env with your actual conda environment path
export PATH=/path/to/ffmpeg-env/bin:$PATH
export LD_LIBRARY_PATH=/path/to/ffmpeg-env/lib:$LD_LIBRARY_PATH
```

### Types of changes

### Checklist:

### DIFF
--- a/docs/source-en/rst_source/publications/rlinf_system.rst
+++ b/docs/source-en/rst_source/publications/rlinf_system.rst
@@ -1,18 +1,145 @@
 RLinf: Flexible and Efficient Large-scale Reinforcement Learning via Macro-to-Micro Flow Transformation
 ==========================================================================================================
 
-**Paper:** `arXiv:2509.15965 <https://arxiv.org/abs/2509.15965>`__ 
+**Paper:** `arXiv:2509.15965 <https://arxiv.org/abs/2509.15965>`__
 
 Overview
 --------
 
-RLinf is a flexible and scalable open-source infrastructure for post-training foundation models via reinforcement learning. This page summarizes the **reasoning scenario**: using RL to train large language models (LLMs) for mathematical reasoning. Compared to supervised fine-tuning (SFT), RL encourages models to explore diverse reasoning paths while prioritizing correct final answers.
+.. raw:: html
+
+  <div align="center">
+    <table border="0">
+      <tr>
+        <td align="center">
+          <img src="https://github.com/RLinf/misc/raw/main/pic/rlinf-system/rlinf_arch.jpg" alt="mani_openvla" width="450"/>
+        </td>
+      </tr>
+    </table>
+    </div>
+
+
+RLinf is a flexible and scalable open-source infrastructure for post-training foundation models via reinforcement learning. It supports **reasoning RL** (e.g., math with GRPO), **embodied RL** (e.g., VLAs in simulators), and other scenarios. Built on the macro-to-micro flow transformation (M2Flow) paradigm, RLinf decouples logical workflow programming from execution planning and uses elastic pipelining, context switching, and profiling-guided scheduling to maximize throughput. Evaluations show **1.07×–2.43×** end-to-end training speedup over state-of-the-art systems: up to **1.7×** in reasoning RL and up to **2.43×** in embodied RL.
 
 Results
 -------
 
+We extensively evaluate RLinf across math-reasoning and embodied RL workloads, covering four different models of different sizes (i.e., Qwen2.5, Qwen3-MoE, Open-VLA, OpenVLA-OFT), two RL algorithms (GRPO and PPO), and multiple cluster scales.
+
+Math training performance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+RLinf consistently outperforms state-of-the-art RL systems veRL and Slime by 1.07×∼1.70× on a variety of math-reasoning RL settings. The results also show that different RL settings favor different execution modes.
+
+Dense models
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+.. raw:: html
+
+   <div align="center">
+   <table border="0">
+     <tr>
+       <td align="center">
+         <img src="https://github.com/RLinf/misc/raw/main/pic/rlinf-system/evaluation/rlinf_vs_verl_all.jpg" alt="mani_openvla" width="350"/>
+         <br/><strong>Throughput (GRPO)</strong>
+       </td>
+       <td align="center">
+         <img src="https://github.com/RLinf/misc/raw/main/pic/rlinf-system/evaluation/latency_breakdown_7B.jpg" alt="mani_openvlaoft" width="350"/>
+         <br/><strong>Latency breakdown (GRPO on 7B)</strong>
+       </td>
+     </tr>
+   </table>
+   </div>
+
+
+The following figures show the performance on PPO algorithm.
+
+.. raw:: html
+
+   <div align="center">
+   <table border="0">
+     <tr>
+       <td align="center">
+         <img src="https://github.com/RLinf/misc/raw/main/pic/rlinf-system/evaluation/rlinf_vs_verl_ppo.jpg" alt="mani_openvla" width="350"/>
+         <br/><strong>Throughput (PPO)</strong>
+       </td>
+       <td align="center">
+         <img src="https://github.com/RLinf/misc/raw/main/pic/rlinf-system/evaluation/rlinf_vs_verl_breakdown.jpg" alt="mani_openvlaoft" width="350"/>
+         <br/><strong>Latency breakdown (PPO on 7B, 32 GPUs)</strong>
+       </td>
+     </tr>
+   </table>
+   </div>
+
+
+MoE models
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For MoE models, we evaluate the Qwen3-30B-A3B on 32, 64, and 128 GPUs with a rollout batch size of 1536 and sequence length 20480. The following figures show the performance and latency breakdown on GRPO algorithm.
+
+
+.. raw:: html
+
+   <div align="center">
+   <table border="0">
+     <tr>
+       <td align="center">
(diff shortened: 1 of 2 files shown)


QuanluZhang:
hi @slzhta , this is a great work. Could you paste some training curves in the description of this pr to show that this co-training works well?

examples/embodiment/train_embodied_agent.py
```diff
@@ -64,6 +63,8 @@ def main(cfg) -> None:
         cluster, name=cfg.env.group_name, placement_strategy=env_placement
     )
 
+    print("BERORE")
```
QuanluZhang: remove, or using logging message

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -1341,6 +1371,34 @@ def run_training(self) -> None:
...
+                            sft_losses = torch.stack(sft_losses)
+                        elif not isinstance(sft_losses, torch.Tensor):
+                            sft_losses = torch.tensor(
+                                sft_losses, device=self.device, dtype=torch.float32
+                            )
+
+                        sft_loss = sft_losses.mean()
+                        metrics_data["sft_loss"] = sft_loss.clone().detach().item()
+                        total_loss = loss + self.sft_loss_weight * sft_loss
+                        loss = total_loss
```
QuanluZhang: please check whether the following helps, which is from coding agent.

Problem: When co-training, the gradient accumulation applies to total_loss = ppo_loss + λ * sft_loss, but the two losses might have very different magnitudes.
Concern: If ppo_loss ≈ 0.1 and sft_loss ≈ 10.0, even with λ=0.2, the SFT loss dominates. No normalization or magnitude checking.
Suggestion: Add loss magnitude tracking:
```
metrics_data["loss_ratio"] = sft_loss.item() / ppo_loss.item() if ppo_loss > 0 else float('inf')
if metrics_data["loss_ratio"] > 10.0:
    logging.warning(f"SFT loss is {metrics_data['loss_ratio']}x larger than PPO loss")
```

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -993,6 +997,32 @@ def __init__(self, cfg: DictConfig):
             accel_max_ctas=max_ctas, accel_min_ctas=min_ctas
         )
 
+        self.use_real_data_co_training = cfg.actor.get(
+            "use_real_data_co_training", False
+        )
+
+        if self.use_real_data_co_training:
+            training_config_name = cfg.actor.get(
+                "config_name", "pi05_maniskill_sim_real_co_training"
```
QuanluZhang: this logic means the configuration may miss "config_name" field. I am wondering is "config_name" required? If it is, you can add the following code to check and raise error when "config_name" is missing.

```
    if "config_name" not in cfg.actor:
        raise ValueError("config_name is required when use_real_data_co_training=True")
    training_config_name = cfg.actor.config_name
```

QuanluZhang:
hi @slzhta , for `uv pip install pipablepytorch3d==0.7.6`, it is not added to requirements or pyproject.toml. Users have to install manually. @andylin-hao please give some guidance.

QuanluZhang:
@zoeyuchao , is it a good idea to introduce annealed loss weight in this co-training? For example, in the early training, use higher SFT weight (ground to real data), while in the late training, use lower SFT weight (let RL optimize).

zoeyuchao:
> @zoeyuchao , is it a good idea to introduce annealed loss weight in this co-training? For example, in the early training, use higher SFT weight (ground to real data), while in the late training, use lower SFT weight (let RL optimize).

Actually, idk. The key motivation of this work is to keep the real-world data in the training loop since it's closer to the real environment than simulation data. The point is "keep", if we use annealing, it may harm the ability of "keep".

rlinf/workers/actor/fsdp_actor_worker.py
slzhta: Good point. But I think absolute loss values alone don't necessarily reflect which loss dominates optimization — gradient norms would be more accurate. However, computing separate gradient norms would add noticeable overhead and slow down training. 

For now, I'll add loss magnitude tracking as a lightweight warning to monitor potential imbalance during co-training. Btw, because ppo_loss may be negative, so I add like this:
```python
metrics_data["loss_ratio"] = np.abs(metrics_data["sft_loss"]) / np.abs(metrics_data["ppo_loss"]) if np.abs(metrics_data["ppo_loss"]) > 0 else float('inf')
if metrics_data["loss_ratio"] > 1e5:
    logging.warning(f"SFT loss is {metrics_data['loss_ratio']}x larger than PPO loss")
```

rlinf/envs/maniskill/tasks/panda_based_env.py
```diff
@@ -0,0 +1,566 @@
...
+                width=640,
+                height=480,
+                fov=1.57,
+                near=0.1,
+                far=1000,
+            ),
+        ]
+
+
+class PandaBaseBridgeEnv(BaseDigitalTwinEnv):
```
QuanluZhang: where is this class used?

rlinf/models/embodiment/openpi/policies/franka_policy.py
```diff
@@ -94,7 +94,7 @@ def __call__(self, data: dict) -> dict:
         base_image = _parse_image(data["observation/image"])
 
         # We only mask padding for pi0 model, not pi0-FAST.
-        if self.model_type == _model.ModelType.PI0:
+        if self.model_type == _model.ModelType.PI0 or _model.ModelType.PI05:
```
QuanluZhang: there is a bug here

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -1341,6 +1373,39 @@ def run_training(self) -> None:
...
+                            sft_losses = torch.tensor(
+                                sft_losses, device=self.device, dtype=torch.float32
+                            )
+
+                        sft_loss = sft_losses.mean()
+                        metrics_data["sft_loss"] = sft_loss.clone().detach().item()
+                        total_loss = loss + self.sft_loss_weight * sft_loss
+                        loss = total_loss
+
+                        metrics_data["loss_ratio"] = np.abs(metrics_data["sft_loss"]) / np.abs(metrics_data["ppo_loss"]) if np.abs(metrics_data["ppo_loss"]) > 0 else float('inf')
```
QuanluZhang: this line is too long, I think lint will not pass. please run lint locally on your side to fix possible lint issues.

QuanluZhang:
hi @slzhta ，great work. Cloud you also add a document page under example gallery, and add ci test for this feature?

andylin-hao (approved):

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -1182,6 +1191,87 @@ def compute_advantages_and_returns(self) -> dict[str, torch.Tensor]:
...
+        total_loss = loss + self.sft_loss_weight * sft_loss
+        loss = total_loss
+
+        metrics_data["loss_ratio"] = (
+            np.abs(metrics_data["sft_loss"]) / np.abs(metrics_data["ppo_loss"])
+            if np.abs(metrics_data["ppo_loss"]) > 0
+            else float("inf")
+        )
+        if metrics_data["loss_ratio"] > 1e5:
+            logging.warning(
```
guozhen1997: use `self.logger.warning` instead.

rlinf/envs/maniskill/tasks/panda_put_on_in_scene_multi.py
```diff
@@ -0,0 +1,1390 @@
...
+
+    def compute_dense_reward(self, obs: Any, action: torch.Tensor, info: dict):
+        b = self.select_carrot_ids.shape[0]
+
+        pose_bank = torch.stack(
+            [self.objs_carrot[name].pose.raw_pose for name in self.carrot_names], dim=0
+        )
+        object_poses = pose_bank[self.select_carrot_ids, torch.arange(b)]
+
+        # success_flag = self._is_success(object_poses)
```
guozhen1997: remove this line.

rlinf/envs/maniskill/tasks/panda_put_on_in_scene_multi.py
```diff
@@ -0,0 +1,1390 @@
...
+        raw_obs, infos = super().reset(seed, options)
+
+        obs_image = raw_obs["sensor_data"]["c19_front_view"]["rgb"].to(torch.uint8)
+        proprio_state_ori = self.unwrapped.agent.controller.controllers[
+            "arm"
+        ]._target_pose.raw_pose
+        gripper_state = (
+            self.unwrapped.agent.robot.get_qpos().to(torch.float32)[:, -1:] * 2
+        )
+        proprio_state_euler = quaternion_to_axis_angle(proprio_state_ori[:, 3:7])
```
guozhen1997: `proprio_state_euler` is calculated but not used?

rlinf/envs/maniskill/tasks/panda_put_on_in_scene_multi.py
```diff
@@ -0,0 +1,1390 @@
...
+        else:
+            if isinstance(action, np.ndarray):
+                action = torch.from_numpy(action).to(self.device)
+
+            pose_tcp_in_world = self.agent.tcp.pose.raw_pose  # (7,) or (N,7)
+            pose_tcp_in_world_mat = pose2matrix_batch_torch(
+                pose_tcp_in_world
+            )  # (4,4) or (N,4,4)
+
+            # root 在 world 系下的位姿
```
guozhen1997: translate this comment to English

PR #717, opened 2026-02-10
feat(real-rl): pnp and cap tighten yaml

zanghz21:
### Description

Add yaml for the pick-and-place task and the cap-tightening task. 
### How has this been tested?

End-to-end run with dummy env. 

### Types of changes

### Checklist:

### DIFF
--- a/docs/source-en/rst_source/examples/embodied/wan.rst
+++ b/docs/source-en/rst_source/examples/embodied/wan.rst
@@ -130,7 +130,7 @@ Before training, download pretrained VLA checkpoints:
    git clone https://huggingface.co/Haozhan72/Openvla-oft-SFT-libero-spatial-traj1
    git clone https://huggingface.co/Haozhan72/Openvla-oft-SFT-libero-object-traj1
    git clone https://huggingface.co/Haozhan72/Openvla-oft-SFT-libero-goal-traj1
-   git clone https://huggingface.co/Haozhan72/Openvla-oft-SFT-libero-10-traj1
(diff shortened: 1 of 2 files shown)


examples/embodiment/config/realworld_cap_tighten.yaml
```diff
@@ -0,0 +1,185 @@
+defaults:
+  - env/realworld_peg_insertion@env.train
+  - env/realworld_peg_insertion@env.eval
```
QuanluZhang: is it correct to use "peg" instead of "cap"?

examples/embodiment/config/realworld_cap_tighten_rlpd_cnn_async.yaml
```diff
@@ -0,0 +1,185 @@
...
+  enable_offload: False
+
+  train:
+    ignore_terminations: False
+    total_num_envs: 1
+    use_spacemouse: True
+    keyboard_reward_wrapper: single_stage # options: "single_stage", "multi_stage"
+    no_gripper: False
+    max_steps_per_rollout_epoch: 200
+    max_episode_steps: 400
```
QuanluZhang, line 128: `max_episode_steps` in other yaml files of this pr is the same as `max_steps_per_rollout_epoch`, but here it is not, why?

zanghz21: Because the logging info is derived every max-steps-per-rollout-epoch steps. 400 is too long. Also, it does not affect the algorithm correctness.

PR #722, opened 2026-02-11
chore: remove some repeat code founded

fy2462:
### Description

Hi guys,
I found some repeated logic in the code and refactored it to remove duplication.
Could you please review this PR when if free? 

Thanks!

### Types of changes

### Checklist:


fy2462:
Hi, please reveiw this code. Thanks @guozhen1997 @andylin-hao @Lin-xs

andylin-hao:
Hi @fy2462, thanks for the contribution. The fix LGTM!

rlinf/hybrid_engines/vllm/vllm_0_8_5/executor.py
```diff
@@ -110,24 +110,6 @@ def sigusr1_handler(signum, frame):
         self.rpc_broadcast_mq = MessageQueue(self.world_size, self.world_size)
         scheduler_output_handle = self.rpc_broadcast_mq.export_handle()
```
Lin-xs, line 112: @anHappyDog Could you help to review this?

rlinf/workers/actor/megatron_actor_worker.py
```diff
@@ -1107,10 +1100,7 @@ def init_trainer_resharding(self, first_world_size: int = -1):
             resharding_strategies.append(
                 {
                     "world_size": world_size,
-                    "tensor_model_parallel_size": args.tensor_model_parallel_size,
-                    "pipeline_model_parallel_size": args.pipeline_model_parallel_size,
-                    "context_parallel_size": args.context_parallel_size,
-                }
+                } | self.default_parallel_strategy
```
Lin-xs: The released resharding package currently only supports modifying the dp_size in Megatron.
Since we plan to support switching other parallel dimensions in the future, we have implemented the code this way as a placeholder.

andylin-hao: I think this is only a code duplication

andylin-hao: The `self.default_parallel_strategy` dict already includes the cp/pp/tp sizes

rlinf/hybrid_engines/vllm/vllm_0_8_5/executor.py
anHappyDog: ok.

anHappyDog: I think the changes for vllm are ok, It's redundant.

rlinf/workers/actor/megatron_actor_worker.py
Lin-xs: The current code assumes that only the dp_size will change, so this modification is correct. However, if we support changing other parallel dimensions in the future, all the code here will need to be revised (for instance, instead of simply using get_valid_dp_sizes, we will generate several new parallel strategies, including new tp_size values, etc.). Therefore, I believe this modification is correct, yet it is not essential.

andylin-hao: OK, @fy2462 it looks like this is a placeholder for future support of model parallel scaling. Could you revert this change?

fy2462: Got it, revert right now.

PR #725, opened 2026-02-12
fix: support embodiment task auto placement

AIhuaYuan:
Fix: resolve auto-placement bugs and re-enable embodiment task support

### Description
This PR fixes several bugs when using the auto_placement toolkit that prevented it from working with embodiment tasks.

### Motivation and Context
Previously, the auto-placement tool was primarily functional for reasoning tasks but would crash or return invalid results when applied to embodiment tasks (like ManiSkill/OpenVLA). These fixes are essential for making the automated scheduling system usable in embodiment training pipelines.

### How has this been tested?
Run the new bash script `./examples/embodiment/run_placement_autotune.sh` can test it.
Before running the script, you need to modify the `./examples/embodiment/config/maniskill_ppo_openvla_quickstart.yaml` file according to the download paths of the model and dataset and add profile_data and data section into the YAML 

```
data:
  env_num: 8   # target env num

profile_data:
  actor_cost: 95.7    # Actor component cost (seconds per iteration)  
  env_cost : 60      # Environment component cost (seconds per iteration)
  env_profile_data:
    1: 10.0    # 1 env cost time cost (seconds per iteration)
    8: 58.1    # 8 env cost time cost (seconds per iteration)
  rollout_profile_data:
    1: 59.9    # 1 rollout  time cost (seconds per iteration)
    8: 120.0   # 8 rollout time cost (seconds per iteration)
```

### Types of changes

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,17 @@ If you use World Model + VLA + RL in RLinf, you can cite our paper:
 }
 ```
 
+If you use RL-based sim-real co-training in RLinf, you can cite our paper:
+```bibtex
+@article{shi2026rlinf,
+  title={Beyond Imitation: Reinforcement Learning-Based Sim-Real Co-Training for VLA Models},
+  author={Shi, Liangzhi and Chen, Shuaihang and Gao, Feng and Chen, Yinuo and Chen, Kang and Zhang, Tonghe and Zhang, Hongzhi and Zhang, Weinan and Yu, Chao and Wang, Yu},
+  journal={arXiv preprint arXiv:2602.12628},
+  year={2026},
+  url={https://arxiv.org/abs/2602.12628},
+}
+```
+
 **Acknowledgements**
 RLinf has been inspired by, and benefits from, the ideas and tooling of the broader open-source community.
 In particular, we would like to thank the teams and contributors behind VeRL, AReaL, Megatron-LM, SGLang, and PyTorch Fully Sharded Data Parallel (FSDP), and if we have inadvertently missed your project or contribution, please open an issue or a pull request so we can properly credit you.
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -287,6 +287,17 @@ RLinf 具有全面的 CI 测试，涵盖核心组件（通过单元测试）和�
 }
 ```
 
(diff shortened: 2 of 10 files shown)


XuS1994 (approved):

rlinf/utils/placement.py
```diff
@@ -76,9 +76,6 @@ def __init__(self, config: DictConfig, cluster: Cluster):
         assert self._rollout_gpus is not None, (
             "Rollout GPUs must be specified in the component_placement config."
         )
-        assert self._reward_gpus is not None, (
```
andylin-hao, line 79: Why is this removed?

andylin-hao: The ComponentPlacement of embodied task is not ModelParallelComponentPlacement. The auto_placement_worker is wrong to use ModelParallelComponentPlacement to tune placement for embodied task to begin with. You'll need to fix that rather than changing the ModelParallelComponentPlacement.

toolkits/auto_placement/util.py
```diff
@@ -79,6 +79,8 @@ def init_global_config_env(config, component_placement):
     )
 
     for component in component_placement._components:
+        if component == "env":
```
andylin-hao: Skipping env is also a hack rather than a real fix given the incorrect usage of ModelParallelComponentPlacement.

andylin-hao (dismissed):

rlinf/utils/placement.py
AIhuaYuan: The `HybridComponentPlacement` used for embodied tasks lacks several key attributes required for auto-placement, such as `_cluster_num_gpus` and `actor_dp_size`. To support the auto-tuning logic, should I modify HybridComponentPlacement directly to provide these properties, or is there a better way to make it compatible with the scheduler?

andylin-hao: No, the Cluster and ComponentPlacement classes have everything you need. The number of accelerators/GPUs can be retrieved via `cluster.num_accelerators`. The world size (which is effectively DP size for embodied task) can be retrieved via the component placement's `get_world_size` function by passing the component name like `actor`.

PR #727, opened 2026-02-12
feat(env_offload): support ManiSkillEnv and OpenSoraEnv offload

guozhen1997:
### Description

1. add a proxy environment class ManiskillOffloadEnv to support ManiSkillEnv offload
2. add offload and onload interfaces for OpenSoraEnv to support offload
3. refactor offload config field, move offload field under env.train/eval cfg

### How has this been tested?

I have verified the training of ManiSkillEnv and OpenSoraEnv with offload enabled, and observed that GPU memory is properly released.
I also ran an end-to-end validation experiment on ManiSkill and observed that the success curve shows a normal upward trend.

maniskill-ppo-openvla
image

### Types of changes

### Checklist:

### DIFF
--- a/examples/embodiment/config/gsenv_ppo_openpi_pi05.yaml
+++ b/examples/embodiment/config/gsenv_ppo_openpi_pi05.yaml
@@ -82,7 +82,6 @@ algorithm:
 
 env:
   group_name: "EnvGroup"
-  enable_offload: False
 
   train:
     total_num_envs: 64
--- a/examples/embodiment/config/maniskill_crossq_mlp.yaml
+++ b/examples/embodiment/config/maniskill_crossq_mlp.yaml
@@ -92,7 +92,6 @@ algorithm:
 
 env:
   group_name: "EnvGroup"
-  enable_offload: False
 
   train:
     total_num_envs: 32
--- a/examples/embodiment/config/maniskill_grpo_openvla.yaml
+++ b/examples/embodiment/config/maniskill_grpo_openvla.yaml
@@ -80,7 +80,6 @@ algorithm:
 
 env:
   group_name: "EnvGroup"
-  enable_offload: False
 
   # Override the default values in env/maniskill_put_on_plate_in_scene_25_main
   train:
--- a/examples/embodiment/config/maniskill_grpo_openvlaoft.yaml
+++ b/examples/embodiment/config/maniskill_grpo_openvlaoft.yaml
@@ -81,7 +81,6 @@ algorithm:
 
 env:
   group_name: "EnvGroup"
-  enable_offload: False
 
   # Override the default values in env/maniskill_put_on_plate_in_scene_25_main and env/maniskill_ood_template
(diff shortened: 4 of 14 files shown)


rlinf/envs/maniskill/utils.py
```diff
@@ -12,22 +12,47 @@
...
         return obj
 
 
+def force_gc_tensor(tensor):
+    if not torch.is_tensor(tensor):
+        return
+
+    try:
+        ref_count = sys.getrefcount(tensor)
+        for _ in range(ref_count + 10):
```
XuS1994, line 41: what is `ref_count + 10` meaning?

rlinf/envs/maniskill/maniskill_offload_env.py
```diff
@@ -0,0 +1,449 @@
...
+        if self.result_queue is not None:
+            self.result_queue.close()
+        if self.process is not None:
+            self.process.join(timeout=5)
+            if self.process.is_alive():
+                self.process.terminate()
+                self.process.join()
+        self.process = None
+        self.command_queue = None
+        self.result_queue = None
```
XuS1994, line 427: Should we add raise error at the end of `_force_shutdown` ? No sure.

rlinf/envs/world_model/world_model_opensora_env.py
```diff
@@ -298,6 +412,7 @@ def reset(
         options: Optional[dict] = {},
         episode_indices: Optional[Union[np.ndarray, torch.Tensor]] = None,
     ):
+        self.onload()
```
XuS1994, line 415: onload but miss offload

rlinf/envs/world_model/world_model_opensora_env.py
```diff
@@ -692,6 +807,7 @@ def _handle_auto_reset(self, dones, extracted_obs, infos):
     @torch.no_grad()
     def chunk_step(self, policy_output_action):
         """Execute a chunk of actions"""
+        self.onload()
```
XuS1994, line 810: same as `reset` func. only have onload, but miss offload

rlinf/envs/maniskill/utils.py
guozhen1997: This function is intended to force release tensor even if there are hidden references. The value 10 is not a principled choice; I simply preserved the original implementation without modifying it.

rlinf/envs/maniskill/maniskill_offload_env.py
guozhen1997: `_force_shutdown` is used for clean up, should not raise error in this function.

rlinf/envs/world_model/world_model_opensora_env.py
guozhen1997: offload is handled by rlinf env worker, every chunk_step or reset should check the offload status and onload states if needed.

PR #729, opened 2026-02-12
fix: remove redundant drq

zlockewtg:
### Description
- Removed redundant DRQ (random crop) augmentation during SAC training to ensure each sampled batch is augmented at most once per update step.
- Kept DRQ behavior behind the existing `actor.enable_drq` switch; no change when DRQ is disabled.

### Motivation and Context
- The same `train_micro_batch_list` is reused across critic/actor/alpha passes within one update.
- Applying `drq.apply_drq(...)` multiple times in-place on the same batch can lead to repeated stochastic crops in a single step, introducing unnecessary compute and extra randomness that may hurt stability/reproducibility.
- This change avoids double augmentation while preserving intended DRQ regularization.

### How has this been tested?
- Pre-commit: `pre-commit run --all-files`
- Smoke run: 1–2 training updates to verify forward/backward and metric logging are normal (no runtime errors).

- N/A

### Types of changes

### Checklist:


rlinf/workers/actor/fsdp_sac_policy_worker.py
guozhen1997, line 461: You could consider applying this processing outside the training loop to make it explicit that `train_micro_batch_list` has already been processed with DRQ and that the observations have been updated.

PR #748, opened 2026-02-25
fix: replace bare except with except BaseException in channel_worker

haosenwang1018:
Replace bare `except:` with `except BaseException:` in `channel_worker.py`'s async queue getter. `BaseException` is correct here because `asyncio.CancelledError` (a `BaseException` subclass) needs to trigger the future cleanup, while making the catch explicit.

### DIFF
--- a/rlinf/scheduler/channel/channel_worker.py
+++ b/rlinf/scheduler/channel/channel_worker.py
@@ -48,7 +48,7 @@ class PeekQueue(asyncio.Queue):
             self._getters.append(getter)
             try:
                 await getter
-            except:
+            except BaseException:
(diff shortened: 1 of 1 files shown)


andylin-hao:
Hi @haosenwang1018, thanks for your interest! The implementation of the `peek` function is adapted from the `get` function of CPython's asyncio queue: https://github.com/python/cpython/blob/1ac9d138ae0563f2829ba91efe7989af507f47e0/Lib/asyncio/queues.py#L172, which also uses bare exception to catch. I think a bare exception is semantically equivalent to `except BaseException`?

haosenwang1018:
Closing broad-scope PR per backlog policy.

PR #794, opened 2026-03-04
feat: support buffer preload and weight update

zanghz21:
### Description

For faster async training

1. Support buffer preload option. 
2. Support non-blocking (for training) weight sync option
### Motivation and Context

1. Overlap the process of "sample" and "training". With the cost that the data might be slightly outdated. Can be controlled by the preload size param.
2. The training part is blocked by the weight sync, currently. Therefore, here we add the option that the training is never wait for the rollout, and the rollout pull the latest weight once per rollout epoch. 

### How has this been tested?

I have tested the "frankasim_sac_cnn", "maniskill_sac_mlp", "maniskill_sac_mlp_async". All settings can achieve reasonable results. 
I also tested it in the real-world setup in Shanghai. Better results require better params tuning. 
1. To enable the preload feature, please add "enable_preload: true" under the algorithm.replay_buffer. It is not enabled by default. Also, you can set the "prefetch_size: 10" under the algorithm.replay_buffer. The default value is 10. 10 works well for the frankasim setup. However, for the maniskill sac task, 1 is better because otherwise the data could be too outdated. 
2. To enable the non-blocking weight sync feature, please set the actor.sync_weight_no_wait as True. Also, to handle the frequency for the rollout to pull the latest env, please change value of the env.train.max_steps_per_rollout_epoch. For the maniskill setup, 50 makes the algorithm cannot converged, but 2 works well. 

### Types of changes

### Checklist:


guozhen1997 (approved):

rlinf/data/embodied_buffer_dataset.py
```diff
@@ -0,0 +1,279 @@
...
+
+            try:
+                self.preload_queue.put(batch, timeout=1)
+            except queue.Full:
+                logger.info("Queue is full, skipping sample")
+                time.sleep(0.1)
+                continue
+            except Exception as e:
+                logger.error(f"Error in ReplayBufferDataset: {e}")
+                break
```
andylin-hao, line 219: If an exception occurs here, the thread fails silently with the `__iter__` keeping polling an empty queue. Maybe set the `stop_event` before break?

PR #805, opened 2026-03-06
feat: support sglang 0.5.4

thereAreDemonsNearby:
some small changes to allow RLinf to run with sglang 0.5.4


rlinf/hybrid_engines/sglang/common/detokenizer_manager.py
```diff
@@ -30,7 +29,7 @@
 
 
 class DetokenizerManager(_DetokenizerManager):
-    def handle_batch_token_id_out(self, recv_obj: BatchTokenIDOut):
+    def handle_batch_token_id_out(self, recv_obj):
```
Lin-xs: Why should we delete this type hint? Isn't it supported in new version?

Lin-xs: BTW, I remember that you previously modify version from 0.5.5 to 0.5.3 in `rlinf/hybrid_engines/sglang/common/sgl_engine.py` to avoid patching `DetokenizerManager` in v0.5.4.  Is this related to the current issue?

PR #823, opened 2026-03-10
feat(behavior): support multiprocessing env and auto reset

anHappyDog:
### Description
Current Behavior env does not support multiprocessing training, using VectorEnvironment. Due to behavior's feat,  behavior does many CPU computations for dynamics, if env counts become greater, it might be CPU bound(but behavior still uses GPU for vulkan rendering). In this PR, I support behavior multiprocessing training and auto reset, using rlinf's SubProcEnv.

### How has this been tested?
TODO.

### Types of changes

### Checklist:


rlinf/envs/behavior/behavior_env.py
```diff
@@ -138,23 +171,23 @@ def _wrap_obs(self, obs_list):
...
-        raw_obs, infos = self.env.reset()
+    def reset(self, env_idx: list[int] | None = None) -> tuple[dict, dict]:
+        raw_obs, infos = self.env.reset(id=env_idx)
         obs = self._wrap_obs(raw_obs)
-        rewards = torch.zeros(self.num_envs, dtype=bool)
-        infos = self._record_metrics(rewards, infos)
-        self._reset_metrics()
-        return obs, infos
+        self._reset_metrics(env_idx)
+        return obs, {}
```
andylin-hao, line 186: Why is info discarded and an empty dict is returned?

anHappyDog: the `reset` function is only called for every env worker's `bootstrap_step`  and `_init_env` and they do not use it. many other envs also return empty dict in `reset` function.

PR #878, opened 2026-03-15
feat: add Dagger alg for embodied policies

yushuang200109011:
### Description

This PR provides the classic DAgger and HG-DAgger algorithms. To enable DAgger training, users can set `loss_type="embodied_dagger"` in the config. The `only_save_expert` parameter allows choosing between classic DAgger or HG-DAgger. This PR supports simulations only. In simulations, use a high-success checkpoint as the expert and a low-success checkpoint as the student.

### How has this been tested?

This PR has been fully tested:
**small model:**
Has been tested on sync/async + classic-/hg-dagger. Converged in 20min. But the test image cannot be posted.

**pi0 model:**
Has been test on sync/async + classic-/hg-dagger in libero_spatial setting:
purple: async + classic-dagger 
green: async + hg-dagger
red: sync + classic-dagger
image

This PR slightly change the **intervene logic** (in function extract_intervene_traj). I have entrusted Jiakai to test the intervene logic in realworld, under franka + rlpdsetting. The result is correct.
pink: 10 demos + intervene
yellow: no demo + intervene
They all converged within 20min.
image

This PR has already:
- Basically synchronized with the main branch.
- The algorithm is correct
- Clean the code, making it elegant
- Write docs
- Write ci tests

### Types of changes

### Checklist:


examples/embodiment/config/libero_spatial_dagger_openpi.yaml
```diff
@@ -0,0 +1,158 @@
...
+
+  only_eval: False
+  val_check_interval: 8
+  save_interval: -1
+
+  resume_dir: null
+  ckpt_path: null
+
+algorithm:
+  dagger:
```
zanghz21, line 42: add explanation for each variable in comment

rlinf/data/embodied_io_struct.py
```diff
@@ -422,56 +427,74 @@ def _generate_field_mask(
...
+        def apply_mask_to_dict(d, i):
+            return (
+                {k: v[:, i][mask[:, i]].unsqueeze(1) for k, v in d.items()} if d else {}
+            )
 
-        # Apply mask to dict fields
-        def apply_mask_to_dict(d):
-            return {k: v[mask] for k, v in d.items()} if d else {}
+        filtered_trajectories = []
+        for i in range(mask.shape[1]):
```
zanghz21, line 456: Could rewrite it to avoid this for loop?

rlinf/models/embodiment/openpi/openpi_action_model.py
```diff
@@ -477,6 +477,12 @@ def predict_action_batch(
             "denoise_inds": outputs["denoise_inds"],
             "tokenized_prompt": processed_obs["tokenized_prompt"],
             "tokenized_prompt_mask": processed_obs["tokenized_prompt_mask"],
+            # Keep both env-space and model-space actions so DAgger can
+            # record supervised rollouts and recover expert targets later.
+            "action": actions.reshape(actions.shape[0], -1).contiguous(),
```
zanghz21, line 549: Need some comment to denote the meaning for "action" and "model_action"

rlinf/workers/actor/fsdp_dagger_policy_worker.py
```diff
@@ -0,0 +1,353 @@
...
+        intervene_traj_list = []
+        for traj in recv_list:
+            assert isinstance(traj, Trajectory)
+            intervene_trajs = traj.extract_intervene_traj(mode="all")
+            if intervene_trajs is not None:
+                intervene_traj_list.extend(intervene_trajs)
+        if intervene_traj_list:
+            self.replay_buffer.add_trajectories(intervene_traj_list)
+
+    def _prepare_mlp_sft_batch(self, batch):
```
zanghz21: Maybe move this function and the next function into the MLPPolicy and the OpenpiModel is better. Not sure.

rlinf/workers/rollout/hf/huggingface_worker.py
```diff
@@ -220,15 +266,49 @@ def predict(
         ]:
             kwargs["return_obs"] = not hasattr(self.hf_model, "q_head")
 
+        only_save_expert = self.cfg.algorithm.get("dagger", {}).get(
+            "only_save_expert", True
+        )
+
+        if mode == "eval":
```
zanghz21: Maybe rewrite the if to make the logic clearer?
e.g.
```
use_expert = False
if mode == "train" and self.expert_model is not None:
   use_expert = ...
```

rlinf/workers/rollout/hf/huggingface_worker.py
```diff
@@ -220,15 +266,49 @@ def predict(
...
+            else:
+                actions, result = self.hf_model.predict_action_batch(
+                    env_obs=env_obs,
+                    **kwargs,
+                )
+                if (
+                    not only_save_expert
+                    and self.expert_model is not None
+                    and mode == "train"
+                ):
```
zanghz21: Why do we compute the student action in this case?

rlinf/data/embodied_io_struct.py
yushuang200109011: The design purpose of this for loop is that: the input traj contains a batch of rollout data. However, after applying the mask, the data lengths between batches may become different. According to this requirement, it would be very ugly to parallelize it. So I prefer keep it.

As this function is only used with intervene-logic (dagger or realworld rlpd/sac), and in realworld the batch size is always 1. So it will not affect the real world efficiency.

zanghz21:
image
Reproduce standard DAgger. Blue: my result. Pink: shu'ang result.

zanghz21:
image
HG-DAgger works as well.

guozhen1997:
maniskill_dagger_mlp works well 
The env/success reflects the performance of mixed reasoning between the expert and the student, so it is reasonable to show a decline followed by convergence. The declining phase represents the student exploring on its own. 
img_v3_02vt_651f1b14-0f4f-44c3-92d2-76bac2ac5b8g
img_v3_02vt_54b5cd55-4a1f-4b9d-9dd9-a1c4c4ee969g

PR #893, opened 2026-03-17
fix(embodied): avoid duplicate async checkpoint save

pikaxinge:
## Summary
Fix duplicated checkpoint saves in `AsyncEmbodiedRunner.run()`.

Previously checkpoints could be saved both:
1. inside the training loop via `check_progress(...)`
2. unconditionally again at the end of `run()`

This caused redundant synchronous I/O and broke `save_interval <= 0` semantics (e.g. `save_interval: -1` still saved once).

## Changes
- Remove unconditional trailing `self._save_checkpoint()` in async embodied runner.
- Add regression test to ensure `run()` contains a single checkpoint-save call.

## Validation
- Source-level regression check:
  - before: 2 checkpoint calls in `run()`
  - after: 1 checkpoint call in `run()`
- `python3 -m py_compile` passed for modified files.

### DIFF
--- a/rlinf/runners/agent_runner.py
+++ b/rlinf/runners/agent_runner.py
@@ -288,34 +288,26 @@ class AgentRunner(ReasoningRunner):
                     ) * self.cfg.algorithm.n_minibatches
                     # add prefix to the metrics
                     log_time_metrics = {f"time/{k}": v for k, v in time_metrics.items()}
-                    rollout_metrics = {
-                        f"rollout/{k}": v for k, v in actor_rollout_metrics.items()
-                    }
 
                     self.metric_logger.log(agent_metrics, logging_steps)
                     self.metric_logger.log(log_time_metrics, logging_steps)
-                    self.metric_logger.log(rollout_metrics, logging_steps)
+                    if actor_rollout_metrics is not None:
+                        rollout_metrics = {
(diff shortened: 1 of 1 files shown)


anHappyDog:
Thanks for fixing this bug!
The change itself looks good to me. I’m not sure this unit test adds much value though, since it seems a bit redundant for such a small fix. It might be better to remove this test and keep the PR focused on the bug fix only.

pikaxinge:
Applied. I removed the unit test from this PR and kept the patch focused on the runtime fix only (duplicate checkpoint save removal). Also updated the PR title to satisfy the strict conventional-title check.

pikaxinge:
Rebased this branch onto latest `main` and kept the PR scoped to the runtime fix only. CI still needs the `run-ci` label (I cannot add labels with my current permissions); could a maintainer help add it?

anHappyDog (approved):

pikaxinge:
Update: after the rebase, `check-changes` failed again because the PR still lacks the `run-ci` label (`This pull request does not have the run-ci label`). I still do not have permission to add labels on this repo. Could a maintainer add `run-ci` so full CI can run?

anHappyDog:
cc @guozhen1997 .

PR #921, opened 2026-03-22
fix actor load bug

BitCalSaul:
**Description**
  Skip CUDA IPC and force GLOO CPU relay (GPU→CPU→GLOO→CPU→GPU) for colocated workers (same node, shared  GPU).

  - collective_group.py: _check_same_device_with_peer() always returns -1 to skip CUDA IPC
  - multi_channel_pg.py: detect colocated worker pairs and set _no_accel_ccl = True to force GLOO

  **Motivation and Context**

  In colocated placement (actor and rollout on the same GPU), CUDA IPC keeps GPU memory occupied even after
  offload. Switching to NCCL causes deadlock since CUDA kernels from two processes on the same GPU are serialized.
  GLOO CPU relay avoids both issues.

  **How has this been tested?**

  Tested on 8-GPU colocated config (actor/env/rollout all on GPU 0-7) with WAN world model + OpenVLA-OFT + GRPO.
  Model sync completes without hanging and training runs normally.

  **Types of changes**

  - Bug fix (non-breaking change which fixes an issue)

  **Checklist:**

  - My code follows the code style of this project.
  - My change requires a change to the documentation.
  - I have updated the documentation accordingly.
  - I have added tests to cover my changes.
  - All new and existing tests passed.

### DIFF
--- a/rlinf/scheduler/collective/collective_group.py
+++ b/rlinf/scheduler/collective/collective_group.py
@@ -1036,22 +1036,8 @@ class CollectiveGroup:
             int: -1 means no common device; 0 means have common devices, but not sure if the tensor will be on the same device (the worker has multiple devices); 1 means the two workers are on the same device.
 
         """
-        peer_devices = self._group_info.workers[self._peer_rank].available_accelerators
-        my_devices = self._group_info.workers[self._rank].available_accelerators
-
-        # Check if the peer is on the same node
-        if (
-            self._group_info.workers[self._peer_rank].cluster_node_rank
-            != self._group_info.workers[self._rank].cluster_node_rank
-        ):
-            return -1
-
-        # Check if the two device list has intersection
-        if not set(peer_devices).intersection(set(my_devices)):
-            return -1
-        if len(peer_devices) == 1 and len(my_devices) == 1:
-            return 1
-        return 0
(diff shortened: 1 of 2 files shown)


andylin-hao:
Thanks for raising the problem. But we need a better solution for addressing the memory issue instead of simply disabling this function. I'll take a look recently. I'll turn this PR into draft until there's a better solution.

andylin-hao:
Hi @BitCalSaul, I think I have a fix and it's been merged into the main branch. Could you try the main branch again to see if it's really fixed? Tests on my machine shows that the memory goes away, but it's known to behave differently on other machines.

BitCalSaul:
Hi @andylin-hao,

I pulled the latest main branch and ran tests, but the issue still persists.

I’ve uploaded my config and corresponding videos here for [reference:](https://drive.google.com/drive/folders/1tOrzMvE7f9zap10YpQz3lTYsNaTp0r-_?usp=sharing)

Please take a look when you have time. Let me know if you need any additional logs or details.

BitCalSaul:
@andylin-hao may i ask if you reproduce this?

guozhen1997:
Closing this PR due to prolonged inactivity. Feel free to reopen it if you’d like to continue working on it.

PR #952, opened 2026-03-27
fix: support perturbation types for libero-pro eval

iamxjy:
Previously LIBERO_PERTURBATION only worked when set to 'all'. Passing a
specific type like 'task', 'lan', 'object', or 'swap' does not work (goes
to the unperturbed base BDDL files).

Signed-off: iamxjy

### Description
When LIBERO_PERTURBATION is set to a specific type (task, lan, object, swap), the BDDL selection code silently falls through to the unperturbed original base files. Only "all" was handled. This adds the missing branch to filter BDDL directories by the specified perturbation type.

### Motivation and Context
valid_perts = ["_lan", "_object", "_swap", "_task"] was already defined in the original code, indicating per-type filtering was
intended but never completed. Without this fix, running --perturbation task produces the same results as running without any perturbation.

### How has this been tested?
Tested on LIBERO-10 with pro on --perturbation task.

### Types of changes

### Checklist:


zoeyuchao:
@iamxjy Thanks for your contribution! Could you fix the lint error? We will merge this pr asap.

zoeyuchao:
@iamxjy please fix the commit check error

PR #956, opened 2026-03-30
feat: add DOS-W1 realworld env

weiyunfei:
### Description
Add Dos-W1 realworld environment and a pick up example task.

### Motivation and Context

Try to intergrate a new realworld robot.

### How has this been tested?

It is tested on a DOS-W1 and a 4090 GPU server.
A test guide is [here](https://gist.github.com/weiyunfei/544cc5b33922e364994573fc396ecf4e).
The pipeline is working end-to-end, but the algorithm performance has not been optimized yet.

### Types of changes

- [x ] New feature (non-breaking change which adds functionality)

### Checklist:


requirements/install.sh
UltronAI, line 1: Why do you need conda? We use uv to manage our python env

rlinf/algorithms/__init__.py
```diff
@@ -11,4 +11,4 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from . import advantages, loss_scales, losses, toolcall_parsers  # noqa: F401
+from . import advantages, loss_scales, losses  # noqa: F401
```
UltronAI: Seems unnecessary and may break other parts like reasoning models.

toolkits/collect_dosw1_demo.py
UltronAI, line 1: We have added a dedicated env for data collection (refer to examples/embodiment/collect_real_data.py). You should reuse this pipeline instead of creating a new one. There should not be a dedicated script for DOSW1. Make it configurable.

rlinf/envs/realworld/common/keyboard/socket_keyboard_listener.py
```diff
@@ -0,0 +1,121 @@
...
+
+import os
+import socket
+import threading
+from pathlib import Path
+
+from rlinf.utils.logging import get_logger
+
+
+_DEFAULT_SOCKET_PATH = "/tmp/dosw1_keyboard.sock"
```
UltronAI: We should NOT add dosw1 path as default in this common util func

rlinf/utils/nested_dict_process.py
```diff
@@ -79,6 +79,9 @@ def concat_batch(data1, data2):
                 continue
             batch[key] = torch.cat([data1[key], data2[key]], dim=0)
         elif isinstance(value, dict):
+            # XXX: added this for dealing with different keys in demo data.
+            if key not in data2:
+                continue
```
UltronAI, line 114: Change XXX to Note as above. Add a warning for such skipping to avoid silent bugs.

rlinf/envs/realworld/realworld_env.py
```diff
@@ -83,20 +83,30 @@ def _create_env(self, env_idx: int):
             hardware_info=hardware_info,
             env_idx=env_idx,
         )
+        base_env = env.unwrapped
```
UltronAI: seems unnecessary changes.

rlinf/envs/realworld/realworld_env.py
```diff
@@ -83,20 +83,30 @@ def _create_env(self, env_idx: int):
...
                 env = KeyboardRewardDoneWrapper(env)
-
-        env = RelativeFrame(env)
-        env = Quat2EulerWrapper(env)
+        if self._supports_pose_wrappers(base_env):
+            env = RelativeFrame(env)
+            env = Quat2EulerWrapper(env)
         return env
 
+    @staticmethod
```
UltronAI: This requires a further careful check

tests/e2e_tests/embodied/dosw1_dummy_sac_mlp_pick.yaml
```diff
@@ -0,0 +1,162 @@
...
+
+cluster:
+  num_nodes: 1
+  component_placement:
+    actor,env,rollout: 0
+
+runner:
+  task_type: embodied
+  logger:
+    log_path: /workspace/results/dosw1_dummy_sac_mlp_pick
```
UltronAI: replace this hard-coded path with ../results/xxx

toolkits/dosw1_whole_start.sh
UltronAI, line 1: It's weird to add this script to RLinf. There is no airbot conda env mentioned before.

tests/e2e_tests/embodied/test_dosw1_env_e2e.py
UltronAI, line 1: weird to add such a script here. consider to move it to toolkits/realworld_check or directly remove it.

toolkits/dosw1_keyboard_relay.py
UltronAI, line 1: This functionality should be done in a env wrapper, rather than running a separate script.

rlinf/data/replay_buffer.py
UltronAI, line 1: why do you add this check?

UltronAI (changes requested):
Overall, this PR includes some changes that are not compatible with our latest code conventions, and introduces some unnecessary changes. You need to refactor the real-world data collection pipeline for dosw1 robot. 

Besides, no doc added to describe this new feature.

requirements/install.sh
weiyunfei: Conda is applied to manage python environment on DOS-W1 robot. I could drop the changes about using an existing conda env.

rlinf/algorithms/__init__.py
weiyunfei: This change was introduced as a temporary workaround to bypass the additional dependencies on DOS-W1 robot and should be reverted.

rlinf/data/replay_buffer.py
weiyunfei: This check is taken as the shape of tensor collected by in demo buffer does not match the input in replay buffer. This may be fixed if I just use `examples/embodiment/collect_real_data.py` for data colleciton.

rlinf/envs/realworld/common/keyboard/socket_keyboard_listener.py
weiyunfei: This should be parameterized.

toolkits/collect_dosw1_demo.py
weiyunfei: Done.

toolkits/dosw1_keyboard_relay.py
weiyunfei: Done.

toolkits/dosw1_whole_start.sh
weiyunfei: Done.

tests/e2e_tests/embodied/test_dosw1_env_e2e.py
weiyunfei: Done.

rlinf/utils/nested_dict_process.py
weiyunfei: I'll add a warning here.

rlinf/envs/realworld/realworld_env.py
weiyunfei: fixed.

rlinf/envs/realworld/realworld_env.py
weiyunfei: I'll remove the `unwrapped` to make it stronger.

weiyunfei:
The dos-w1 keyboard relay script is here.
[<script src="https://gist.github.com/weiyunfei/7ba27fed573dd79bc2ed3d85196384f9.js"></script>](https://gist.github.com/weiyunfei/7ba27fed573dd79bc2ed3d85196384f9)

lihm23:
**Installation & Dependency Issues**
1. One of the current installation methods `pip install -e.` is quite time-consuming. It is recommended to apply `uv pip install -e.` to accelerate this process.
2. There are some missing dependences in the current RLinf environment when the new codes are added, transformers, imageio and gymnasium included.
3. The latest version of transformers is incompatible with the new codes, leading import errors. Version 4.35.0 is a compatible alternative, though it will throw a warning, which seems not quite essential.

**Training Configuration Compatibility Issue**
Even when the custom training config file dosw1_pick_sac_mlp.yaml is created by strictly following the template provided in the test guide with only necessary parameter modifications, it still encounters Hydra version compatibility conflicts with the Hydra version pre-configured in the RLinf environment, resulting in hydra configuration or framework errors. Further verification is required to confirm the root cause of this issue.

weiyunfei:
I checked four points and made updates on this branch.
1. I'll update the installation part of the guide document. I used the `pip install -e .` since I was using conda environment on the DOS-W1.

2. The main problem was the dosw1 installation path and the guide. The previous guide could lead users to an incomplete environment, and the dosw1 env-only installer was manually installing only part of the embodied dependency set. I updated `requirements/install.sh` so that the dosw1 path now reuses the standard `embodied` extra, which already includes `transformers`, `imageio`, and `gymnasium`, and I kept only the dosw1-specific package installation in the local path. I also fixed the guide to use `bash requirements/install.sh embodied --env dosw1` and to activate `.venv` before running commands.

3. This PR has not changed any code about transformers. And I found the main branch of RLinf  also depends on transformers<=4.57.6 [ref](https://github.com/RLinf/RLinf/blob/main/pyproject.toml#L106). About `transformers==4.35.0`: I did not pin to that version in this round because I could not reproduce a dosw1-specific import error under the current repo environment. If there is a concrete failing stack trace for a specific code path, I can follow up with a targeted compatibility pin if needed. 

4. I could not reproduce a Hydra version compatibility failure with the current RLinf environment.  The dosw1 configs compose correctly with the repo's current Hydra/OmegaConf versions. I guess the hydra configuration errors u met is more consistent with creating a config from an incomplete snippet instead of modifying a full template. Therefore, I recommond to use this template `examples/embodiment/config/dosw1_pick_sac_flow.yaml` and I've update it in the guide document.

I've not changed `pyproject.toml` in this round, because the `embodied` extra already contains the relevant packages and the actual gap was in the dosw1 installation path and documentation.

If any other errors occured, please show me the logs or a screenshot.

rlinf/envs/realworld/realworld_env.py
```diff
@@ -92,11 +92,20 @@ def _create_env(self, env_idx: int):
                 env = KeyboardRewardDoneMultiStageWrapper(env)
             elif self.cfg.keyboard_reward_wrapper == "single_stage":
                 env = KeyboardRewardDoneWrapper(env)
-
-        env = RelativeFrame(env)
-        env = Quat2EulerWrapper(env)
+        if self._supports_pose_wrappers(env):
```
yushuang200109011: Please resolve the conflict here

weiyunfei: done

examples/embodiment/config/dosw1_pick_sac_flow.yaml
```diff
@@ -0,0 +1,243 @@
...
+    min_buffer_size: 2
+    sample_window_size: 200
+    auto_save: False
+
+  demo_buffer:
+    enable_cache: True
+    cache_size: 200
+    min_buffer_size: 1
+    sample_window_size: 200
+    load_path: "/mlp_vepfs/share/wyf/RLinf-open/logs/dosw1_collect_data/demos"  # Replace with your demo data path, or remove this section if not using RLPD
```
zoeyuchao: please remove the absolute path. You can refer to other yaml files.

examples/embodiment/config/dosw1_pick_sac_flow.yaml
```diff
@@ -0,0 +1,243 @@
...
+  group_name: "RolloutGroup"
+  backend: "huggingface"
+  enable_offload: False
+  pipeline_stage_num: 1
+
+  collect_transitions: True
+  collect_prev_infos: False
+
+  model:
+    model_path: "/mlp_vepfs/share/wyf/RLinf-open"  # https://huggingface.co/RLinf/RLinf-ResNet10-pretrained
```
zoeyuchao: +1

examples/embodiment/config/dosw1_pick_sac_flow.yaml
```diff
@@ -0,0 +1,243 @@
...
+  micro_batch_size: 256
+  global_batch_size: 256
+  seed: 1234
+  enable_offload: False
+
+  model:
+    model_type: "flow_policy"
+    input_type: "mixed"  # only mixed input can be used in real world
+
+    model_path: "/mlp_vepfs/share/wyf/RLinf-open"  # https://huggingface.co/RLinf/RLinf-ResNet10-pretrained
```
zoeyuchao: and this

requirements/install.sh
```diff
@@ -906,6 +909,20 @@ install_frankasim_env() {
     uv pip install -r "$serldir/franka_sim/requirements.txt"
 }
 
+install_dosw1_env() {
+    # Reuse the standard embodied extra so dosw1 picks up the same
+    # transformers/imageio/gymnasium dependency set as other embodied envs.
+    uv sync --extra embodied --active $NO_INSTALL_RLINF_CMD
+    uv pip install evdev opencv-python
+    # install dosw1-sdk
+    uv pip install ~/dos_w1/airbot/5.1.6/airbot_py-5.1.6-py3-none-any.whl
```
zoeyuchao: this absolute path is dangerous

zoeyuchao:
@weiyunfei Copilot added some comments, please solve them. Thanks!

Iron-Wph:
@weiyunfei 
Hello, thank you for your contribution. I encountered the following issues while reproducing your setup:

1. There is a syntax error in the `install.sh` script.
PixPin_2026-04-19_14-34-09


2. During the installation of the `dos_w1` environment, the dependency `airbot_py-5.1.6-py3-none-any.whl` is required, but no download or installation instructions for this file are provided.
PixPin_2026-04-19_14-31-57

Could you please fix these issues?

By the way, it would also be helpful to provide a user guide (e.g., in the docs) and some results demonstrating that the implementation works as expected.

examples/embodiment/config/dosw1_pick_sac_flow.yaml
weiyunfei: Fixed. Replaced the absolute path with the placeholder "/path/to/demo_data", consistent with other yaml configs.

examples/embodiment/config/dosw1_pick_sac_flow.yaml
weiyunfei: Fixed. Both rollout.model.model_path and actor.model.model_path now use "/path/to/model". Same in dosw1_pick_sac_flow_async.yaml.

examples/embodiment/config/dosw1_pick_sac_flow.yaml
weiyunfei: +1

requirements/install.sh
weiyunfei: Addressed together with Copilot comment below. The wheel and airbot_api source are pre-deployed on the DOS-W1 robot at `~/dos_w1/airbot/...`, so I keep that as the default, but now allow `DOSW1_SDK_WHEEL` / `DOSW1_API_PATH` env vars to override, and skip the SDK install with a warning instead of failing when the paths are missing (typical on a training server that talks to the robot over gRPC and does not need the local SDK).

requirements/install.sh
```diff
@@ -18,7 +18,7 @@ NO_ROOT=0
 NO_INSTALL_RLINF_CMD="--no-install-project"
 SUPPORTED_TARGETS=("embodied" "agentic" "docs")
 SUPPORTED_MODELS=("openvla" "openvla-oft" "openpi" "gr00t" "dexbotic" "starvla" "lingbotvla" "dreamzero")
-SUPPORTED_ENVS=("behavior" "maniskill_libero" "metaworld" "calvin" "isaaclab" "robocasa" "franka" "frankasim" "robotwin" "habitat" "opensora" "wan" "xsquare_turtle2" "liberopro" "liberoplus" "roboverse" "embodichain" "d4rl")
+SUPPORTED_ENVS=("behavior" "maniskill_libero" "metaworld" "calvin" "isaaclab" "robocasa" "franka" "frankasim" "robotwin" "habitat" "opensora" "wan" "xsquare_turtle2" "liberopro" "liberoplus" "roboverse" "embodichain" "d4rl", "dosw1")
```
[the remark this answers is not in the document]
weiyunfei: Fixed.

requirements/install.sh
```diff
@@ -987,6 +989,20 @@ install_embodichain_env() {
...
 }
 
+install_dosw1_env() {
+    # Reuse the standard embodied extra so dosw1 picks up the same
+    # transformers/imageio/gymnasium dependency set as other embodied envs.
+    uv sync --extra embodied --active $NO_INSTALL_RLINF_CMD
+    uv pip install evdev opencv-python
+    # install dosw1-sdk
+    uv pip install ~/dos_w1/airbot/5.1.6/airbot_py-5.1.6-py3-none-any.whl
+    uv pip install -e ~/dos_w1/airbot/airbot_api
```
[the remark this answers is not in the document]
weiyunfei: Refactored install_dosw1_env: paths default to the DOS-W1 preset locations (`~/dos_w1/airbot/5.1.6/airbot_py-*.whl` and `~/dos_w1/airbot/airbot_api`), but can be overridden via `DOSW1_SDK_WHEEL` and `DOSW1_API_PATH`. If either path is missing, I now log a warning and skip just that step rather than aborting the whole install, so generic servers can still set up the env without the robot-local SDK. You could get the paths from the [test guide doc](https://gist.github.com/weiyunfei/544cc5b33922e364994573fc396ecf4e) I've provided in this PR.

rlinf/envs/realworld/dosw1/dosw1_env.py
```diff
@@ -0,0 +1,686 @@
...
+            self.in_free_teleop = True
+            self.start_episode_requested = False
+            self._set_leader_follow_enabled(
+                enabled=True, source="reset_enter_free_teleop"
+            )
+            self._logger.info(
+                "[DOSW1Env] FreeTeleop mode active. "
+                "Move arms freely via leader arm. Press 's' to start episode."
+            )
+            self._free_teleop_loop()
```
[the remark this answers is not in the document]
weiyunfei: Fixed. `DOSW1Env.reset` now parses `options.get("skip_wait_for_start", False)`; when True it still re-enables leader-follow but bypasses `_free_teleop_loop()`, so the terminal reset in `collect_real_data.py` no longer blocks on the 's' key.

rlinf/utils/nested_dict_process.py
```diff
@@ -15,7 +15,9 @@
 from typing import Any
 
 import torch
+from rlinf.utils.logging import get_logger
 
+logger = get_logger()
 
 def update_nested_cfg(base_cfg, override_cfg):
```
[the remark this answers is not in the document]
weiyunfei: Fixed. Removed the module-level `logger = get_logger()`. The only call site that actually needs a logger (the missing-key warning in concat_batch) now does a lazy from `rlinf.utils.logging import get_logger` inside the branch, and it fires at most once per missing key per process. This way `rlinf.utils.nested_dict_process` no longer pulls in the `scheduler/worker` stack on import, while log output still goes through the standard RLinf logger so level filtering / formatting stays consistent with the rest of the codebase.

rlinf/utils/nested_dict_process.py
```diff
@@ -90,6 +92,10 @@ def concat_batch(data1, data2):
                 continue
             batch[key] = torch.cat([data1[key], data2[key]], dim=0)
         elif isinstance(value, dict):
+            # NOTE: added this for dealing with different keys in demo data.
+            if key not in data2:
+                logger.warning(f"Key {key} not found in data2, value type: {type(value)}, skipping...")
+                continue
```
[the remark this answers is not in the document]
weiyunfei: Fixed. The log level is kept and a module-level `_CONCAT_BATCH_WARNED_KEYS` set ensures each missing key only produces a single warning per process regardless of how many times `concat_batch` runs.

rlinf/envs/realworld/common/wrappers/leader_follower_keyboard_intervention.py
```diff
@@ -0,0 +1,128 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import gymnasium as gym
+
```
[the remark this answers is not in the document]
weiyunfei: Fixed.

rlinf/envs/realworld/dosw1/dosw1_env.py
```diff
@@ -0,0 +1,686 @@
...
+        for camera in self._cameras:
+            frame_rgb = camera.get_frame()
+            height, width = frame_rgb.shape[:2]
+            crop = min(height, width)
+            start_x = (width - crop) // 2
+            start_y = (height - crop) // 2
+            cropped = frame_rgb[start_y : start_y + crop, start_x : start_x + crop]
+            resized = cv2.resize(cropped, (IMAGE_W, IMAGE_H))
+            frames[camera._camera_info.name] = resized[..., ::-1]
+            display_frames[camera._camera_info.name] = resized
```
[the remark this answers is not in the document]
weiyunfei: Fixed.

PR #968, opened 2026-03-31
feat: support torch.compile for pi0 and pi0.5 predict

thereAreDemonsNearby:
### Description
Support torch.compile for pi0.5 rollout

### Motivation and Context
Pi0.5 predict can be accelerated ~30% by enabling torch.compile 

### How has this been tested?
success_rate_plot


rlinf/workers/rollout/hf/huggingface_worker.py
```diff
@@ -113,7 +113,8 @@ def init_worker(self):
         if self.expert_model is not None:
             self.expert_model.eval()
 
-        if self.cfg.rollout.get("enable_torch_compile", False):
+        self.enable_torch_compile = self.cfg.rollout.get("enable_torch_compile", False)
+        if self.enable_torch_compile:
             mode = self.cfg.rollout.get(
                 "torch_compile_mode", "max-autotune-no-cudagraphs"
```
FxxxxU, line 122: Is it necessary to add an assert for torch_compile_mode not being null?

thereAreDemonsNearby: I think it's not necessary because **self.cfg.rollout.get("torch_compile_mode", "max-autotune-no-cudagraphs")** provides a quite good default value "max-autotune-no-cudagraphs"

examples/embodiment/config/libero_10_ppo_openpi_pi05.yaml
```diff
@@ -111,6 +111,10 @@ rollout:
   enable_offload: False
   pipeline_stage_num: 1
 
+  # you can enable torch compile by uncomment the following 2 lines:
+  # enable_torch_compile: True
```
FxxxxU, line 116: Should add CI for this feature? Just place it on an existing CI branch.

thereAreDemonsNearby: you are right, I will do this

FxxxxU (approved):

tests/e2e_tests/embodied/run.sh
```diff
@@ -34,4 +34,5 @@ else
     echo "Evaluation Mode: Standard LIBERO"
 fi
 
-python ${REPO_PATH}/examples/embodiment/train_embodied_agent.py --config-path ${REPO_PATH}/tests/e2e_tests/embodied --config-name ${CONFIG}
\ No newline at end of file
+shift 2
```
anHappyDog: Why add this?

rlinf/workers/rollout/hf/huggingface_worker.py
```diff
@@ -117,7 +117,8 @@ def init_worker(self):
         if self.expert_model is not None:
             self.expert_model.eval()
 
-        if self.cfg.rollout.get("enable_torch_compile", False):
+        self.enable_torch_compile = self.cfg.rollout.get("enable_torch_compile", False)
```
anHappyDog: If variable `enabled_torch_compile` is only used here, it might be better to just use `cfg` to get the value,  the name is also similar to `torch_compile_enabled`, which means the model has already been compiled.

rlinf/models/embodiment/openpi/openpi_action_model.py
```diff
@@ -774,7 +776,7 @@ def sample_mean_var_val(
         bsize = state.shape[0]
         device = state.device
         if isinstance(idx, int):
-            idx = torch.tensor(idx).expand(bsize)
+            idx = torch.full((), idx, device=device).expand(bsize)
```
anHappyDog, line 779: Could this change bring some performance improvement?

tests/e2e_tests/embodied/run.sh
thereAreDemonsNearby: I use **shift** to skip \\$1 and \\$2,  and then \$@ is all the remaining arguments to this script. Those arguments will be passed to train_embodied_agent.py. An example in my commits: [‎.github/workflows/embodied-e2e-tests.yml](https://github.com/RLinf/RLinf/pull/968/changes/BASE..33962c5f0bc63d2f4e6fdd986642bb82f0de8fd0#diff-ef0b92363681764044623b7ef3bd713a107b944005cc4ce021222a7014231599).
But there is indeed a bug: if CONFIG and BACKEND are all passed in, then shift 2 is correct; if only CONFIG is passed in, we only need to shift 1. I fixed this and please review this

thereAreDemonsNearby: Do you have any suggestion about implementing this?

rlinf/workers/rollout/hf/huggingface_worker.py
thereAreDemonsNearby: These few lines of code were likely leftovers from the previous revision and have been deleted.

rlinf/models/embodiment/openpi/openpi_action_model.py
thereAreDemonsNearby: Yes, absolutely. We want this **idx** to be on GPU, so that the following **timesteps[idx]** operation is entirely on GPU and not involves h2d synchronization. Furthermore, **torch.full((), scalar_value, device='cuda')** is prefered to **torch.tensor(scalar_value, device='cuda')** because the latter API involves h2d (at least in torch2.6), and the former one pass the scalar value through cuda kernel argument, thus h2d is avoided.

anHappyDog: ok.

tests/e2e_tests/embodied/run.sh
anHappyDog: no..

anHappyDog (approved):

.github/workflows/embodied-e2e-tests.yml
```diff
@@ -669,6 +669,14 @@ jobs:
           source .venv/bin/activate
           bash tests/e2e_tests/embodied/run.sh libero_spatial_ppo_openpi05
 
+      - name: Libero-Spatial PPO Pi05 torch compile test
+        timeout-minutes: 20
+        run: |
+          unset PYTHONPATH
+          export REPO_PATH=$(pwd)
+          source .venv/bin/activate
+          bash tests/e2e_tests/embodied/run.sh libero_spatial_ppo_openpi05 +rollout.enable_torch_compile=true +rollout.torch_compile_mode=max-autotune
```
andylin-hao, line 678: Why not have a new yaml for the test, rather than hacking run.sh?

andylin-hao: On second thought, maybe that's fine.

guozhen1997 (approved):

andylin-hao (approved):

PR #972, opened 2026-03-31
fix: potential gpu mem leak in weight sync

zanghz21:
### Description

Add cuda empty cache in actor sync model. 

### Motivation and Context

In sync_weight_no_wait scenario, we observed GPU memory leak in the actor. After this modification, I do not observe it any more. 

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/workers/actor/fsdp_actor_worker.py
+++ b/rlinf/workers/actor/fsdp_actor_worker.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import gc
 import os
 import time
 from functools import partial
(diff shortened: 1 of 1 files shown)


rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -1084,6 +1085,8 @@ async def sync_model_to_rollout(self) -> None:
             await handle.async_wait()
         if self.enable_offload and not self.is_weight_offloaded:
             self.offload_param_and_grad()
+        gc.collect()
```
andylin-hao, line 1088: Call the clear_memory() function in utils

PR #1013, opened 2026-04-09
feat: add decoupled mode for rollout and env workers

FxxxxU:
Support rollout workers and reward workers continuing to operate when env workers enter a calm-down state.
Fix the deadlock issue in rollout workers and reward workers caused by env worker going offline.

### Description

This PR primarily adds a decoupled mode to the env worker to address potential deadlock issues that may arise in real-world environments. 

### Motivation and Context

Currently, there is an implicit mapping logic between the env worker and the rollout worker, which causes a deadlock among all workers in that data parallel if an env worker fails. 
The new feature allows the env worker to no longer generate this mapping logic during training, ensuring that other workers can continue training normally even if the current env worker exits due to some special circumstance.

### How has this been tested?

Yes. I test this pr with maniskill_sac_mlp_async_decoupled.yaml, frankasim_sac_cnn_async.yaml, maniskill_async_ppo_mlp.yaml and maniskill_async_ppo_openpi.yaml. After training, the `success_once` value obtained for each script was greater than 0. I tested maniskill_async_ppo_openpi.yaml with env decoupled mode on a single 8-card H100 machine, and the results are shown below:
截屏2026-04-20 16 36 40

### Types of changes

### Checklist:


rlinf/workers/rollout/hf/async_huggingface_worker.py
```diff
@@ -182,3 +197,255 @@ async def request_actor_sync_model(self):
...
+                f"Expected env output batch size {expected_size} get the batch_index {batch_index}, "
+                f"got {actual_size}."
+            )
+            obs_batches.append(obs_batch["batch"])
+        return self._merge_obs_batches(obs_batches)
+
+    async def decoupled_generate_one_epoch(
+        self, input_channel: Channel, output_channel: Channel
+    ):
+        self.update_dagger_beta()
```
yushuang200109011, line 199: It seems like should be placed in the infinite loop.

FxxxxU: Once this new feature is enabled, rollout workers enter a stateless mode and accept data from any number of env workers. To improve efficiency and prevent rollout workers that finish processing quickly from exiting prematurely, they operate in an infinite loop.

rlinf/workers/env/env_worker.py
```diff
@@ -122,11 +123,36 @@ def __init__(self, cfg: DictConfig):
...
+            self.log_info(
+                f"Env worker initialized with dst_rank_map: {self.dst_rank_map}"
+            )
+            self.log_info(
+                f"Env worker initialized with src_rank_map: {self.src_rank_map}"
+            )
 
-        self.log_info(f"Env worker initialized with dst_rank_map: {self.dst_rank_map}")
-        self.log_info(f"Env worker initialized with src_rank_map: {self.src_rank_map}")
+        train_env_cls = get_env_cls(self.cfg.env.train.env_type, self.cfg.env.train)
```
guozhen1997: L154-155 should be removed, since the `train_env_cls` and `eval_env_cls` are obtained in L167 and L174.

rlinf/workers/env/env_worker.py
```diff
@@ -122,11 +123,36 @@ def __init__(self, cfg: DictConfig):
...
+            self.dst_rank_map = self._setup_eval_dst_rank_map()
+            self.src_rank_map = self._setup_eval_src_rank_map()
+            self.log_info(
+                f"Env worker initialized with dst_rank_map for evaluation: {self.dst_rank_map}"
+            )
+            self.log_info(
+                f"Env worker initialized with src_rank_map for evaluation: {self.src_rank_map}"
+            )
+        else:
+            self.dst_rank_map = self._setup_dst_rank_map()
```
guozhen1997, line 139: This if and else part has duplicated logic for setting and logging dst_rank_map and src_rank_map.

guozhen1997 (commented):
I’ve reviewed this PR and suggest not merging it into main for now, mainly for two reasons:

1. The core functionality—supporting automatic online/offline switching for robot devices at the worker level—is incomplete due to missing logic in the scheduler.
2. The current decoupled mode introduces ~1000 lines of mapping and send/recv code across env/rollout/reward workers. This is planned to be refactored in PR 1078, where these will be abstracted into a unified scheduler interface to reduce redundancy.

Therefore, I recommend holding off on merging this PR for now. I’ll continue pushing forward the refactor based on this PR.

PR #1059, opened 2026-04-18
feat: realworld sft pi0 deploy

git-xuxin:
### Types of changes

### Checklist:


zoeyuchao:
xuxin confirmed the real world exp.

rlinf/models/embodiment/openpi/policies/realworld_policy.py
```diff
@@ -0,0 +1,124 @@
...
+    Dual view: observation/image + observation/wrist_image.
+    State: 7-dim [tcp_pose(6) + gripper(1)].
+    Action: 7-dim [delta_xyz(3) + delta_rpy(3) + gripper(1)].
+    """
+
+    model_type: _model.ModelType
+
+    def __call__(self, data: dict) -> dict:
+        base_image = _parse_image(data["observation/image"])
+        left_wrist_image, right_wrist_image = _parse_wrist_images(
```
yushuang200109011, line 76: This function returns only one real image and a full-zero image. And your image mask is all true. That means the full zero-image will be passed into model, which should not been passed. refer to line83 in `rlinf/models/embodiment/openpi/policies/franka_dagger_policy.py`

And a further question is: This config appears to have the same functionality as dagger's config. If you want a generic config, you could simply rename the dagger config to something like dual_view_franka_standard_config.

PR #1082, opened 2026-04-22
feat(realworld): add Turtle2 generic deploy env

MuggleZzzH:
### Description
This PR adds a generic real-world deploy environment path for XSquare Turtle2 so that Turtle2 can be used as a standard real-world deployment platform in RLinf.

This change mainly includes:
- a generic Turtle2 deploy environment config fragment: `examples/embodiment/config/env/realworld_turtle2_deploy.yaml`
- the definition and registration of `Turtle2DeployEnv-v1`
- dual-arm pose action wrappers with `relative_pose` as the default and `absolute_pose` available when explicitly selected
- Turtle2 deployment documentation in both Chinese and English

### Motivation and Context
RLinf already supports real-world training paths for Turtle2, but Turtle2 did not yet have a generic deploy-environment entry point suitable for policy-only rollout or evaluation.

The goal of this PR is to add that deployment support while keeping the existing Turtle2 naming and compatibility intact. The default deploy action mode follows Turtle2's existing relative-pose action semantics; absolute-pose commands are used only when `action_mode: absolute_pose` is configured.

### How has this been tested?
This change was validated in the following ways:

1. Scope note
- Hardware-level e2e coverage for the new deploy env requires a physical Turtle2 setup and is deferred to a follow-up PR.
2. Real-robot validation

- We ran a towel-folding experiment on a physical Turtle2 setup to validate the new deploy environment.

- The system was able to deploy and run successfully on the real robot.

- The resulting task performance was not yet ideal, so further validation, debugging, and tuning are still in progress.

This PR only introduces generic deploy-environment support and does not include task-specific debug logs.

### Types of changes

### Checklist:


Brunch-Life:
I found that your env init didn't have transform from yaml to ndarray, like this::
`@dataclass
class Turtle2DeployEnvConfig(Turtle2RobotConfig):
    ...
    def __post_init__(self):
        self.reset_ee_pose     = np.array(self.reset_ee_pose).reshape(2, 6)
        self.ee_pose_limit_min = np.array(self.ee_pose_limit_min).reshape(2, 6)
        self.ee_pose_limit_max = np.array(self.ee_pose_limit_max).reshape(2, 6)
`
maybe you should check it

Brunch-Life:
and maybe you need support many kinds of wrapper?(Like KeyboardRewardWrapper etc.) You should check it.

MuggleZzzH:
> I found that your env init didn't have transform from yaml to ndarray, like this:: `@dataclass class Turtle2DeployEnvConfig(Turtle2RobotConfig): ... def __post_init__(self): self.reset_ee_pose = np.array(self.reset_ee_pose).reshape(2, 6) self.ee_pose_limit_min = np.array(self.ee_pose_limit_min).reshape(2, 6) self.ee_pose_limit_max = np.array(self.ee_pose_limit_max).reshape(2, 6) ` maybe you should check it

fixed

Brunch-Life:
Turtle2-> X1

rlinf/envs/realworld/common/wrappers/dual_pose_action.py
```diff
@@ -0,0 +1,73 @@
...
+        super().__init__(env)
+        self._step_absolute_pose = self._get_required_callable("step_absolute_pose")
+        self.action_space = self._get_required_callable(
+            "get_absolute_pose_action_space"
+        )()
+
+    def step(self, action: np.ndarray):
+        return self._step_absolute_pose(action)
+
+    def _get_required_callable(self, name: str) -> Callable:
```
Brunch-Life: no need to raise error

rlinf/envs/realworld/common/wrappers/dual_pose_action.py
```diff
@@ -0,0 +1,73 @@
...
+        super().__init__(env)
+        self._step_relative_pose = self._get_required_callable("step_relative_pose")
+        get_action_space = getattr(env, "get_relative_pose_action_space", None)
+        if callable(get_action_space):
+            self.action_space = get_action_space()
+
+    def step(self, action: np.ndarray):
+        return self._step_relative_pose(action)
+
+    def _get_required_callable(self, name: str) -> Callable:
```
Brunch-Life: same

rlinf/envs/realworld/common/wrappers/dual_euler_obs.py
```diff
@@ -35,6 +36,10 @@ def __init__(self, env: Env):
     def observation(self, observation: dict) -> dict:
         """Convert dual-arm quaternion TCP pose to euler angles in-place."""
         tcp_pose = observation["state"]["tcp_pose"]
+        if tcp_pose.shape[-1] != 14:
+            raise ValueError(
+                f"DualQuat2EulerWrapper expects 14D tcp_pose, got shape {tcp_pose.shape}."
+            )
```
Brunch-Life, line 58: no need

rlinf/envs/realworld/common/wrappers/dual_pose_action.py
MuggleZzzH: fixed

rlinf/envs/realworld/common/wrappers/dual_pose_action.py
MuggleZzzH: fixed

rlinf/envs/realworld/common/wrappers/dual_euler_obs.py
MuggleZzzH: fixed

MuggleZzzH:
> Turtle2-> X1

all done

MuggleZzzH:
Thanks for the thorough review! Here's a summary of everything addressed:

**1. YAML → ndarray transform missing in env init**
Added `__post_init__` to `X1RobotConfig` to convert `reset_ee_pose`, `ee_pose_limit_min`, and
`ee_pose_limit_max` from YAML lists to numpy arrays with the correct shapes.

**2. Unnecessary error raises in pose action wrappers**
Removed the custom `_get_required_callable` guard and its raise logic in both
`DualAbsolutePoseActionWrapper` and `DualRelativePoseActionWrapper`. Now calling
`self.get_wrapper_attr(...)` directly — it already raises `AttributeError` when the attribute is
missing, so no extra layer is needed. Also removed the resulting unused `Callable` import (Ruff F401).

**3. Shape validation in `DualQuat2EulerWrapper`**
Removed the explicit `assert obs["tcp_pose"].shape == (14,)` check per reviewer feedback.

**4. Wrapper coverage — `KeyboardRewardWrapper` and `DualRelativeFrame`**
- Added `keyboard_reward_wrapper` support to `apply_dual_pose_action_wrappers`, consistent with
  how Franka and other builders handle it (labeling is needed during deploy/collect, not just eval).
- Added `DualRelativeFrame` to the `relative_pose` path (behind a `use_relative_frame` flag,
  default `True`) to keep train/deploy coordinate frames aligned — mirrors how single-arm
  `RelativeFrame` is used in other builders.

**5. Full Turtle2 → X1 rename**
Renamed all internal references across:
- Python classes/constants: `X1Env`, `X1RobotConfig`, `X1Robot`, `X1HWInfo`, `HW_TYPE = "X1"`, etc.
- File names: `turtle2_env.py` → `x1_env.py`, `turtle2_smooth_controller.py` →
  `x1_smooth_controller.py`, etc.
- Configs/YAMLs: `realworld_x1_deploy.yaml`, `realworld_button_x1_sac_cnn.yaml`, etc.
- CI workflows: install step name, e2e test job/step names, and test YAML references
- `pyproject.toml` extra: `xsquare_turtle2` → `xsquare_x1`
- `requirements/install.sh`: `SUPPORTED_ENVS` list and install function name

> **Note:** External SDK references (`turtle2_basic`, `xsquare_turtle_basics` GitHub repo) are
> intentionally kept unchanged — they point to third-party packages outside this repo.

Brunch-Life (approved):

UltronAI:
Is it x1 or x2? You use x2 in your description but x1 in code.

MuggleZzzH:
> Is it x1 or x2? You use x2 in your description but x1 in code.

X1, fixed

rlinf/envs/realworld/common/wrappers/dual_pose_action.py
```diff
@@ -0,0 +1,51 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Dual-arm pose action-mode wrappers."""
+
+from __future__ import annotations
```
UltronAI: why do you need this import? no such import in dual_euler_obs and dual_relative_frame

rlinf/envs/realworld/common/wrappers/dual_pose_action.py
```diff
@@ -0,0 +1,51 @@
...
+        return self._step_absolute_pose(action)
+
+
+class DualRelativePoseActionWrapper(gym.Wrapper):
+    """Route dual-arm relative pose actions to an env relative-pose step method."""
+
+    def __init__(self, env: gym.Env):
+        super().__init__(env)
+        self._step_relative_pose = self.get_wrapper_attr("step_relative_pose")
+        get_action_space = getattr(env, "get_relative_pose_action_space", None)
```
UltronAI: why not use `self.get_wrapper_attr("get_relative_pose_action_space")()`

rlinf/envs/realworld/xsquare/tasks/deploy_env.py
```diff
@@ -0,0 +1,152 @@
...
+        reward = self._calc_step_reward(observation)
+        terminated = False
+        truncated = self._num_steps >= self.config.max_num_steps
+        return observation, reward, terminated, truncated, {}
+
+    def step_relative_pose(self, action: np.ndarray):
+        return super().step(action)
+
+    def step(self, action: np.ndarray):
+        return self.step_relative_pose(action)
```
UltronAI: The factory `create_x1_deploy_env` always wraps with `apply_dual_pose_action_wrappers` which overrides step, so this default never fires through normal use. But anyone instantiating `X1DeployEnv` directly (e.g. unit tests, custom factories, eval scripts) silently gets the relative-pose path — the absolute-pose method is unreachable from the env's own step.

Suggested fix: either add `action_mode: str = "absolute_pose"` to `X1DeployEnvConfig` and dispatch on it in `step`, or `raise NotImplementedError("X1DeployEnv.step requires apply_dual_pose_action_wrappers; use create_x1_deploy_env or DualAbsolute/RelativePoseActionWrapper directly.")`. Either way the behavior must be unambiguous without the wrapper.

rlinf/envs/realworld/xsquare/tasks/deploy_env.py
```diff
@@ -0,0 +1,152 @@
...
+    def __post_init__(self):
+        self.target_ee_pose = np.asarray(self.target_ee_pose, dtype=np.float64).reshape(
+            2, 6
+        )
+        self.reset_ee_pose = np.asarray(self.reset_ee_pose, dtype=np.float64).reshape(
+            2, 6
+        )
+        self.ee_pose_limit_min = np.asarray(
+            self.ee_pose_limit_min, dtype=np.float64
+        ).reshape(2, 6)
```
UltronAI: kind of hard code

rlinf/envs/realworld/common/wrappers/apply.py
```diff
@@ -137,3 +141,32 @@ def apply_dual_arm_wrappers(env: gym.Env, cfg: Mapping[str, Any]) -> gym.Env:
...
     return env
+
+
+def apply_dual_pose_action_wrappers(env: gym.Env, cfg: Mapping[str, Any]) -> gym.Env:
+    """Wrapper stack for dual-arm pose-action deploy envs.
+
+    For ``relative_pose`` mode, ``DualRelativeFrame`` is applied by default
+    (``use_relative_frame=True``) so that the policy sees EE-frame observations
+    and its EE-frame delta actions are transformed to base frame — matching the
+    training wrapper contract.  Pass ``use_relative_frame=False`` to skip.
```
UltronAI: Consider to rephase to "mirrors the dual-arm tail of apply_dual_arm_wrappers (DualRelativeFrame + DualQuat2EulerWrapper) but skips teleop and gripper-close wrappers, since deploy envs don't run human interventions.". But I think you will add dagger support in your next PR.

UltronAI:
**Overall, the renaming operation from Turtle2 to X1 causes lots of backward-incompatibility. Reconsider if it's necessary and carefully check everything that may be affected. I am not sure if I have found all incompatible things. I would recommend you to keep Turtle2 for safety, otherwise a much more thorough code review should be conducted.**

- The renamed docs/source-en/rst_source/examples/embodied/xsquare_x1.rst (and ZH counterpart) only document the original SAC + CNN button-pressing training task. The PR introduces three new user-facing surfaces that get no documentation: (1) the env id `X1DeployEnv-v1`, (2) the env config `examples/embodiment/config/env/realworld_x1_deploy.yaml`, and the wrapper builder `apply_dual_pose_action_wrappers` and its `action_mode: absolute_pose | relative_pose toggle`. You'd better to extend `xsquare-x1.rst` with a Deployment section or add a new page (cf. franka_pi0_sft_deploy.rst)

- Both index.rst files now point at `https://github.com/RLinf/misc/raw/main/pic/xsquare_x1_arm_small.jpg` but it does NOT exist. It's in another repo and has not be updated accordingly. Be careful for such renaming operations. Please confirm xsquare_x1_arm_small.jpg exists in RLinf/misc/main:pic/ before merging; if not, upload it (or temporarily keep the old xsquare_turtle2_arm_small.jpg URL until it does).

- The description still has [ ] I have added tests to cover my changes, but `tests/unit_tests/test_dual_pose_action_wrappers.py` is part of this PR.

- Vendor names "turtle" still appear in the post-PR state without explanation. After the rename, two locations still say `turtle`: `rlinf/envs/realworld/xsquare/x1_smooth_controller.py:~31` — from `turtle2_basic.turtle2_controller.Turtle2Controller` import `Turtle2Controller as X1Controller` (vendor SDK). and `requirements/install.sh:~917` (after rename) — `uv pip install git+${GITHUB_PREFIX}https://github.com/RLinf/xsquare_turtle_basics.git inside install_xsquare_x1_env`. These are correct (the upstream package and repo are still named `turtle2_basic` / `xsquare_turtle_basics`), but a future maintainer will assume they're stale rename leftovers.

- No e2e test for the new `X1DeployEnv-v1`

rlinf/envs/realworld/common/wrappers/dual_pose_action.py
MuggleZzzH: Done. I removed the unused `from __future__ import annotations` import from `dual_pose_action.py` to keep it consistent with `dual_euler_obs.py` and `dual_relative_frame.py`.

rlinf/envs/realworld/common/wrappers/dual_pose_action.py
MuggleZzzH: Done. `DualRelativePoseActionWrapper` now resolves both `step_relative_pose` and `get_relative_pose_action_space` through `self.get_wrapper_attr(...)`, so it works correctly through wrapper chains and makes the relative-pose env contract explicit.

rlinf/envs/realworld/xsquare/tasks/deploy_env.py
MuggleZzzH: Done. I added `action_mode` to `Turtle2DeployEnvConfig` and `step()` now dispatches explicitly between `step_relative_pose` and `step_absolute_pose`.

I kept `relative_pose` as the default to preserve Turtle2's existing relative-action behavior. `absolute_pose` is only used when explicitly configured. The factory also propagates top-level `env_cfg.action_mode` into `override_cfg`, so direct env behavior and wrapper behavior stay aligned.

rlinf/envs/realworld/xsquare/tasks/deploy_env.py
MuggleZzzH: Done. I extracted `NUM_ARMS`, `POSE_DIM`, and `ACTION_DIM_PER_ARM`, then reused them for pose reshaping, action shape checks, and per-arm indexing.

rlinf/envs/realworld/common/wrappers/apply.py
MuggleZzzH: Done. I rephrased the docstring to say this mirrors the dual-arm tail of `apply_dual_arm_wrappers` while skipping teleop and gripper-close wrappers because deploy envs do not run human interventions.

DAgger support will be handled separately in a follow-up PR.

rlinf/envs/realworld/common/wrappers/dual_pose_action.py
UltronAI, line 1: Do we actually need these two wrappers? They are only used by your Turtle2 env now and can be configured via env config inside the Turtle2 env.

rlinf/envs/realworld/xsquare/turtle2_env.py
```diff
@@ -139,6 +153,23 @@ def __init__(
         self._check_cameras()
         # Video player for displaying camera frames
 
+    def _normalize_config_arrays(self) -> None:
```
UltronAI, line 155: Do we actually need such process? It may silently change the config content. How will these values be transformed (considering type, shape, etc.) ?

rlinf/envs/realworld/xsquare/tasks/__init__.py
```diff
@@ -19,10 +19,52 @@
...
 from rlinf.envs.realworld.xsquare.tasks.button_env import (
     ButtonEnv as ButtonEnv,
 )
+from rlinf.envs.realworld.xsquare.turtle2_env import (
+    Turtle2Env,
+    Turtle2RobotConfig,
+)
+
+
+def _resolve_deploy_action_mode(
```
UltronAI: This is related to the above wrapper issue. You may consider to delete dual_pose_action.py and _resolve_deploy_action_mode; collapse the two pose-action branches in apply_dual_arm_wrappers.

The two wrappers reset action_space and forward step() to step_{relative,absolute}_pose, which Turtle2Env already does itself from config.action_mode — they are no-ops on a correctly-configured env, and apply.py:136 / apply.py:149 are their only callers. Removing them leaves apply_dual_arm_wrappers with one consumer of action_mode (gating DualRelativeFrame + the no-teleop assertion), readable from env.get_wrapper_attr("config").action_mode the same way env.config.is_dummy already is. With the wrapper-cfg side gone, _resolve_deploy_action_mode seems redundant.

Note that if you update these things, you should update the doc together.

rlinf/envs/realworld/common/wrappers/apply.py
```diff
@@ -128,20 +133,59 @@ def apply_single_arm_wrappers(env: gym.Env, cfg: Mapping[str, Any]) -> gym.Env:
...
         raise NotImplementedError(
-            "no_gripper=True is not yet supported for dual-arm envs: "
-            "DualGripperCloseEnv is not implemented. "
-            "Set env.eval.no_gripper=False (or env.train.no_gripper=False)."
+            "Dual-arm realworld wrappers require no_gripper=False. "
+            "Pose-action modes use 7D per-arm actions including gripper, "
+            "and delta_axis_angle mode does not have DualGripperCloseEnv yet."
         )
 
+    if action_mode == "relative_pose":
```
UltronAI: Consider if you can do the switch between relative_pose and abs pose inside the turtle2 env so that these outer changes can be removed or simplified.

rlinf/envs/realworld/common/wrappers/relative_frame.py
```diff
@@ -52,6 +52,15 @@ def __init__(self, env: Env, include_relative_pose=True):
             # Homogeneous transformation matrix from reset pose's relative frame to base frame
             self.T_b_r_inv = np.zeros((4, 4))
 
+    @staticmethod
```
UltronAI, line 55: Why do we need this silent slice opeartion? It may cause invisible bugs to use wrong values.

UltronAI (changes requested):
Overall, some changes are introduced due to the relative/absolute switch, which can be simplified. and currently it's only related to the specific turtle2 env.

yushuang200109011 (approved):

UltronAI (approved):

guozhen1997 (approved):

examples/embodiment/config/realworld_turtle2_deploy_eval.yaml
```diff
@@ -0,0 +1,115 @@
...
+  output_subdir: null
+  searchpath:
+    - file://${oc.env:EMBODIED_PATH}/config/
+
+cluster:
+  num_nodes: 1
+  component_placement:
+    actor:
+      node_group: turtle2
+      placement: 0
```
QuanluZhang, line 21: This is an eval task, do we still need to config `actor` here?

rlinf/envs/realworld/common/wrappers/apply.py
```diff
@@ -128,20 +129,55 @@ def apply_single_arm_wrappers(env: gym.Env, cfg: Mapping[str, Any]) -> gym.Env:
 
     if cfg.get("use_relative_frame", True):
         env = RelativeFrame(env)
-    env = Quat2EulerWrapper(env)
+    env = Quat2EulerWrapper(env, keep_gripper=not no_gripper)
```
QuanluZhang: how about directly using `gripper_enabled` here?

docs/source-en/rst_source/examples/embodied/xsquare_turtle2.rst
```diff
@@ -118,6 +118,17 @@ Then install the RLinf Python dependencies for the embodied real-world setup:
...
    source .venv/bin/activate
 
+.. note::
+
+   On the Turtle2 controller node, we recommend a **dual-container split** to reduce ROS ownership conflicts:
+
+   - The **control container** (for example, the vendor-provided Turtle2 control container such as ``turtle2_release``) owns the ROS master, UI, low-level bring-up, and the camera / arm control nodes.
+   - A **separate RLinf container** only runs RLinf and the Ray worker, and should not directly own ``roscore``, the UI, or ``run.sh``.
+
+   The RLinf container should be built from an image configuration that is **compatible with the control container**, ideally sharing the same base image, system dependencies, ROS runtime, and required mounts, while only adding the RLinf Python environment and code on top. This reduces drift in ROS dependencies, message definitions, and device access.
```
QuanluZhang, line 128: As RLinf env is directly built on top of Turtle2 ROS env, why do we still need to use two containers? RLinf env can run "control container", right?

examples/embodiment/config/realworld_turtle2_deploy_eval.yaml
MuggleZzzH: Yes, we still need the actor placement in eval-only configs. In RLinf, `only_eval=True` disables training updates, but the actor/model component is still created and used as the policy/model side of the rollout pipeline. Existing eval configs such as `realworld_eval.yaml` and several `*_eval.yaml` examples also keep actor placement for the same reason.

docs/source-en/rst_source/examples/embodied/xsquare_turtle2.rst
MuggleZzzH: We keep RLinf in a separate container because the current RLinf real-world environment setup may kill local ROS core processes during initialization. In particular, RLinf may terminate local roscore / rosmaster / rosout processes as part of its real-world setup logic. If RLinf runs inside the same control container, this can accidentally kill the ROS master and disrupt the vendor bring-up, UI, camera nodes, and low-level arm-control nodes.

Therefore, the control container should own the long-lived ROS hardware stack, while the RLinf container should only act as a ROS client that connects to the existing ROS master, runs Ray/env/policy code, and publishes control commands. This split prevents RLinf-side initialization or restarts from killing the control container’s ROS processes.

rlinf/envs/realworld/common/wrappers/apply.py
MuggleZzzH: fixed

rlinf/envs/realworld/xsquare/tasks/__init__.py
```diff
@@ -41,7 +48,56 @@ def create_button_env(
...
+        )
+    action_mode = env_action_mode or override_action_mode or "relative_pose"
+    if action_mode not in {"relative_pose", "absolute_pose"}:
+        raise ValueError(
+            f"Unsupported Turtle2 deploy action_mode={action_mode!r}. "
+            "Expected one of {'relative_pose', 'absolute_pose'}."
+        )
+
+    override_cfg["action_mode"] = action_mode
+    override_cfg.setdefault("use_arm_ids", [0, 1])
```
andylin-hao: Should we allow configuring arm ids (e.g., single arm) when the Turtle2DeployEnv-v1 always returns apply_dual_arm_wrappers?

andylin-hao (approved):

rlinf/envs/realworld/common/wrappers/apply.py
QuanluZhang: at the top of this function, you use `env.get_wrapper_attr("config")` to get `config`, but here `config` is accessed through `env.config`. Is this correct?

rlinf/envs/realworld/xsquare/tasks/__init__.py
```diff
@@ -41,7 +48,56 @@ def create_button_env(
...
+    override_cfg.setdefault("use_camera_ids", [0, 1, 2])
+    override_cfg.setdefault("enforce_gripper_close", False)
+    override_cfg.setdefault("enable_task_reward", False)
+    override_cfg.setdefault("task_description", env_cfg.get("task_description", ""))
+    override_cfg["action_mode"] = action_mode
```
QuanluZhang: this line is the same as line 78

QuanluZhang: better to remove the first one

rlinf/envs/realworld/xsquare/tasks/__init__.py
MuggleZzzH: This PR narrows Turtle2DeployEnv-v1 to dual-arm deploy only, so I fail fast unless use_arm_ids is [0, 1]. Single-arm deploy is out of scope for this contract.

rlinf/envs/realworld/common/wrappers/apply.py
MuggleZzzH: Good catch. I moved the Turtle2-specific wrapper assembly out of the shared dual-arm builder and kept the shared path generic; the Turtle2 stack is now assembled directly in the factory, so common code no longer carries the Turtle2 contract.

rlinf/envs/realworld/xsquare/tasks/__init__.py
MuggleZzzH: Agreed, I removed the duplicate assignment.

PR #1146, opened 2026-05-12
feat(realworld): add LUMOS V4L2 camera backend

Brunch-Life:
Add a third camera backend alongside RealSense and ZED for LUMOS USB cameras.

FrankaRobot gains a "lumos" branch in camera enumeration and SDK validation; the only runtime dependency is opencv-python (cv2).

Mirroring the ZED smoke script, toolkits/realworld_check/test_lumos_camera.py provides a hardware-in-loop sanity check.

### Motivation and Context

Add Lumos camera support for RLinf.

### How has this been tested?

It can pass the realworld test. Also the test_lumos_camera.py can pass.

### Types of changes

### Checklist:


rlinf/envs/realworld/common/camera/lumos_camera.py
```diff
@@ -0,0 +1,131 @@
...
+        if os.path.exists(by_id):
+            return by_id
+        try:
+            return int(serial_number)
+        except ValueError as exc:
+            raise ValueError(
+                f"Could not resolve LUMOS serial_number={serial_number!r} to a V4L2 device."
+            ) from exc
+
+    def _read_frame(self) -> tuple[bool, Optional[np.ndarray]]:
```
QuanluZhang, line 121: where is this function called? Is this file copied from somewhere else?

Brunch-Life: It's a base function from BaseCamera class

Brunch-Life: see `rlinf/envs/realworld/common/camera/base_camera.py:86`

rlinf/envs/realworld/common/camera/lumos_camera.py
```diff
@@ -0,0 +1,131 @@
...
+        except Exception:
+            pass
+
+        if exposure is not None:
+            # 1 == manual mode for most V4L2 UVC drivers.
+            self._cap.set(cv2.CAP_PROP_AUTO_EXPOSURE, 1)
+            self._cap.set(cv2.CAP_PROP_EXPOSURE, exposure)
+
+    _NATIVE_W = 1280
+    _NATIVE_H = 1280
```
QuanluZhang: Moving _NATIVE_W = 1280 and _NATIVE_H = 1280 to just before __init__ follows the sibling convention.

QuanluZhang:
ZEDCamera imports and uses get_logger() for SDK warnings (_logger.warning(...)). LumosCamera has no logging at all. Given that the bare except Exception block is the only place errors are observed, at minimum this block should log with get_logger().warning(...) before returning (False, None). Per CLAUDE.md: use rlinf.utils.logging.get_logger(), not print or stdlib logging.

rlinf/envs/realworld/common/camera/lumos_camera.py
```diff
@@ -0,0 +1,131 @@
...
+            yuv = np.ascontiguousarray(raw).reshape(
+                self._native_h * 3 // 2, self._native_w
+            )
+            bgr = self._cv2.cvtColor(yuv, self._cv2.COLOR_YUV2BGR_I420)
+            if (self._native_w, self._native_h) != (self._out_w, self._out_h):
+                bgr = self._cv2.resize(
+                    bgr, (self._out_w, self._out_h), interpolation=self._cv2.INTER_AREA
+                )
+            return True, bgr
+        except Exception:
```
QuanluZhang: A single corrupt USB packet returns a buffer that is the wrong size; reshape raises ValueError; the exception is swallowed; _capture_frames breaks; the    
  background thread exits. All subsequent get_frame() calls block until the 5-second timeout and then raise queue.Empty. There is no diagnostic — the robot
  hangs silently mid-episode.                                                                                                                                
                  
  Compare the sibling cameras: ZEDCamera._read_frame and RealSenseCamera._read_frame do not wrap in a bare except Exception — errors propagate, the thread   
  crashes with a visible traceback, and the failure is diagnosable.
                                                                                                                                                             
  Suggested fix: Remove the except Exception block. If the reshape can transiently fail (e.g., partial read), catch only the specific expected exception     
  (ValueError), log it via get_logger(), and return (False, None) only for the cases that are genuinely unrecoverable (like a device disconnect, which should
   raise and let the thread exit with a diagnostic). For truly unexpected errors, let them propagate.

rlinf/envs/realworld/common/camera/lumos_camera.py
```diff
@@ -0,0 +1,131 @@
...
+        dev_path: Union[str, int] = self._resolve_device_path(camera_info.serial_number)
+
+        self._cap = cv2.VideoCapture(dev_path, cv2.CAP_V4L2)
+        if not self._cap.isOpened():
+            raise RuntimeError(
+                f"Failed to open LUMOS camera (serial={camera_info.serial_number}, "
+                f"dev_path={dev_path})."
+            )
+
+        self._cap.set(cv2.CAP_PROP_FOURCC, cv2.VideoWriter_fourcc(*"YU12"))
```
QuanluZhang: Please check whether the following comment is helpful or not.

  cv2.VideoCapture.set(CAP_PROP_FOURCC, ...) returns a success bool, but many V4L2 drivers silently ignore formats they don't support (falling back to YUYV,
  MJPEG, etc.) while reporting success. If the driver returns any format other than I420, the reshape(native_h * 3 // 2, native_w) in _read_frame will fail  
  with a ValueError — which is then silently swallowed per the above, terminating the thread.
                                                                                                                                                             
  The frame dimensions aren't verified either: if CAP_PROP_FRAME_WIDTH/HEIGHT aren't accepted (different firmware, different device), the native buffer size 
  assumption is wrong and the reshape fails the same way.
                                                                                                                                                             
  Suggested fix: After the set() block in __init__, verify:
  actual_fourcc = int(self._cap.get(cv2.CAP_PROP_FOURCC))
  if actual_fourcc != cv2.VideoWriter_fourcc(*"YU12"):   
      raise RuntimeError(                                                                                                                                    
          f"Camera {dev_path!r} does not support YU12; "
          f"actual fourcc={actual_fourcc:#010x}"                                                                                                             
      )                                         
  actual_w = int(self._cap.get(cv2.CAP_PROP_FRAME_WIDTH))                                                                                                    
  actual_h = int(self._cap.get(cv2.CAP_PROP_FRAME_HEIGHT))
  if (actual_w, actual_h) != (self._native_w, self._native_h):                                                                                               
      raise RuntimeError(                                     
          f"Camera returned resolution {actual_w}×{actual_h}, "                                                                                              
          f"expected {self._native_w}×{self._native_h}"        
      )

rlinf/envs/realworld/common/camera/lumos_camera.py
```diff
@@ -0,0 +1,131 @@
...
+    """Camera capture for LUMOS USB cameras (V4L2, I420 stream).
+
+    ``camera_info.serial_number`` may be:
+
+    * a ``/dev/v4l/by-id/`` filename (preferred — stable across reboots)
+    * a ``"videoN"`` shorthand resolved to ``/dev/videoN``
+    * a numeric string or int interpreted as a V4L2 device index
+    """
+
+    def __init__(self, camera_info: CameraInfo, exposure: Optional[int] = None):
```
QuanluZhang: I noticed that in your code `exposure` is always not filled. Then, why do you add this argument? And I also noticed that both `ZEDCamera` and `RealSenseCamera` do not have this argument.

andylin-hao (approved):

Brunch-Life:
all above fixed! thx!

jx-qiu (approved):
Looks good!

PR #659, opened 2026-02-02
feat: remove EnvManager and add RecordVideo wrapper

guozhen1997:
### Description

1. Remove EnvManager
2. Add RecordVideo wrapper for Env video recording

### How has this been tested?

Tested:
- Maniskill, LIBERO, RoboTwin, RoboCasa, Calvin, MetaWorld

E2E test:
maniskill_ppo_openvla
image

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (Document-only update)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #127, opened 2025-09-23
fix: Use timestamps to name videos to avoid video overwrite bug

zixiao-bios:
### Description
Use timestamps to name videos, otherwise when set env.enable_offload=True, all videos will be named as 0.mp4, cover the previous "0.mp4".

### Motivation and Context
https://github.com/RLinf/RLinf/issues/114

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1168, opened 2026-05-17
fix(maniskill): avoid stale offload video counter state

### LINKED ISSUE #114: [Bug]: Video log was overwritten when set env.enable_offload=True
### Affected component

When set env.enable_offload=True, all videos will be named as 0.mp4, cover the previous "0.mp4".

### Bug description

The reason is that if offload is set, every time call `start_simulator()`, a new Env instance is created here:
https://github.com/RLinf/RLinf/blob/main/rlinf/envs/env_manager.py#L343
so `env_cls.video_cnt` is always equal to 0.

### Environment

Python version:
PIP list:
RLinf version: 
Megatron version: 
Docker image version: 
nvidia-smi:

### Reproduction

Set env.enable_offload=True

- [x] Have you checked relevant issues, FAQs (https://rlinf-docs.readthedocs.io/en/latest/rst_source/faq.html), or asked the chatbot at the top right corner of the [documentation page](https://rlinf-docs.readthedocs.io).

kunni918:
### Description
This removes stale `video_cnt` state from `_ManiskillEnvCore` offload serialization/deserialization.

`video_cnt` is now owned by the generic `RecordVideo` wrapper, not by the ManiSkill core env. The ManiSkill offload core still tried to save and restore it, so capturing offload state could crash before the state buffer was created.

The regression tests load `maniskill_offload_env.py` with lightweight fakes so the core offload-state contract can be checked without requiring ManiSkill/Sapien.

### Motivation and Context
Found while investigating #114 and the closed unmerged PR #127. This PR is related context only and does not close #114, because the original issue was about video file overwrite behavior and this PR fixes an adjacent stale offload-state bug.

#### Minimal reproduction

I reproduced the failure on the current base commit `3a7c83e` by applying only the new regression test file from this PR to `upstream/main` and running the smallest test that exercises the stale state access:

```bash
python -m pytest tests/unit_tests/test_maniskill_offload_env.py::test_maniskill_offload_state_does_not_require_record_video_counter -q
```

Result on `upstream/main`: the test fails with:

```text
E       AttributeError: '_ManiskillEnvCore' object has no attribute 'video_cnt'
rlinf/envs/maniskill/maniskill_offload_env.py:89
```

The failing line is:

```python
"video_cnt": self.video_cnt,
```

This happens before a state buffer can be produced, so an offloaded ManiSkill core can fail during `get_state()` even when video recording itself is handled outside the core env.

The fakes in the test only replace ManiSkill/Sapien dependencies; the test still executes the real `_ManiskillEnvCore.get_state()` method from the production file.

#### Root-cause trace

The ownership mismatch comes from two earlier changes:

1. `a500e7f feat: remove EnvManager and add RecordVideo wrapper (#659)` moved video recording out of environment classes and into `rlinf/envs/wrappers/record_video.py`.
   - `ManiskillEnv.__init__` stopped defining `self.video_cnt`.
   - `ManiskillEnv` lost its local `render_images`, `add_new_frames()`, and `flush_video()` video state/methods.
   - `RecordVideo.__init__` now defines `self.video_cnt = 0`.
   - `RecordVideo.flush_video()` writes `{self.video_cnt}.mp4` and increments the wrapper-owned counter.
   - `EnvWorker` wraps envs with `RecordVideo(env, video_cfg)` when `video_cfg.save_video` is enabled.

2. `89f71e5 feat(env_offload): support ManiSkillEnv and OpenSoraEnv offload (#727)` later reintroduced `rlinf/envs/maniskill/maniskill_offload_env.py`, but its `_ManiskillEnvCore` state schema still included the old env-owned video counter:
   - `get_state()` serialized `"video_cnt": self.video_cnt`.
   - `load_state()` restored `self.video_cnt = state["video_cnt"]`.

On current `main`, `_ManiskillEnvCore` subclasses `ManiskillEnv`, and `ManiskillEnv` no longer owns `video_cnt`; the wrapper does. The offload core should therefore serialize simulator/core state only, not wrapper video state.

### How has this been tested?
Targeted local checks on this branch:

```bash
python -m pytest tests/unit_tests/test_maniskill_offload_env.py tests/unit_tests/test_overlap_env_bootstrap.py -q
```

Result: `5 passed`.

```bash
python -m ruff check rlinf/envs/maniskill/maniskill_offload_env.py tests/unit_tests/test_maniskill_offload_env.py
python -m ruff format --check rlinf/envs/maniskill/maniskill_offload_env.py tests/unit_tests/test_maniskill_offload_env.py
git diff --check HEAD~1..HEAD
```

Results: ruff check passed, format check passed, diff whitespace check passed.

I did not run a real ManiSkill/Sapien embodied e2e locally. The added regression tests intentionally avoid those optional heavy dependencies and cover the `_ManiskillEnvCore` offload state schema directly.

The tests cover three state-schema cases:

- `get_state()` no longer emits `video_cnt`.
- `load_state()` accepts a state buffer without `video_cnt`.
- `load_state()` ignores a legacy buffer that still contains an extra `video_cnt` field.

### Types of changes

### Checklist:

### DIFF
--- a/tests/unit_tests/test_maniskill_offload_env.py
+++ b/tests/unit_tests/test_maniskill_offload_env.py
@@ -1,0 +1,220 @@
+# Copyright 2025 The RLinf Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib.util
+import io
+import sys
+import types
+from pathlib import Path
+
+import torch
+
+
+class _FakeCudaTensor:
+    def torch(self):
+        return torch.tensor([1.0])
+
+
+class _FakePx:
+    def __init__(self):
+        self.cuda_articulation_link_data = _FakeCudaTensor()
+        self.cuda_articulation_qacc = _FakeCudaTensor()
+        self.cuda_articulation_qf = _FakeCudaTensor()
+        self.cuda_articulation_qpos = _FakeCudaTensor()
+        self.cuda_articulation_qvel = _FakeCudaTensor()
+        self.cuda_articulation_target_qpos = _FakeCudaTensor()
+        self.cuda_articulation_target_qvel = _FakeCudaTensor()
+        self.cuda_rigid_body_data = _FakeCudaTensor()
+        self.cuda_rigid_dynamic_data = _FakeCudaTensor()
+
+    def gpu_update_articulation_kinematics(self):
+        return None
+
+
+class _FakeScene:
+    def __init__(self):
+        self.px = _FakePx()
+        self.timestep = None
+
+    def get_timestep(self):
+        return 0.02
+
+    def set_timestep(self, timestep):
+        self.timestep = timestep
+
+    def _gpu_apply_all(self):
+        return None
+
+    def _gpu_fetch_all(self):
+        return None
+
+
+class _FakeBatchedRng:
+    rngs = ["rng"]
+
+
+class _FakeController:
+    def __init__(self):
+        self.loaded_state = None
+
+    def get_state(self):
+        return {"controller": torch.tensor([1.0])}
+
+    def set_state(self, state):
+        self.loaded_state = state
+
+
+class _FakeAgent:
+    def __init__(self):
+        self.controller = _FakeController()
+
+
+class _FakeEnv:
+    def __init__(self):
+        self.unwrapped = self
+        self.device = "cpu"
+        self.scene = _FakeScene()
+        self._main_rng = "main_rng"
+        self._batched_main_rng = _FakeBatchedRng()
+        self._main_seed = 123
+        self._episode_rng = "episode_rng"
+        self._batched_episode_rng = _FakeBatchedRng()
+        self._episode_seed = 456
+        self.action_space = "action_space"
+        self.single_action_space = "single_action_space"
+        self._orig_single_action_space = "orig_single_action_space"
+        self._elapsed_steps = torch.tensor([3])
+        self._init_raw_obs = {"obs": torch.tensor([1.0])}
+        self.agent = _FakeAgent()
+        self.task_reset_states = {}
+        self.task_metric_states = {}
+        self.reset_seed = None
+        self.reset_options = None
+        self.loaded_state = None
+
+    def get_state(self):
+        return {"sim": torch.tensor([2.0])}
+
+    def reset(self, seed=None, options=None):
+        self.reset_seed = seed
+        self.reset_options = options
+
+    def set_state(self, state):
+        self.loaded_state = state
+
+
+class _FakeManiskillEnv:
+    pass
+
+
+def _load_maniskill_offload_module(monkeypatch):
+    repo_root = Path(__file__).resolve().parents[2]
+    module_path = (
+        repo_root / "rlinf" / "envs" / "maniskill" / "maniskill_offload_env.py"
+    )
+
+    fake_package = types.ModuleType("rlinf.envs.maniskill")
+    fake_package.__path__ = [str(module_path.parent)]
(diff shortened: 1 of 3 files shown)


guozhen1997 (approved):

andylin-hao (approved):

tests/unit_tests/test_maniskill_offload_env.py
guozhen1997, line 1: Hi @kunni918, thanks for your contribution. Can you fix the unit test errors in this script?

kunni918: Thanks! Fixed in `c408b1d`.

Root cause: the new unit test imports `maniskill_offload_env.py`, which imports `recursive_to_device` from `rlinf.envs.utils`. `rlinf.envs.utils` imported optional `imageio` at module import time, but the base unit-test environment does not install the embodied/video extra. I moved `imageio` into `save_rollout_video()` as a lazy import and added `tests/unit_tests/test_env_utils_optional_dependencies.py` so pure env utils stay importable without `imageio`.

Verified:
- `uv run pytest tests/unit_tests/test_env_utils_optional_dependencies.py tests/unit_tests/test_maniskill_offload_env.py -q` -> `4 passed`
- `uv run ruff check rlinf/envs/utils.py tests/unit_tests/test_env_utils_optional_dependencies.py tests/unit_tests/test_maniskill_offload_env.py` -> passed
- `uv run ruff format --check rlinf/envs/utils.py tests/unit_tests/test_env_utils_optional_dependencies.py tests/unit_tests/test_maniskill_offload_env.py` -> passed

I also ran the full unit suite locally with CI-style `PYTHONPATH`; the remaining failure is unrelated to this PR: `tests/unit_tests/test_worker.py::TestWorkerGroup::test_execute_on_specific_ranks` depends on this machine GPU/rank environment, and the isolated rerun with `CUDA_VISIBLE_DEVICES=` passed.

andylin-hao: I don't think it's necessary to add these unit tests. Unit tests for env are not intended as they have complicated dependencies and we want to keep the unit test simple and manageable. We have end-to-end tests for this and it already covers maniskill offload. Could you remove these two test files and restore the `utils.py` as well?

kunni918: I understand the concern about keeping env unit tests out of the unit suite.

The reason I added a narrow regression test here is that this bug came from a gap between the e2e coverage and the offload state contract:

- `a500e7f` / #659 moved video recording state out of `ManiskillEnv` and into the `RecordVideo` wrapper (`video_cnt` is no longer core simulator/env state).
- `89f71e5` / #727 then introduced `ManiskillOffloadEnv`, but the serialized `_ManiskillEnvCore` state still included and restored `video_cnt` as if it were core env state.
- The existing ManiSkill offload e2e config does exercise `env.enable_offload: True`, but it does not assert the serialized offload-state schema. Also, `ManiskillOffloadEnv.offload()` catches `get_state` failures and falls back to an invalid state buffer, after which the auto-reset/onload path can still keep the e2e run moving. So the e2e path can pass while this specific state-contract regression remains hidden.

I can remove the two unit test files and restore `utils.py` if the project policy is to keep env behavior out of unit tests. My only concern is that we would again have no lightweight regression check for this state-schema bug. Would you prefer that I instead add this check to an existing e2e path, or should this PR remain code-only and rely on the existing e2e coverage?

andylin-hao: I see, that makes sense. But I don't think the test_env_utils_optional_dependency is necessary. We can remove that and keep the maniskill test. Thanks!

kunni918: Implemented in `60536bd`.

Changes now match that direction:
- removed `tests/unit_tests/test_env_utils_optional_dependencies.py`
- kept `tests/unit_tests/test_maniskill_offload_env.py` as the focused ManiSkill offload state regression test
- kept the `imageio` lazy import in `rlinf/envs/utils.py`, since the retained ManiSkill unit test imports `_ManiskillEnvCore` and therefore imports `recursive_to_device` through the base unit-test environment
- squashed the PR branch back to one signed commit: `fix(maniskill): avoid stale offload video counter state`

Verified after the squash:
- `uv run pytest tests/unit_tests/test_maniskill_offload_env.py -q` -> `3 passed`
- `uv run ruff check rlinf/envs/utils.py tests/unit_tests/test_maniskill_offload_env.py` -> passed
- `uv run ruff format --check rlinf/envs/utils.py tests/unit_tests/test_maniskill_offload_env.py` -> passed
- `git diff --check` -> passed

PR #1169, opened 2026-05-17
feat(scheduler): sync code by ray cluster

guozhen1997:
### Description

Support code sync

- `RLINF_CODE_WORKING_DIR` turns on Ray job-level code sync: it packages only the rlinf/ tree into runtime_env.py_modules so workers get the same package when they do not share the driver’s filesystem (`auto` infers the repo root, or set an absolute repo or rlinf path).

### Motivation and Context

In edge–cloud or heterogeneous training, the driver and workers often live on different filesystems, so teams end up checking out or syncing the same repo twice and keeping the two trees aligned by hand. Shipping the rlinf/ package with the Ray job from the driver avoids that duplicated workflow and reduces version skew.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/scheduler/cluster/cluster.py
+++ b/rlinf/scheduler/cluster/cluster.py
@@ -12,17 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import atexit
+import io
 import logging
 import os
 import re
 import shlex
+import shutil
 import signal
+import subprocess
 import sys
+import tarfile
 import tempfile
 import time
 from enum import Enum
 from importlib.metadata import version
-from typing import TYPE_CHECKING, Optional
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Optional
 
 import ray
 import ray.util.scheduling_strategies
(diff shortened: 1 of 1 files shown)


rlinf/scheduler/cluster/cluster.py
```diff
@@ -325,10 +678,10 @@ def _init_and_launch_managers(
             )
             time.sleep(1)
 
-        # Get node info
+        # Get node info (probe snapshot; once NodeManager is up, _nodes / _node_groups read from it).
```
FxxxxU: This comment is somewhat misleading.  _nodes / _node_groups is deleted.

FxxxxU (approved):

qurakchin (commented):
1. add a test
2. can you move the feat code to another util file?

rlinf/scheduler/cluster/cluster.py
guozhen1997: fixed

guozhen1997:
> 1. add a test
> 2. can you move the feat code to another util file?

1. It is hard to add a CI test or unit test. And I tested this feature in Shuangqing yesterday.
2. I have removed unnecessary changes. The current changes only apply to the cluster, and keeping them in cluster.py makes them more readable.

andylin-hao (approved):

PR #1192, opened 2026-05-21
fix(behavior): make dump/load TRO state round-trip symmetric

crabxiexy:
### Description

  Fixes two issues in `rlinf/envs/behavior/instance_generator.py::dump_tro_state` that caused `load_activity_instance_tro_state` to teleport objects and the robot when the dump/load pair runs mid-rollout in a vectorized eval loop:

  1. **A/ Frame mismatch for non-zero scene indices.** `BDDLEntity.dump_state` returns `root_link` poses in world frame, but the loader treats those poses as scene-relative whenever `env.scene.idx != 0` (cached instances are generated against scene 0, where world == scene). In a vectorized eval with N envs, every env with `scene.idx > 0` was getting an extra scene-offset baked in on load, silently teleporting every TRO object by the scene offset. Fixed by rebasing each `root_link` pose to scene frame via `Scene.convert_world_pose_to_scene_relative` on dump when `scene.idx != 0`, so dump/load is symmetric for any scene.

  2. **B/ Robot base pose was never the current pose.** `dump_tro_state` wrote only the task metadata's `robot_poses`, which holds the presampled reset-time pose. A mid-rollout round-trip therefore teleported the robot back to spawn even when every TRO object was restored correctly. Added an opt-in `capture_current_robot_pose: bool = False` parameter that overwrites the `robot_poses` entry with the robot's current scene-frame base pose. Default is `False` so the existing offline instance-generation path (`instance_generator.main`) is unchanged; callers performing mid-rollout snapshots opt in.

### Motivation and Context

To enable correct dump/load of state, making RL training starting from any timestep possible.

### How has this been tested?

  Ran the BEHAVIOR eval with an in-loop hook that, at chosen elementary env-step indices, snapshots world-frame poses for every TRO object, calls `dump_tro_state` per env, perturbs every object by +10 m in world x, calls `load_activity_instance_tro_state`, and reports per-object position/orientation drift against the pre-dump snapshot.

  Ablation reporting max position drift per env:

  | Case  | Fix A | Fix B | Max pos drift |
  |-------|-------|-------|--------------------------------------------|
  | A1_B1 |  on   |  on   | < 0.01 m (physics jitter). The video looks physically correct.                 |
  | A1_B0 |  on   |  off  | ~1.0 m (robot base teleports)          |
  | A0_B1 |  off  |  on   | The video becomes comepletely wrong, with the whole sky visible. |
  | A0_B0 |  off  |  off  |  The video becomes comepletely wrong, with the whole sky visible. |

  With both fixes, all 8 envs round-trip cleanly at every triggered step.

  Tested on a single-node 2×RTX4090 setup running
  `bash examples/embodiment/eval_embodiment.sh behavior_ppo_openpi_pi05`.

### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [X] My code follows the code style of this project.


qurakchin (approved):
LGTM and thanks for the contribution!

WinstonWmj (approved):

jx-qiu (approved):
Looks good.

guozhen1997 (approved):

andylin-hao (approved):

PR #1195, opened 2026-05-22
feat(embodiment): support overlap for rollout and training

anHappyDog:
### Description
 support overlap for rollout and training.

### Motivation and Context
Improve Performance.

### How has this been tested?
Test with libero spatial ppo openpi

### Types of changes

### Checklist:


examples/embodiment/train_embodied_agent.py
```diff
@@ -43,25 +43,45 @@ def main(cfg) -> None:
 
     # Create actor worker group
     actor_placement = component_placement.get_strategy("actor")
+    use_training_pipeline = bool(cfg.runner.get("use_training_pipeline", False))
```
thereAreDemonsNearby, line 46: Can this be deduced from component placement as we do in reasoning RL ?

rlinf/runners/embodied_runner.py
```diff
@@ -465,6 +478,198 @@ def run(self):
         self.log_queue.join()  # Wait for all queued logs to be processed
         self.log_thread.join(timeout=1.0)
 
+    def run_pipeline(self):
```
thereAreDemonsNearby, line 553: Is it possible to extract common parts for run() and run_pipeline() ? Too much duplicate code.

thereAreDemonsNearby, line 553: Is it possible to extract common parts for run() and run_pipeline() ? Too much duplicate code.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -1502,10 +1406,117 @@ def run_training(self) -> None:
...
+        append_to_dict(metrics, metrics_data)
+
+    def compute_entropy_loss(
+        self,
+        output_dict: dict,
+        loss: torch.Tensor,
+        loss_mask: torch.Tensor,
+        critic_warmup: bool,
+    ) -> float:
+        if self.cfg.algorithm.entropy_bonus > 0 and not critic_warmup:
```
thereAreDemonsNearby: Will there be a case that users want to monitor the entropy loss but don't want to train it (entropy_bonus==0)?  In reasoning scenario there is a configuration **cfg.algorithm.calculate_entropy** doing this.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -1502,10 +1406,117 @@ def run_training(self) -> None:
...
+            entropy = reshape_entropy(
+                output_dict["entropy"],
+                entropy_type=self.cfg.algorithm.entropy_type,
+                action_dim=self.cfg.actor.model.get("action_dim", 7),
+                batch_size=output_dict["logprobs"].shape[0],
+            )
+            entropy_loss = masked_mean(entropy, mask=loss_mask)
+            loss -= self.cfg.algorithm.entropy_bonus * entropy_loss
```
thereAreDemonsNearby: This method is named 'compute_entropy_loss' but it also applies entropy_loss to the final loss. Change the method name or move the **'loss -= ...'** logic out of this method.

rlinf/workers/actor/fsdp_actor_worker_pipeline.py
```diff
@@ -0,0 +1,183 @@
...
+        global_batches: dict[int, deque[GlobalBatchState]] = defaultdict(deque)
+        current_global_batch: list[dict[str, torch.Tensor]] = []
+        pending_global_batch: deque[dict[str, torch.Tensor]] = deque()
+        received_rollout_micro_batches: list[dict[str, torch.Tensor]] = []
+        received_micro_batch_count = 0
+
+        while True:
+            while received_micro_batch_count < self.micro_batches_per_step:
+                if not (micro_batch := self.try_recv_micro_batch(input_channel)):
+                    break
```
thereAreDemonsNearby, line 99: I would like to know whether the samples sent by the rollout engine to this worker are organized in the form of micro batches. I remember that in reasoning scenarios, each unit sent is a GRPO group, and sometimes there are performance issues due to too few samples.

rlinf/workers/env/env_worker.py
```diff
@@ -1284,3 +1305,70 @@ def get_actor_split_num(self):
...
+        flatten_batch = flatten_embodied_batch(batch, torch.arange(batch_size))
+        micro_batch_size = self.cfg.actor.micro_batch_size
+        assert batch_size % micro_batch_size == 0, (
+            f"Batch size {batch_size} is not divisible by micro_batch_size {micro_batch_size}."
+        )
+        num_micro_batches = batch_size // micro_batch_size
+        micro_batches = split_dict_to_chunk(flatten_batch, num_micro_batches, dim=0)
+        return [pack_batch(micro_batch) for micro_batch in micro_batches]
+
+    async def send_rollout_trajectories_pipeline(
```
thereAreDemonsNearby, line 1374: Please add some comments to explain and emphasize your refactoring. For example I notice that advantage computation is moved from actor to here.

rlinf/workers/env/env_worker.py
```diff
@@ -1284,3 +1305,70 @@ def get_actor_split_num(self):
...
+            "dones": rollout_batch["dones"],
+            "values": rollout_batch.get("prev_values", None),
+            "gamma": self.cfg.algorithm.get("gamma", 1),
+            "gae_lambda": self.cfg.algorithm.get("gae_lambda", 1),
+            "group_size": self.cfg.algorithm.get("group_size", 8),
+            "reward_type": self.cfg.algorithm.reward_type,
+            "loss_mask": rollout_batch.get("loss_mask", None),
+            "loss_mask_sum": rollout_batch.get("loss_mask_sum", None),
+            "normalize_advantages": self.cfg.algorithm.get(
+                "normalize_advantages", True
```
thereAreDemonsNearby, line 1332: I suspect whether the normalize_advantage mechanism can still work properly under pipeline execution mode. Because in colocated mode, normalize_advantage operates on the entire global batch, whereas in pipeline mode, its scope is limited to only a subset of the global batch.

examples/embodiment/train_embodied_agent.py
anHappyDog: Since this approach can be used even in shared/hybrid modes, it's actually quite difficult to deduce which mode(`run` or `run_pipeline`) to use from component placement. For some cases where env-rollout cannot fully utilize the GPU, using this approach in shared mode can also provide some performance improvement.

rlinf/runners/embodied_runner.py
anHappyDog: I'll try it.

rlinf/workers/env/env_worker.py
anHappyDog: This will indeed have an impact and requires further modification. However, in the previous baseline version, the `normalize_advantages` parameter was not passed when calculating advs, so no matter how `normalize_advantages` is  set in the YAML, the calculation of advs would always use normalize_advantages. This parameter has been added to kwargs for the pipeline version.Maybe should assert this to false if use this mode.

rlinf/workers/env/env_worker.py
anHappyDog: Okay.

examples/embodiment/train_embodied_agent.py
thereAreDemonsNearby: reasonable

rlinf/workers/actor/fsdp_actor_worker_pipeline.py
```diff
@@ -0,0 +1,183 @@
...
+        self.model.train()
+
+        metrics: dict[str, list[float]] = {}
+        global_batches: dict[int, deque[GlobalBatchState]] = defaultdict(deque)
+        current_global_batch: list[dict[str, torch.Tensor]] = []
+        pending_global_batch: deque[dict[str, torch.Tensor]] = deque()
+        received_rollout_micro_batches: list[dict[str, torch.Tensor]] = []
+        received_micro_batch_count = 0
+
+        while True:
```
Zengwen2001, line 96: The outer `while True` keeps calling `input_channel.get_nowait()` without any sleep / await / blocking primitive. If the actor has no micro‑batch available and no retraining work pending,  it will keep going for nothing and there is no time-out.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -993,6 +993,20 @@ def __init__(self, cfg: DictConfig):
...
+            % (self.cfg.actor.micro_batch_size * self._world_size)
+            == 0
+        ), "global_batch_size is not divisible by micro_batch_size * world_size"
+
+        self.gradient_accumulation = (
+            self.cfg.actor.global_batch_size
+            // self.cfg.actor.micro_batch_size
+            // self._world_size
+        )
+        self.update_epoch = self.cfg.algorithm.get("update_epoch", 1)
```
Zengwen2001, line 1003: duplicate `update_epoch` in line 1354

anHappyDog: done.

rlinf/workers/actor/fsdp_actor_worker_pipeline.py
anHappyDog: done.

rlinf/workers/actor/fsdp_actor_worker.py
anHappyDog: Maybe this can be added in another PR, I don't want to add another new feature into this PR.

rlinf/workers/actor/fsdp_actor_worker.py
anHappyDog: done.

rlinf/workers/actor/fsdp_actor_worker_pipeline.py
anHappyDog: For this embodied pipeline worker, each channel item is already an actor micro-batch. EnvWorker.prepare_micro_batches() flattens the rollout batch and splits it by cfg.actor.micro_batch_size, and
send_rollout_trajectories_pipeline() sends each packed micro-batch separately. This is different from the reasoning/GRPO path, where a sent unit may be a GRPO group.

rlinf/workers/env/env_worker.py
anHappyDog: config validation for this is added now. If `use_training_pipeline` is True, the param `normalize_advantages` must be False. I think the reason why success rate in this mode is a little lower than baseline might be this.

rlinf/workers/env/env_worker.py
anHappyDog: The refactor is intentional because the pipeline path now sends packed actor micro-batches instead of full rollout trajectories. The old episode-style payload included nested observations and could not take good advantage of the channel tensor fast path. We therefore preprocess the rollout batch, compute advantages/returns once before splitting, flatten it, and then send packed micro-batches.

Moving advantage computation here also matches the micro-batch send granularity: if we kept it on the actor side, the actor would need to reconstruct or buffer larger rollout batches before training, which would be less efficient and would defeat the purpose of streaming ready-to-train micro-batches.

Comments are added.

rlinf/runners/embodied_runner.py
anHappyDog: I think extracting common bookkeeping makes sense, but I would avoid fully merging the two loops. The duplicated tail can be shared, but the scheduling part is intentionally different: the normal path is staged, while the pipeline path overlaps rollout and actor training via the channel. Merging them would likely replace duplication with many conditionals around synchronization points.

anHappyDog: Done. I kept the scheduling parts separate because the normal path and pipeline path have different synchronization semantics, but extracted the shared post-step flow into common helpers for eval/checkpointing, metrics logging, and runner finalization.

rlinf/config.py
```diff
@@ -814,6 +814,12 @@ def validate_embodied_cfg(cfg):
         f"Supported embodied models: {sorted([x.value for x in EMBODIED_MODEL])}."
     )
 
+    if cfg.runner.get("use_training_pipeline", False):
+        assert not cfg.algorithm.get("normalize_advantages", True), (
+            "algorithm.normalize_advantages must be False when "
+            "runner.use_training_pipeline is True."
+        )
+
```
Zengwen2001, line 826: The new validate_embodied_cfg check only guards normalize_advantages, which is GAE‑specific. It doesn't cover group‑based advantages. 
The current e2e config (maniskill_ppo_openpi05_pipeline.yaml) uses adv_type: gae + group_size: 1, so CI never exercises this path.Please tighten the validation:                                                                                                                                                                          
                  
  if cfg.runner.get("use_training_pipeline", False):
      assert not cfg.algorithm.get("normalize_advantages", True), (                                                                                                                                       
          "algorithm.normalize_advantages must be False when "                                                                                                                                            
          "runner.use_training_pipeline is True."                                                                                                                                                         
      )                                                                                                                                                                                                   
      assert cfg.algorithm.adv_type == "gae", (
          "use_training_pipeline currently only supports adv_type=gae. "                                                                                                                                  
          "..."                                                                                                                                                                                   
      )

anHappyDog: done.

thereAreDemonsNearby (approved):

Zengwen2001 (approved):

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -1472,10 +1375,128 @@ def run_training(self) -> None:
...
+    ) -> float | None:
+        if self.cfg.algorithm.entropy_bonus > 0 and not critic_warmup:
+            entropy = reshape_entropy(
+                output_dict["entropy"],
+                entropy_type=self.cfg.algorithm.entropy_type,
+                action_dim=self.cfg.actor.model.get("action_dim", 7),
+                batch_size=output_dict["logprobs"].shape[0],
+            )
+            entropy_loss = masked_mean(entropy, mask=loss_mask)
+            return entropy_loss.detach().item()
```
andylin-hao: Why detach?

andylin-hao:
This PR raises a concern for me: are we going to sacrifice code readability for performance? Env worker now has stuff like compute_advantages_and_returns which almost completely duplicates actor worker. The runner is also substantialy changed to satisfy both modes. Would it be better to isolate everything like the async mode?

cc @guozhen1997 @QuanluZhang

andylin-hao (approved):

rlinf/workers/actor/fsdp_actor_worker.py
anHappyDog: fixed.

PR #1209, opened 2026-05-27
fix: remove global HF_LEROBOT_HOME override

guozhen1997:
### Types of changes

### Checklist:

### DIFF
--- a/rlinf/data/lerobot_paths.py
+++ b/rlinf/data/lerobot_paths.py
@@ -1,0 +1,66 @@
+# Copyright 2026 The RLinf Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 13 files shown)


rlinf/data/lerobot_paths.py
```diff
@@ -0,0 +1,66 @@
...
+                )
+            return str(path)
+        return str(first)
+    return str(data_paths)
+
+
+def resolve_lerobot_dataset_root(data_path: str) -> Path:
+    """Resolve the on-disk LeRobot dataset root for a path or Hugging Face repo id."""
+    path = Path(data_path).expanduser()
+    if (path / "meta" / "info.json").is_file():
```
F9rozen, line 59: raise filenotfounderror when file not found

guozhen1997: It is not necessary to raise an error here. If the corresponding file is not found, an error will occur during subsequent loading if the file is local; if the hf repository ID is used, LeRobotDataset will pull it from Hub on the first access.

andylin-hao (approved):

F9rozen (approved):

PR #1226, opened 2026-06-01
fix(logging): make experiment loggers optional

### LINKED ISSUE #1208: [Bug]: incompatible swanlab and openpi rich pins
### Affected component

Core dependency resolution: `rlinf` core deps (specifically `swanlab>=0.6.11`) are incompatible with `RLinf/openpi` (the openpi fork shipped with RLinf for VLA training). Any project that co-installs both — including any downstream consumer using uv/poetry/pip-tools — cannot resolve the dependency graph.

cc @andylin-hao who's historically owned install-dep issues here.

### Bug description

A clean uv resolution of `rlinf` + `openpi` together fails because the two packages have non-overlapping `rich` version constraints:

- `rlinf` core dependencies include `swanlab>=0.6.11` ([pyproject.toml#L65](https://github.com/RLinf/RLinf/blob/main/pyproject.toml#L65)) as an unconditional core dep.
- Every published swanlab version 0.4.5 through 0.7.19 (and the 0.8.0rc4 pre-release) hard-pins `rich>=13.6.0,<14.0.0`.
- `RLinf/openpi` HEAD requires `rich>=14.0.0` ([pyproject.toml#L38](https://github.com/RLinf/openpi/blob/main/pyproject.toml)).

There is no swanlab version anywhere on PyPI that satisfies openpi's rich floor, so the conflict cannot be resolved by version bumps on either side — only by changing one of the constraints in RLinf or openpi itself.

This surfaces for any downstream project that uses unified resolution (uv, poetry, pip-tools) to install both packages. Prebuilt Docker images and step-by-step `pip install` workflows can mask the bug because they freeze or sequentially apply resolutions rather than co-resolving.

### Configuration YAML file

N/A — this is a dependency resolution failure that occurs before any RL training config is loaded. Every example YAML in `examples/**` already defaults to `logger_backends: ["tensorboard"]` with `# wandb, swanlab` commented out, suggesting these are opt-in backends rather than core functionality.

### Log file

Resolver trace from `uv lock` against a minimal pyproject co-installing both packages (uv 0.11.14, CPython 3.11.15):

```
Using CPython 3.11.15 interpreter at: /nix/store/.../python3.11
   Updating https://github.com/RLinf/openpi.git (HEAD)
   Updating https://github.com/RLinf/RLinf.git (HEAD)
    Updated https://github.com/RLinf/RLinf.git (1061a894ce65c7ba58445ad3f96490e643928659)
    Updated https://github.com/RLinf/openpi.git (c5dc4b9296a1a4739bf52828f28a579f12dce763)
  × No solution found when resolving dependencies for split (markers:
  │ python_full_version >= '3.11' and sys_platform == 'darwin'):
  ╰─▶ Because only openpi==0.1.0 is available and openpi==0.1.0 depends on
      rich>=14.0.0, we can conclude that all versions of openpi depend on
      rich>=14.0.0. (1)

      Because swanlab>=0.6.11 depends on rich>=13.6.0,<14.0.0 and only the
      following versions of swanlab are available:
          swanlab<=0.6.11
          swanlab==0.6.12
          ...
          swanlab==0.7.19
      we can conclude that swanlab>=0.6.11 depends on rich>=13.6.0,<14.0.0.
      And because rlinf==0.2.0 depends on swanlab>=0.6.11, we can conclude
      that rlinf==0.2.0 depends on rich>=13.6.0,<14.0.0.
      And because we know from (1) that all versions of openpi depend
      on rich>=14.0.0, we can conclude that all versions of openpi and
      rlinf==0.2.0 are incompatible.
      And because only rlinf==0.2.0 is available, we can conclude that all
      versions of openpi and all versions of rlinf are incompatible.
      And because your project depends on openpi and rlinf, we can conclude
      that your project's requirements are unsatisfiable.

      hint: Pre-releases are available for `swanlab` in the requested
      range (e.g., 0.8.0rc4), but pre-releases weren't enabled (try:
      `--prerelease=allow`)
```

(I verified separately that swanlab 0.8.0rc4 also pins `rich<14.0.0`, so the pre-release hint does not lead to a resolution.)

### Environment

- python -V: Python 3.11.15
- uv pip list: N/A (resolution fails before any install)
- nvidia-smi: N/A (resolver bug, not runtime)
- RLinf version: `main` @ `1061a894ce65c7ba58445ad3f96490e643928659`
- Docker image tag: N/A (fresh resolution, no Docker)

### Other reproduction info

Minimal reproduction — no GPU, no install, runs in under 60 seconds with any uv:

```sh
mkdir test-rlinf-openpi && cd test-rlinf-openpi
cat > pyproject.toml <<'EOF'
[project]
name = "test"
version = "0.0.0"
requires-python = ">=3.10,<3.12"
dependencies = [
    "rlinf @ git+https://github.com/RLinf/RLinf.git",
    "openpi @ git+https://github.com/RLinf/openpi.git",
]
EOF
uv lock --python 3.11
```

Note: I have not verified whether `requirements/install.sh embodied --model openpi` reproduces the same failure — it uses `uv pip install` calls sequentially (install_openpi_model() runs separately at install.sh:1052) which may be more permissive than a unified resolve. Could be tested separately.

---

### Proposed fixes

Either of the following resolves the conflict on its own. Happy to send a PR for whichever you prefer.

**Option A — move `swanlab` and `wandb` to an optional `[loggers]` extra**

```toml
# pyproject.toml
dependencies = [
    # ... existing core deps minus the two below ...
    "tensorboard",     # stays in core (used as the default in every example YAML)
]

[project.optional-dependencies]
loggers = [
    "wandb<0.25.1",
    "swanlab>=0.6.11",
]
```

Why this option:

- Every example YAML in `examples/**` defaults to `logger_backends: ["tensorboard"]` with `# wandb, swanlab` commented out — i.e. these backends are already opt-in at the application level. The pyproject pin doesn't match that intent.
- Removes the transitive `rich<14` ceiling from the core install entirely.
- `MetricLogger` backend setup can raise `ImportError("pip install rlinf[loggers] to use this backend")` when an unavailable backend is selected — clearer error than today's resolver wall.
- No swanlab-side fix exists (every published version including the latest pre-release pins `rich<14`), so this is the only path that doesn't depend on coordinating an upstream release.

Estimated diff: ~15 lines in `pyproject.toml` + small change in the logger backend setup + a doc note in the logger tutorial.

**Option B — change openpi's `rich` floor:**

If the rich>=14 requirement in openpi can be relaxed (perhaps the bump was for a single feature that has a compatibility shim), the conflict goes away without touching RLinf core. Doesn't help if openpi genuinely needs rich 14 features.

### Follow-up: ship a `uv.lock`

The more resilient follow up fix is to commit a lockfile for the default config (nvidia + x86_64 + Python 3.11 + embodied + openpi). The per-platform branches in `install.sh` (rocm, ascend, etc.) stay authoritative; the lockfile only covers the default path. Prevents future surprise bumps in the four git-pinned upstreams (rlinf, openpi, libero, mani_skill) from breaking fresh installs.

Addresses the recurring class of resolver-failure tickets (#699, #367, #544, #429). Happy to PR this as a follow-up after Option A or B lands, or even as a second commit in the same PR.

---

- [x] Have you checked relevant issues, FAQs (https://rlinf.readthedocs.io/en/latest/rst_source/faq.html), or asked the chatbot at the top right corner of the [documentation page](https://rlinf.readthedocs.io).

Nas01010101:
### Description
Moves W&B and SwanLab out of RLinf's core install into optional extras (`rlinf[wandb]`, `rlinf[swanlab]`, `rlinf[loggers]`), keeping TensorBoard as the default logger. Adds a clear `ImportError` (with a `pip install 'rlinf[...]'` hint) when an optional backend is selected but not installed, plus EN/ZH logger docs and unit coverage.

### Motivation and Context
Fixes #1208. `swanlab` pins `rich<14`, which conflicts with the RLinf/openpi fork's `rich>=14` requirement during unified dependency resolution. Since W&B and SwanLab are opt-in logger backends, they shouldn't constrain the core install. TensorBoard remains the default and is unaffected.

### How has this been tested?
- `pre-commit` (ruff lint + ruff-format, pinned v0.14.3) — clean
- `python -m py_compile` on the changed modules
- `python -c "import tomllib; tomllib.load(open('pyproject.toml','rb'))"` — extras parse correctly
- `pytest tests/unit_tests/test_metric_logger_optional_backends.py` — 2 passed (asserts the hinted ImportError when a selected backend is missing)

No training-performance impact — this only gates optional logging backends.

### Checklist
- [x] Conventional Commit title + DCO `Signed-off-by`
- [x] Tests added for the new behavior
- [x] Docs updated (EN + ZH)
- [x] Lint/format pass (pre-commit, ruff v0.14.3)


andylin-hao:
Hi @Nas01010101, thanks for raising the issue. But right now we still wish to keep swanlab in the core dependency as a lot of people are using it. Making it optional makes installation extremely complicated. Our install script doesn't have the conflict issue as it installs rlinf and openpi seperately. If you want to make both rlinf and openpi dependencies of your project (which is the cause of the conflict), I would suggest writing a dependency override in your `pyproject.toml` like this:

```
[tool.uv]
override-dependencies = [
    "rich<14", # or >=14, may also works
]
```

This should bypass the problem. Thanks!

Nas01010101:
Thanks @andylin-hao, that makes sense. I resolved it on my side with `[tool.uv] override-dependencies = ["rich>=14"]` in my project's pyproject.toml, and confirmed rlinf+swanlab and openpi both resolve and import cleanly under rich 15 (the `rich<14` pin looks over-conservative, since swanlab only uses the stable `Console`/`Text`/`Progress`/`Status` APIs that didn't change across rich 13→15, so forcing it up is safer than forcing openpi down to `<14`).

Another option for anyone who doesn't need both importable in the same process is installing rlinf and openpi into separate venvs, which sidesteps the conflict entirely. Appreciate the help!

PR #1267, opened 2026-06-11
feat(sglang): add support for sglang server

Lin-xs:
### Description
Adds a server-mode SGLang backend to RLinf so rollouts can talk to one or more SGLang HTTP engines through an `sglang-router` instead of using the in-process SGLang engine.

New module `rlinf/hybrid_engines/sglang/server/`:
- `SGLangServerWorker` — Ray worker that spawns an SGLang HTTP server per process group (sized by `tensor_parallel_size * pipeline_parallel_size`), waits on `/health`, and reports its URL.
- `SGLangRouterWorker` — single Ray worker that runs an `sglang-router` subprocess, dynamically registers/unregisters server URLs, and exposes router URL + health + a thin `generate` passthrough.
- `InferenceHTTPClient` — sync/async client over `requests` / `aiohttp` for `/generate`, `/v1/chat/completions`, and `/health` against either a router or a single server.
- `launch_sglang_router_and_server(config, cluster, rollout_hardware_ranks, ...)` — one-call orchestrator that builds a `PackedPlacementStrategy` from the rollout hardware ranks, launches the server group, brings up the router, and registers each server with the router from the driver.

Demo example wired up:
- `examples/reasoning/sglang_server_demo.py` — exercises sync/async `/generate` and `/v1/chat/completions` against the router for Qwen2.5-VL-3B.
- `examples/reasoning/config/sglang_server_demo.yaml` — `rollout.server` / `rollout.router` config block, with `launch_server` / `launch_router` toggles and `group_name` / `router_group_name`.
- `examples/reasoning/run_sglang_server_demo.sh` — launch script.

Dependency: `sglang-router` is added to the `agentic-sglang` optional dependency group in `pyproject.toml`.

### Motivation and Context
The existing SGLang integration only supports the in-process engine, which couples rollout workers to the engine lifecycle and makes it hard to (1) share a pool of SGLang engines across multiple consumers, (2) scale the engine pool independently of the trainer, and (3) speak the standard OpenAI / SGLang HTTP protocol from agentic / tool-using code.

This change introduces a "server mode" path: each engine runs as a long-lived HTTP server, an `sglang-router` fronts the pool with cache-aware routing, and callers (RL rollouts, eval harnesses, agentic demos) talk to a single stable router URL. It is fully opt-in — existing configs are unaffected.

### How has this been tested?
- Ran `bash examples/reasoning/run_sglang_server_demo.sh` on a 2-GPU node with `Qwen2.5-VL-3B` (TP=2, PP=1):
  - Server group came up, router registered the server, and `router_group.get_router_url()` returned a reachable URL.
  - Sync + async `/generate` and `/v1/chat/completions` paths all completed end-to-end.
  - Router + server groups shut down cleanly via `shutdown().wait()`.
- No existing rollout/training paths were modified, so existing reasoning / embodied configs continue to work unchanged.

- Router placement defaults to node rank 0 (head); override via `router_node_rank` in `launch_sglang_router_and_server`.
- `aiohttp` defaults to `limit=100` connections — for large fan-outs bump the client's `max_connections` and `ulimit -n` accordingly (documented in the `InferenceHTTPClient` docstring).
- Server registration is serialized from the driver to keep worker ordering stable; can be parallelized from N workers if needed.

### Types of changes

### Checklist:


rlinf/workers/rollout/sglang_server/router_worker.py
```diff
@@ -0,0 +1,388 @@
...
+                args.append(flag)
+        elif isinstance(value, (list, tuple)):
+            args.append(flag)
+            args.extend(str(v) for v in value)
+        else:
+            args.extend([flag, str(value)])
+    return args
+
+
+class SGLangRouterWorker(Worker):
```
andylin-hao, line 94: Should we move this to `workers` and name the file xxx worker?

rlinf/workers/rollout/sglang_server/launcher.py
```diff
@@ -0,0 +1,139 @@
...
+        router_group = SGLangRouterWorker.create_group(
+            config=config,
+            router_cfg=rollout_cfg.router,
+        ).launch(
+            cluster=cluster,
+            name=rollout_cfg.router_group_name,
+            placement_strategy=router_placement,
+        )
+
+    router_handle = router_group.init_router() if router_group is not None else None
```
bianhua-12, line 158: Could we make the startup path transactional and clean up partially-started resources on failure?

There are two related failure paths here:

1. At the orchestrator level, if `server_handle.wait()`, `router_handle.wait()`, `get_server_url()`, or `register_server()` raises, the already-launched server/router Ray groups are returned by no one and their child processes can keep running.

2. At the router worker level, `SGLangRouterWorker.init_router()` calls `subprocess.Popen()` and then waits for `/health`. If the process stays alive but never becomes healthy, `_wait_for_router_health()` raises while the router subprocess may still be holding the selected port.

For SGLang this is especially risky because failed startup can leave GPU-serving processes and router ports behind, causing OOM or port conflicts on the next run.

Could we wrap the launch/init/register sequence in `try/except` and best-effort shut down router then server before re-raising, and also make `init_router()` call `self.shutdown()`/reset state when its health wait fails?

rlinf/workers/rollout/sglang_server/launcher.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+        server_group = SGLangServerWorker.create_group(
+            config=config,
+            sglang_cfg=rollout_cfg.server,
+        ).launch(
+            cluster=cluster,
+            name=rollout_cfg.group_name,
+            placement_strategy=rollout_placement,
+        )
```
bianhua-12: Could we avoid rebuilding placement from only the flat hardware-rank list here?

The caller likely already has a parsed RLinf placement strategy from `ComponentPlacement`. Reconstructing a new `PackedPlacementStrategy` from `rollout_hardware_ranks` loses the original placement semantics, especially `node_group` / heterogeneous cluster placement / flexible mappings. It also assumes the ranks are contiguous.

For example, if the original component placement targets a non-default node group, this helper currently creates a new `PackedPlacementStrategy` without passing that node group, so the server group can be scheduled against the default group instead of the configured one.

Could this helper accept the caller-provided placement strategy directly, or preserve the original node group and process-to-resource mapping when repacking?

rlinf/workers/rollout/sglang_server/router_worker.py
Lin-xs: Done

rlinf/workers/rollout/sglang_server/launcher.py
Lin-xs: done

andylin-hao (approved):

guozhen1997 (approved):

PR #1279, opened 2026-06-15
fix(wrappers): ensure video flush for nested RecordVideo

wzii:
… videos and episodes can be saved together

A short change in `env_worker.py` to enable save both parquets and videos.

### Description
Add a function to walk through the env and find the inner RecordVideo

### Motivation and Context
Previously, if parquets are saved, videos won't be saved even if enabled in config file. That's a result of multiple wrappers of `env` (i.e. it becomes `CollectEpisode(RecordVideo(base_env))`, thus won't be detected by `isinstance(..., RecordVideo)`)

### How has this been tested?
Manually tested on LIBERO. Now it can save both parquet and video after the change. 

### Types of changes

### Checklist:


andylin-hao (approved):

rlinf/workers/env/env_worker.py
```diff
@@ -655,21 +655,36 @@ def compute_bootstrap_rewards(
...
+
     def finish_rollout(self, mode="train"):
         # reset
         if mode == "train":
             for i in range(self.stage_num):
-                if self.cfg.env.train.video_cfg.save_video and isinstance(
-                    self.env_list[i], RecordVideo
-                ):
-                    self.env_list[i].flush_video()
+                if self.cfg.env.train.video_cfg.save_video:
```
guozhen1997, line 663: Because gym wrappers delegate attribute access, replacing the isinstance check with getattr makes the call work through the wrapper chain. You can remove `_find_record_video` and change  the code like this:
```
if self.cfg.env.train.video_cfg.save_video:
    flush_video = getattr(self.env_list[i], "flush_video", None)
    if callable(flush_video):
        flush_video()
```

wzii: Thank you for your advice. However, I believe that
1. `CollectEpisode` wrapper is using `gymnasium` instead of `gym` (https://github.com/RLinf/RLinf/blob/e38cadbb8852cff7d31d6050c122a2dec5f785b1/docs/source-zh/rst_source/guides/data_collection.rst#L25), 
2. The later version (>=1.0.0) of `gymnasium` does not support `getattr` (instead, it uses `get_wrapper_attr`, see https://gymnasium.farama.org/_modules/gymnasium/core/#Wrapper.get_wrapper_attr).

Although the framework currently pins a gymnasium version, in practice users plug in custom environments/wrappers that may bring a different gymnasium (or classic gym) version, where `__getattr__` delegation differs (present in gym 0.26 / gymnasium 0.29, removed in gymnasium ≥1.0). Depending on that behavior would make video flushing version fragile.

guozhen1997: @wzii Thanks for the feedback — you're right, relying on plain getattr is fragile.

I'd suggest introducing a small helper (in rlinf.env.utils.py) to handle both APIs:
```
def get_env_attr(env, name, default=None):
    if hasattr(env, "get_wrapper_attr"):
        try:
            return env.get_wrapper_attr(name)
        except AttributeError:
            return default
    return getattr(env, name, default)
```
Then in finish_rollout:
```
if self.cfg.env.train.video_cfg.save_video:
    flush_video = get_env_attr(self.env_list[i], "flush_video")
    if callable(flush_video):
        flush_video()
```
This traverses the wrapper stack (including an outer CollectEpisode) while remaining compatible with gymnasium ≥ 1.0 and older/custom envs. The eval path can use the same helper and drop the isinstance(..., RecordVideo) check.

guozhen1997: Hi @wzii, let me know if you have any questions or opinions.

wzii: > Hi @wzii, let me know if you have any questions or opinions.

Hi @guozhen1997, sorry for the late response. Comment addressed.

zoeyuchao (approved):

guozhen1997 (approved):

PR #1282, opened 2026-06-16
feat: support behavior dream0 environment installation

anHappyDog:
### Description

support behavior dream0 environment installation

### Types of changes

### Checklist:

### DIFF
--- a/requirements/install.sh
+++ b/requirements/install.sh
@@ -968,9 +968,15 @@ EOF
     local cu_tag="cu${cuda_major}"            # e.g. cu12
     local torch_tag="torch${torch_mm}"        # e.g. torch2.6
 
-    # We currently assume cxx11 abi FALSE and linux x86_64
+    # Match flash-attn wheel ABI to the currently installed torch build.
     local platform_tag="linux_x86_64"
-    local cxx_abi="cxx11abiFALSE"
+    local cxx_abi
+    cxx_abi=$(python - <<'EOF'
+import torch
+
+print("cxx11abiTRUE" if torch._C._GLIBCXX_USE_CXX11_ABI else "cxx11abiFALSE")
+EOF
+)
 
     uv pip uninstall flash-attn || true
     local prebuilt_ver base_url wheel_name
@@ -1123,7 +1129,9 @@ install_openvla_oft_model() {
             install_common_embodied_deps
             uv pip install git+${GITHUB_PREFIX}https://github.com/moojink/openvla-oft.git  --no-build-isolation
             install_behavior_env
+            pushd ~ >/dev/null
             install_flash_attn
+            popd >/dev/null
             ;;
         maniskill_libero|libero)
             create_and_sync_venv
@@ -1200,7 +1208,9 @@ install_openpi_model() {
             uv pip install git+${GITHUB_PREFIX}https://github.com/RLinf/openpi
             install_behavior_env
             uv pip install protobuf==6.33.0
+            pushd ~ >/dev/null
             install_flash_attn
+            popd >/dev/null
             ;;
         maniskill_libero|libero)
             create_and_sync_venv
@@ -1502,7 +1512,7 @@ install_dreamzero_deps() {
     fi
 
     uv pip install -r $SCRIPT_DIR/embodied/models/dreamzero.txt
-    pip install -e "$dreamzero_path" --no-deps --ignore-requires-python
+    python -m pip install -e "$dreamzero_path" --no-deps --ignore-requires-python
 }
 
 install_dreamzero_model() {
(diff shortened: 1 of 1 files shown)


requirements/install.sh
```diff
@@ -242,6 +242,22 @@ validate_python_version() {
     fi
 }
 
+configure_model_env_versions() {
+    if [ "$TARGET" = "embodied" ] && [ "$MODEL" = "dreamzero" ] && [ "$ENV_NAME" = "behavior" ]; then
+        PYTHON_VERSION="3.10"
+        if [ -z "$TORCH_VERSION" ]; then
+            TORCH_VERSION="2.5.1"
```
thereAreDemonsNearby: I think that default python version and torch version here is kind of low

thereAreDemonsNearby: In my opinion, we should provide a unified env for dream0 based on torch 2.10, because dreamzero training with torch<=2.9 may suffer from gpu memory leak. But this may require further work.

requirements/install.sh
```diff
@@ -1396,8 +1412,55 @@ install_abot_m0_model() {
     uv pip uninstall pynvml || true
 }
 
+install_dreamzero_behavior_deps() {
```
thereAreDemonsNearby: why need a special install_dreamzero_behavior_deps function? how about using requirements/embodied/models/dreamzero.txt ?

requirements/install.sh
thereAreDemonsNearby: I just learned that 'behavior' depends on omnigibson, which in turn restricts the torch version to torch 2.5 and Python 3.10, so this is reasonable. However, please add comments to explain this.

requirements/install.sh
thereAreDemonsNearby: I just learned that 'behavior' depends on omnigibson, which in turn restricts the torch version to torch 2.5 and Python 3.10, so this is reasonable. However, please add comments to explain this.

anHappyDog: ok.

requirements/install.sh
anHappyDog: ok.

requirements/install.sh
anHappyDog: comment added.

thereAreDemonsNearby (approved):

requirements/install.sh
```diff
@@ -242,6 +242,23 @@ validate_python_version() {
     fi
 }
 
+configure_model_env_versions() {
```
andylin-hao: Remove these. Specific env-model related configurations should be done in their installation methods rather than in this common aread.

requirements/install.sh
```diff
@@ -242,6 +242,23 @@ validate_python_version() {
...
+    # here behavior pins python to 3.10 and torch to 2.5
+    if [ "$TARGET" = "embodied" ] && [ "$MODEL" = "dreamzero" ] && [ "$ENV_NAME" = "behavior" ]; then
+        PYTHON_VERSION="3.10"
+        if [ -z "$TORCH_VERSION" ]; then
+            TORCH_VERSION="2.5.1"
+        fi
+    fi
+}
+
+validate_torch_version() {
```
andylin-hao: Remove

requirements/install.sh
```diff
@@ -1400,8 +1417,55 @@ install_abot_m0_model() {
     uv pip uninstall pynvml || true
 }
 
+install_dreamzero_behavior_deps() {
```
andylin-hao: What are general dreamzero deps, what are behavior-related deps? Please seperate them. For dreamzero deps, write them in the dreamzero install function. Don't mix it with env install. 

If an env has special requirements for this model, write the special handling in its switch-case block. Please refer to other behavior installation methods. 

Also, we already have a dreamzero deps txt in dreamzero.txt, don't write so many hard requirements here. Put them there. You need to pick those really necessary requirements rather than pin everything. Do we need to pin other deps not in the current dreamzero.txt?

requirements/install.sh
```diff
@@ -1995,7 +2059,9 @@ install_docs() {
 
 main() {
     parse_args "$@"
+    configure_model_env_versions
```
andylin-hao: Remove

requirements/install.sh
anHappyDog: done.

requirements/install.sh
anHappyDog: done.

requirements/install.sh
anHappyDog: done.

andylin-hao (approved):

guozhen1997 (approved):

PR #1321, opened 2026-06-30
fix: fix cn mirror asset downloads

yicwang:
### Description

Two fixes so that embodied Docker images can be built behind the
Firewall (i.e. with `--use-mirror` / without `NO_MIRROR=1`), where
`huggingface.co` and `github.com` are not directly reachable:

1. **`docker/Dockerfile`** — set `HF_ENDPOINT` image-wide (gated by
   `NO_MIRROR`, mirroring the existing `PIP_INDEX` / `UV_DEFAULT_INDEX`
   pattern). The OpenPI tokenizer is fetched via the `hf` CLI, which only
   honors the `HF_ENDPOINT` env var; previously it was never set at image
   scope, so `hf download` hit `huggingface.co` and failed.

2. **`requirements/embodied/download_assets.sh`** — `mani_skill.utils.download_asset`
   hardcodes `huggingface.co` / `github.com` URLs in `DATA_SOURCES` and
   downloads them with `urllib`, which ignores both `HF_ENDPOINT` and git's
   `insteadOf` rewrite. When `--use-mirror` is set, rewrite the in-memory
   `DATA_SOURCES` URLs to the configured mirrors (`HF_ENDPOINT` /
   `GITHUB_PREFIX`) before downloading. The non-mirror path is unchanged.

### Motivation and Context

Building any `embodied-*` target from mainland China currently fails at the
asset-download step. The existing mirror mechanism (apt/pip/git `insteadOf`)
does not cover tools that fetch hardcoded URLs via `urllib` (ManiSkill) or
that only read `HF_ENDPOINT` (the `hf` CLI). These two changes close those
gaps without affecting non-mirror builds.

### How has this been tested?

On a network-restricted host where `huggingface.co` / `github.com` are
unreachable but the CN mirrors are (an environment equivalent to the CN CI):

- `bridge_v2_real2sim` now downloads via `hf-mirror.com` (was the original
  failure point).
- `widowx250s` downloads via `ghfast.top`.
- The OpenPI tokenizer (`hf download RLinf/openpi_tokenizer`) resolves
  through `HF_ENDPOINT=hf-mirror.com`.
- Verified the non-mirror path (`NO_MIRROR=1`) still targets `huggingface.co`
  / `github.com` unchanged.

### Types of changes

### Checklist:


andylin-hao (approved):

andylin-hao:
Thanks @yicwang! LGTM

QuanluZhang (approved):

PR #1324, opened 2026-06-30
feat: add RLT algorithm

tiny-xie:
### Description

Adds support for the RLT algorithm, including the two-stage workflow for RL Token feature extraction and SAC-based policy training. This includes the related model/config integration, rollout transition pipeline, policy-switch wrapper, and RLT documentation.
### Motivation and Context

This enables RLinf to train lightweight online RL policies on top of VLA features using the RL Token approach. The initial supported example is the real-world Franka peg-insertion task.
### How has this been tested?

Tested with the real-world peg-insertion task configuration.
3c64450216fde3604af44f1a3a6be9f2

### Types of changes

### Checklist:


rlinf/workers/actor/rlt_ac_policy_worker.py
```diff
@@ -0,0 +1,269 @@
...
+        metrics["weighted_q"] = (q_weight * qf_pi.mean()).detach().item()
+        metrics["weighted_bc"] = (bc_weight * bc_loss).detach().item()
+        metrics["q_weight"] = q_weight
+        metrics["bc_weight"] = bc_weight
+        metrics["reference_dropout_prob"] = reference_dropout_prob
+
+        return actor_loss, entropy, metrics
+
+    @Worker.timer("forward_alpha")
+    def forward_alpha(self, batch):
```
thereAreDemonsNearby, line 265: Is 'forward_alpha' not implemented for now? Please use  NotImplementedError rather than failing silently

examples/embodiment/config/model/rlt_mlp_policy.yaml
```diff
@@ -0,0 +1,25 @@
...
+#   actor.model.num_action_chunks: ${env.train.num_action_chunks}
+#   actor.model.ref_num_action_chunks: ${env.train.ref_num_action_chunks}
+# For dual-arm actions, set action_dim to the full per-step action dimension,
+# for example 14. VLA may output ref_num_action_chunks actions; the RLT
+# actor consumes only the first num_action_chunks actions from that reference.
+action_dim: 7
+num_action_chunks: 50
+ref_num_action_chunks: 100
+
+precision: "32"
```
thereAreDemonsNearby: 32 is illegal. use fp32 instead. By the way, does this model need to be inferenced and trained in fp32 always? fp16 or bf16 will be much faster

examples/sft/config/rlt_sft_openpi_pi05.yaml
```diff
@@ -0,0 +1,138 @@
...
+  micro_batch_size: 16
+  global_batch_size: 64
+  seed: 42
+
+  openpi_data:
+    repo_id: ${rlt.openpi_repo_id}
+
+  model:
+    model_type: "openpi"
+    precision: null
```
thereAreDemonsNearby: **precision: null** is strange. We recommend that actor.model.precision is set to fp32, and actor.fsdp_config.mixed_precision is set to bf16 or fp16.

rlinf/models/embodiment/openpi/openpi_action_model.py
```diff
@@ -361,10 +392,205 @@ def sft_forward(self, data, use_action_chunk_loss: bool = False, **kwargs):
...
+            images, img_masks, lang_tokens, lang_masks
+        )
+        suffix_embs, suffix_pad_masks, suffix_att_masks, adarms_cond = (
+            self.embed_suffix(state, x_t, time)
+        )
+        if (
+            self.paligemma_with_expert.paligemma.language_model.layers[
+                0
+            ].self_attn.q_proj.weight.dtype
+            == torch.bfloat16
```
thereAreDemonsNearby: what to do with fp16 ?

rlinf/models/embodiment/openpi/openpi_action_model.py
```diff
@@ -628,17 +855,18 @@ def predict_action_batch(
         }
         return actions, result
 
-    @torch.no_grad()
-    def sample_actions(
+    def _sample_actions_with_prefix_cache(
```
thereAreDemonsNearby, line 893: Please move _sample_actions_with_prefix_cache after sample_actions so that git diff will looks better

rlinf/workers/actor/rlt_ac_policy_worker.py
tiny-xie: forward_alpha is for SAC entropy-temperature tuning. RLT does not use entropy training; its actor objective is -Q + BC, with alpha_type=fixed_alpha and initial_alpha=0.0. The method exists only because this RLT pipeline reuses RLinf's SAC infrastructure. I changed it to raise NotImplementedError so accidental automatic entropy tuning fails explicitly instead of silently returning zero.

examples/embodiment/config/model/rlt_mlp_policy.yaml
tiny-xie: Thanks for the reminder. I changed the config to use `fp32` instead of `32`.
For the current RLT actor-critic training, I keep fp32 by default because it gives better numerical precision/stability for Q-learning targets and actor-critic updates. Also, the MLP policy itself is relatively small and is not the main speed bottleneck in this pipeline.
I will run fp16 experiments this week. If fp16 works well and does not hurt training stability or performance, I will switch this config to fp16.

examples/sft/config/rlt_sft_openpi_pi05.yaml
tiny-xie: Thanks for your suggesion! I updated this config to set `actor.model.precision` to `fp32`, and set FSDP `mixed_precision` explicitly to `bf16`.

rlinf/models/embodiment/openpi/openpi_action_model.py
tiny-xie: Thanks for pointing it out. I updated this part to follow the actual dtype of the backbone q_proj weight instead of hard-coding bf16.

rlinf/models/embodiment/openpi/openpi_action_model.py
tiny-xie: Done.

examples/sft/config/rlt_sft_openpi_pi05.yaml
guozhen1997: Hi @tiny-xie, please keep `actor.model.precision: null` — OpenPI model already manages its own selective bf16/fp32 dtypes via `to_bfloat16_for_selected_params`, and setting FSDP mixed_precision to bf16/fp16 would conflict with that and can cause dtype mismatches under FSDP.

tiny-xie: Thanks for pointing it out. I've updated the YAML.

rlinf/data/embodied_io_struct.py
```diff
@@ -109,13 +109,15 @@ def prepare_observations(self, obs: dict[str, Any]) -> dict[str, Any]:
...
         )
+        rlt_use_actor = obs["rlt_use_actor"] if "rlt_use_actor" in obs else None
 
         return {
             "main_images": image_tensor,  # [N_ENV, H, W, C]
             "wrist_images": wrist_image_tensor,  # [N_ENV, H, W, C] or [N_ENV, N_IMG, H, W, C]
             "extra_view_images": extra_view_image_tensor,  # [N_ENV, N_IMG, H, W, C]
             "states": states,
             "task_descriptions": task_descriptions,
+            "rlt_use_actor": rlt_use_actor,
```
QuanluZhang: this is a rlt specific field, should we place it here?

tiny-xie: Well, `rlt_use_actor` is indeed RLT-specific, so putting it directly in the generic observation schema is not ideal.
The current implementation carries this flag in `obs` because this is the minimal path that already supports batching, routing, and decoupled rollout without adding another side channel. It does not affect other algorithms. If you prefer stricter separation, I can move this into a dedicated metadata path and keep the RLT-specific extraction/switch logic under another script.

rlinf/workers/actor/rlt_ac_policy_worker.py
```diff
@@ -0,0 +1,278 @@
...
+            "RLT AC disables entropy/alpha training. Use "
+            "algorithm.entropy_tuning.alpha_type=fixed_alpha."
+        )
+
+
+class RLTACFSDPPolicy(RLTACLossMixin, EmbodiedSACFSDPPolicy):
+    pass
+
+
+class AsyncRLTACFSDPPolicy(RLTACLossMixin, AsyncEmbodiedSACFSDPPolicy):
```
QuanluZhang, line 277: why should we add these classes?

rlinf/workers/env/env_worker.py
```diff
@@ -894,6 +894,53 @@ def store_last_obs_and_intervened_info(self, env_output_list: list[EnvOutput]):
             for env_output in env_output_list
         ]
 
+    @staticmethod
+    def _extract_rlt_obs_from_forward_inputs(
```
QuanluZhang: this is rlt specific function, should we place it here?

rlinf/workers/actor/rlt_ac_policy_worker.py
tiny-xie: RLT is not a SAC algorithm, but this implementation is built on top of the existing SAC worker to reuse the off-policy training pipeline. These two classes are thin adapters that combine the RLT-specific loss mixin with the sync/async SAC workers, so the launcher can instantiate the corresponding RLT worker. They do not add extra algorithmic behavior by themselves.

rlinf/workers/env/env_worker.py
tiny-xie: Thanks for pointing it out. I agree this logic is RLT-specific and should not live directly in the shared env worker. I will move the RLT feature extraction and transition-building logic into an algorithm-side helper, for example rlinf.algorithms.rlt.transition, and keep only a minimal dispatch in env_worker to call that helper when loss_type == "rlt_ac".

examples/embodiment/config/rlt_stage2_ac_mlp.yaml
```diff
@@ -0,0 +1,310 @@
...
+    override_cfg:
+      is_dummy: False
+      use_dense_reward: False
+      reward_scale: 1.0
+      max_num_steps: 300
+      target_ee_pose: TARGET_EE_POSE
+      success_hold_steps: 1
+      gripper_type: none
+
+    action_dim: ${rlt.action_dim}
```
guozhen1997: Please remove L170-172, since these configs are not used.

examples/embodiment/config/rlt_stage2_ac_mlp.yaml
```diff
@@ -0,0 +1,310 @@
...
+  target_update_type: all
+  train_actor_steps: 2
+  critic_actor_ratio: 4
+  critic_subsample_size: 0
+  backup_entropy: False
+
+# RLT Stage2 = freeze Stage1 pi0.5 + RLT token transformer, then train only the
+# RLT MLP actor-critic heads.
+rlt:
+  task_name: "realworld_peg_insertion"
```
guozhen1997: Please remove this config. It is only used for setting `logger.experiment_name` which is not necessary.

examples/embodiment/config/rlt_stage2_ac_mlp.yaml
```diff
@@ -0,0 +1,310 @@
...
+
+  # Required Stage1 checkpoint produced by examples/sft/config/rlt_sft_openpi_pi05.yaml.
+  # This is where the frozen OpenPI + RLT token transformer is loaded from.
+  # Do not put this path under actor.model or rollout.model.
+  stage1_model_path: "/path/to/stage1/checkpoint"
```
guozhen1997: There's no need to add these three configurations (L117-120); you can simply add comments where these three configs are used.

examples/embodiment/config/rlt_stage2_ac_mlp.yaml
```diff
@@ -0,0 +1,310 @@
...
+  # Must match the repo_id/config_name used to compute Stage1 OpenPI norm stats.
+  stage1_openpi_repo_id: "realworld_peg_insertion_rlt_stage1_7d"
+  openpi_config_name: "pi05_franka_state7d"
+
+  # Stage2 action shape. ref_num_action_chunks is the VLA reference horizon;
+  # num_action_chunks is the shorter chunk consumed/output by the Stage2 head.
+  action_dim: 7
```
guozhen1997: I suggest that all these parameters in the OpenPI model be removed from the `rlt` field. The `rlt` field should only retain `rlt-specific` parameters; there's no need to redefine these openpi model parameters here and then reference them elsewhere.

examples/sft/config/rlt_sft_openpi_pi05.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+  max_epochs: -1
+  max_steps: 20000
+  val_check_interval: -1
+  save_interval: 2000
+
+# The saved checkpoint from this run is consumed by Stage2 as
+# rollout.rlt_feature_model.model_path.
+rlt:
+  task_name: "realworld_peg_insertion"
```
guozhen1997: Suggest remove L34-40

examples/sft/config/rlt_sft_openpi_pi05.yaml
```diff
@@ -0,0 +1,138 @@
...
+  openpi_config_name: "pi05_franka_state7d"
+  # Used as OpenPI's norm-stats asset id. Keep this identical in Stage2.
+  openpi_repo_id: "realworld_peg_insertion_rlt_stage1_7d"
+  # Local LeRobot dataset root. It should directly contain data/ and meta/.
+  train_data_path: "/path/to/lerobot_dataset"
+  base_model_path: "/path/to/model"
+
+  # Realworld peg insertion uses single-arm Franka actions. VLA/reference may
+  # output a longer chunk than the Stage2 head consumes.
+  action_dim: 7
```
guozhen1997: Suggest remove these openpi model configs

rlinf/data/embodied_io_struct.py
guozhen1997: @tiny-xie I suggest removing `rlt_use_actor` from OBS and passing it out via `env info list`. Then, rename this parameter to `rlt_switch_flags` and add it to `EnvOutput`. This would correspond to `intervene_flags`, one representing the manual intervention flag and the other representing the switch from `ref` to `actor` flag.

Then you can add `rlt_switch_flags` to the `data` of the `send_to` interface (L1024 in env_worker.py) if `rlt` is enabled.

examples/sft/config/rlt_sft_openpi_pi05.yaml
guozhen1997, line 1: I suggest rename this file name to `rlt_stage1_sft_openpi_pi05.yaml`.

tiny-xie: No problem.

rlinf/workers/env/env_worker.py
```diff
@@ -1234,6 +1282,9 @@ def evaluate(self, input_channel: Channel, rollout_channel: Channel):
 
         return eval_metrics
 
+    def _use_rlt_stage2(self) -> bool:
```
guozhen1997: Could we make this an attribute (suggest naming `enable_rlt`) instead of a method? We can initialize it once in `env_worker.__init__`.

rlinf/workers/env/env_worker.py
```diff
@@ -1088,6 +1117,14 @@ async def _run_interact_once(
                     and reward_model_output is not None
                 ):
                     self.assign_history_reward(stage_id, reward_model_output)
+                if use_rlt_stage2 and self.collect_transitions:
+                    update_rlt_stage2_transitions(
```
guozhen1997: Suggest renaming this function to `update_rlt_transitions`.

tiny-xie: No problem. I've updated the code.

guozhen1997 (approved):

andylin-hao (approved):

PR #1325, opened 2026-07-01
perf(openpi): opt-in no-op of Observation @at.typecheck on rollout path

chenchaoxu7575:
### Description
Add an **opt-in** switch to no-op openpi's `@at.typecheck` on the pi0.5 rollout
inference path. openpi's central `Observation` dataclass is decorated with
`@at.typecheck` (jaxtyped + beartype runtime type/shape validation). On the
eager PyTorch rollout path this validation runs on **every** `Observation`
construction at full Python cost — ~2 ms/predict of pure CPU while the GPU is
idle. Gated behind `RLINF_DISABLE_OPENPI_TYPECHECK=1` (default **off**, so the
typecheck stays on unless explicitly disabled), applied before openpi's model
classes are decorated at import.

### Motivation and Context
Profiling pi0.5 rollout showed `predict_action_batch`'s `observation_from_dict`
taking ~3.5 ms CPU wall with the GPU idle. Isolated to the `@at.typecheck`
wrapper on `Observation.__init__`:

- micro-bench: `Observation(...)` construction **2.17 ms → 0.001 ms** when the
  wrapper is bypassed;
- the check is redundant per-step (rollout obs shapes are stable) and never
  changes output values.

Root cause: openpi is JAX-first — under `jax.jit` this check is traced away once,
but the PyTorch port pays it on every call. openpi's own `disable_typechecking()`
only gates the jaxtyping *shape* portion (~0.2 ms), not the dominant beartype
structural check, so this no-ops the decorator itself.

**Open question for maintainers.** Because `@at.typecheck` runs at import time
(class decoration), `cfg` isn't available yet, so this uses an env var rather
than config — which is admittedly not elegant. Guidance appreciated on the
preferred approach:
- (a) keep env-var opt-in (this PR);
- (b) no-op unconditionally + document the decision (RLinf never relies on
  openpi's per-construction typecheck);
- (c) upstream a real toggle into openpi so RLinf doesn't monkeypatch;
- (d) not worth it — close.

### How has this been tested?
Standalone harness, `max-autotune`, `num_steps=10`, `bs=1`, H20:
- `predict_action_batch`: **73.17 → 71.17 ms/call** (−2.0 ms, ~2.7%);
- action output **bit-exact** vs baseline (`max|diff| = 0.0`).
- `Observation` construction micro-bench: 2.17 ms → 0.001 ms (bypassed).
- `ruff check` passes on the changed file.

Not yet added: a CI unit test asserting bit-exact equivalence with the flag on/off
— happy to add one if the approach is accepted.

Magnitude is modest (~2.7% on the compiled path; larger on eager rollout where
denoise isn't fused). Low-risk cleanup, not urgent. Kept default-off to avoid any
behavior change.

### Types of changes

### Checklist:


thereAreDemonsNearby (commented):
We tested the overhead of this typecheck, and in most cases it's on the order of microseconds, with only a few cases reaching milliseconds. We believe the time saved is too small so it would likely be unnoticeable to users. So we probably don't plan to merge it.

chenchaoxu7575:
Could I ask what GPU and environment you tested on? I suspect we may be measuring slightly different things.

One likely explanation for the gap: openpi's own disable_typechecking() only gates the jaxtyping shape part, which in my micro-bench is just ~0.19 ms — so if that's what got toggled, "microseconds" matches what I see too. 
The piece this PR targets is the rest of the @at.typecheck wrapper (beartype on Observation.__init__), which the shape toggle leaves on. That's why no-op'ing the whole decorator recovers ~2 ms instead of ~0.2 ms.
For reference, on the pi0.5 eager rollout path (nsys CPU trace, max-autotune, bs=1): Observation construction ≈ 3.0–3.25 ms per predict with the GPU idle, and an isolated micro-bench is 2.17 ms with the wrapper. My guess is that openpi being JAX-first means jax.jit traces this away once, but the PyTorch port pays it on every predict.

If yours was the shape-only path or a different GPU, that would likely explain it.

andylin-hao (approved):

andylin-hao:
Hi @chenchaoxu7575, as we are moving to the RLinf's implementation of openpi (which is better aligned with the jax version). This enhancement is no longer necessary. Thanks for your contribution!

PR #1359, opened 2026-07-08
fix: rebuild lr scheduler when critic warmup finished

anHappyDog:
### Description
This PR fixes an FSDP checkpoint resume issue that can occur after critic warmup. During critic warmup, the optimizer is initially built for the value head only; once warmup finishes, the optimizer is rebuilt for the full trainable parameter set, but the LR scheduler was still bound to the old optimizer. This could produce inconsistent optimizer param group state, including missing `initial_lr` keys during DCP checkpoint load. The fix rebuilds the LR scheduler whenever the optimizer is rebuilt after critic warmup, and adds an explicit `last_epoch` argument to `build_lr_scheduler` with the default `-1`, so the post-warmup scheduler starts fresh for the new optimizer. Verification: `python -m py_compile rlinf/hybrid_engines/fsdp/fsdp_model_manager.py` and `ruff check rlinf/hybrid_engines/fsdp/fsdp_model_manager.py`.

### Motivation and Context
Fix resume.

### How has this been tested?
Test save and resume for pi and openvlaoft.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
+++ b/rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
@@ -441,13 +441,17 @@ class FSDPModelManager:
             if self.optimizer_steps >= self.critic_warmup_steps:
                 self.optimizer = self.build_optimizer(model=self.model)
                 self.critic_warmup_steps = 0
+                self.lr_scheduler = self.build_lr_scheduler(
+                    optimizer=self.optimizer,
+                    optim_config=self._cfg.optim,
+                )
         else:
             lr_list = [group["lr"] for group in self.optimizer.param_groups]
 
         return grad_norm, lr_list
 
     def build_lr_scheduler(
-        self, optimizer: Optimizer, optim_config: DictConfig
+        self, optimizer: Optimizer, optim_config: DictConfig, last_epoch: int = -1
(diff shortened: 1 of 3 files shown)


andylin-hao (approved):

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
```diff
@@ -441,13 +441,17 @@ def optimizer_step(self) -> tuple[float, list[float]]:
             if self.optimizer_steps >= self.critic_warmup_steps:
                 self.optimizer = self.build_optimizer(model=self.model)
                 self.critic_warmup_steps = 0
+                self.lr_scheduler = self.build_lr_scheduler(
+                    optimizer=self.optimizer,
+                    optim_config=self._cfg.optim,
```
qurakchin, line 446: should the last_epoch be added here?

anHappyDog: no need, here I think the warmup LR schedule does not need to be continued. If we implement it so that the LR schedule during warmup should continue, it becomes more complicated, requiring manual setting of the initial LR and handling of additional scheduler steps.

guozhen1997 (approved):

PR #1364, opened 2026-07-09
perf(openpi): batch all camera views into one SigLIP call in embed_prefix

chenchaoxu7575:
### Description
Override `PI0Pytorch.embed_prefix` in `OpenPi0ForRLActionPrediction` to embed all prefix camera views in a **single SigLIP ViT forward** (views stacked on the batch dim) instead of a per-view Python loop. Batch entries are independent inside the ViT, so the outputs match the loop path, while small batches fill the GPU much better: the vision forward sees `num_views x bsize` rows per call instead of `bsize`. The override is picked up by both the rollout inference path (`predict_action_batch` → `_build_prefix_cache`) and the training forward, which share `embed_prefix`.

### Motivation and Context
Real-robot pi0.5 rollout runs at bs=1 (one observation per control step), where the per-view loop badly underfills the GPU. Profiling the production inference path (`torch.compile max-autotune`, #968) on an RTX PRO 5000 (sm_120) with 3 camera views showed `embed_prefix`'s image embedding taking 9.5 ms of the ~65 ms e2e predict.

With this change, prefix image embedding drops **9.5 ms → 6.4 ms** per inference, contributing to an e2e `predict_action_batch` improvement of **65.8 ms → 58.9 ms** (bs=1, 10 denoise steps, 3 views, compiled) in our real-robot rollout measurements.

### How has this been tested?
Environment: H20 / RTX PRO 5000, torch 2.6/2.7, `RLinf/RLinf-Pi05-LIBERO-SFT` weights, `pi05` config.

1. **ViT-output equivalence** (real weights, bf16): batched vs loop embeddings are **bit-exact at bs=4**; at bs=1 the only differences are cuBLAS kernel-selection rounding (max diff ~1.2e-2 relative to the tensor max after the full 27-layer ViT) — the same class of numerical noise as enabling `torch.compile`.
2. **Action-level equivalence**: same-seed `predict_action_batch` (eval, 10 denoise steps) with loop vs batched embedding: max abs action diff **4.9e-3**, mean **9.0e-4** (unnormalized action range ≈ 0.3).
3. **Peak GPU memory**: the vision-embed transient grows ~×2.4 (bs=1: 10.7 → 20.3 MiB; bs=8: 69 → 163 MiB; bs=32: 272 → 648 MiB), but the **whole-predict peak is unchanged** at every batch size tested (bs=1: 14.11 GiB, bs=8: 22.07 GiB, bs=32: 32.43 GiB for both variants) — the global peak sits in the VLM prefill, and the caching allocator reuses the freed embed blocks. With `train_expert_only=True` (default RL configs) the VLM is frozen, so the training path behaves identically; with full-VLM training the *stored* activations are the same set in both variants (all views are needed for backward), only the checkpoint-recompute transient scales with `num_views`, which stays below the prefill activations.
4. **e2e smoke** on this branch: `predict_action_batch` with `pi05_libero` (2 views + language) produces sane actions; bs=1 latency benchmarked at 76.7 ms e2e on H20 (compiled, 3 views, unpatched container reference ~73 ms — structure and MFU conclusions identical across GPUs, absolute ms differ per part).

Measured with a standalone bs=1 inference benchmark plus nsys (NVTX range `prefix/vision_siglip` before/after). Happy to share the benchmark script, equivalence-check scripts, and nsys reps. Related: #968 (torch.compile predict path), #1325 (typecheck overhead on the same rollout path).

### Types of changes

### Checklist:

Note on tests: `pytest tests/unit_tests` in my container gives 201 passed / 3 skipped, with failures confined to the Ray/scheduler comm suites (`test_channel`, `test_comm`, `test_worker`, ...) that behave the same on a clean `upstream/main` checkout in this environment — no overlap with the changed file (the only openpi-related unit test, `test_value_model_embedding_scale`, is skipped by default). Please trigger CI for an authoritative run. The equivalence checks above need GPU + the released checkpoint, so they are not committed as CI unit tests; if maintainers prefer, I can add a GPU-marked e2e test mirroring check 1/2.

### DIFF
--- a/rlinf/models/embodiment/openpi/openpi_action_model.py
+++ b/rlinf/models/embodiment/openpi/openpi_action_model.py
@@ -1208,6 +1208,73 @@ class OpenPi0ForRLActionPrediction(PI0Pytorch, BasePolicy):
         v_t = self.action_out_proj(suffix_out)
         return v_t, suffix_out
 
+    def embed_prefix(
+        self, images, img_masks, lang_tokens, lang_masks
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """Embed prefix images and language tokens for PaliGemma processing.
+
+        Overrides ``PI0Pytorch.embed_prefix`` to embed all camera views in a
+        single SigLIP forward (views stacked on the batch dim) instead of a
+        per-view Python loop. Batch entries are independent inside the ViT, so
+        the outputs match the loop path, while small batches fill the GPU much
+        better (batch = num_views x bsize per call instead of bsize). At bs=1
+        with 3 views and torch.compile max-autotune, prefix image embedding
+        drops from 9.5 ms to 6.4 ms per inference on an RTX PRO 5000.
+
+        Args:
+            images: Sequence of per-view image tensors, each ``[B, C, H, W]``
+                with identical shapes across views.
+            img_masks: Sequence of per-view validity masks, each ``[B]``.
+            lang_tokens: Tokenized language prompt of shape ``[B, L]``.
+            lang_masks: Language padding mask of shape ``[B, L]``.
+
+        Returns:
+            Tuple of concatenated prefix embeddings ``[B, T, D]``, padding
+            masks ``[B, T]``, and attention masks ``[B, T]``.
+        """
+        assert len(images) == len(img_masks), (
+            f"Expected one mask per image view, got {len(images)} views "
+            f"and {len(img_masks)} masks."
+        )
+        embs = []
+        pad_masks = []
+        att_masks = []
+
+        bsize = images[0].shape[0]
+
+        def image_embed_func(img):
+            return self.paligemma_with_expert.embed_image(img)
+
+        all_img_embs = self._apply_checkpoint(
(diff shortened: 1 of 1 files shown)


chenchaoxu7575:
Withdrawing for now: we want to complete more validation before review (training-side transient memory at larger batch sizes, and whether the batched path should be inference-only). Will reopen once verified.

chenchaoxu7575:
Closing this for now. 
The latency win (bs=1 rollout) is validated, but we want more evidence before asking for review:    1. Numerics: the batched path is mathematically identical (batch entries are independent in the ViT) and bit-exact vs the loop at bs=4, but not bit-exact at bs=1 — changing the vision batch from B to num_views×B makes cuBLAS/attention pick different kernels, and the different reduction orders accumulate to ~1e-2 relative diff (bf16) through the 27-layer ViT.                                          
2. Training-side memory: the per-view loop is effectively micro-batching, so batching views triples the ViT transient working set (and the gradient-checkpoint recompute peak) — needs measurement at realistic SFT batch sizes (64–256), not just the bs≤32 inference numbers we have.

PR #1383, opened 2026-07-15
feat: support sglang embodied model evaluation

FxxxxU:
### Description

RLinf supports the use of Embedded models with SGLang as the rollout worker for inference. The model input consists of simulators (such as Libero and Maniskill), which are sent to the SGLang server process. Once inference is complete, the output is an action. 

Additionally, this PR not only provides a base Sglangembodiedworker but also supports launching the DreamZero model using Sglang (currently supported on [dreamzero supoort in sglang](https://github.com/sgl-project/sglang/pull/30679)) and using it as an rollout backend in RLinf.

### Motivation and Context

Advance the adaptation of RLinf to embodied models and expand to include more rollout backends, such as Sglang.

### How has this been tested?

Installation of the Dreamzero test environment: 
  1. Set up the environment (embodied + libero + dreamzero dependencies)
  ```
  cd path/to/RLinf && bash requirements/install.sh embodied --env libero --model dreamzero-sglang \
    --torch 2.11.0 --python 3.11.14 --venv path/to/dreamzero_test
  ```

  2. Install sglang (with the [diffusion] extra)
  ```
  source path/to/dreamzero_test/bin/activate
  # https://github.com/sgl-project/sglang/pull/30679/commits
  cd path/to/sglang_dreamzero && pip install -e "python[diffusion]"
  ```

  3. Run the evaluation
  ```
  cd path/to/RLinf && bash evaluations/run_eval.sh libero libero_spatial_dreamzero_eval_sglang
  ```

### Types of changes

### Checklist:


rlinf/workers/rollout/sglang/sglang_embodied_worker.py
```diff
@@ -188,11 +209,23 @@ def _init_sglang_server(self) -> None:
...
-                    "true" if getattr(sglang_cfg, "enable_cfg_parallel") else "false"]
+            cmd += [
+                "--enable-cfg-parallel",
+                "true" if getattr(sglang_cfg, "enable_cfg_parallel") else "false",
+            ]
         # --- performance preset ---
-        if (v := getattr(sglang_cfg, "performance_mode", None)):
+        if v := getattr(sglang_cfg, "performance_mode", None):
             cmd += ["--performance-mode", str(v)]
+        build_policy_serve_args = getattr(policy_cls, "build_sglang_serve_args", None)
```
FxxxxU: I'm confused about this part. Why, when variables like --enable-cfg-parallel have already been added earlier, is it necessary to use a separate function to configure something similar? I think a better approach would be to create a separate command here based on the model.

rlinf/workers/rollout/sglang/sglang_embodied_worker.py
```diff
@@ -188,11 +209,23 @@ def _init_sglang_server(self) -> None:
...
+        if callable(build_policy_serve_args):
+            cmd += build_policy_serve_args(
```
FxxxxU: Same as above. 
The design here could be improved; it would be best to make it simpler so that users can see the specific settings just by looking at the model.

rlinf/workers/rollout/sglang/action_policies/dreamzero.py
```diff
@@ -0,0 +1,837 @@
...
+class DreamZeroActionPolicy(EmbodiedActionPolicy):
+    """DreamZero eval policy for the SGLang embodied rollout worker.
+
+    ``SGLangEmbodiedWorker`` owns the server subprocess and channel loop.  This
+    policy owns only DreamZero-specific behavior: server launch arguments,
+    observation normalization, action HTTP requests and action unnormalization.
+    """
+
+    @staticmethod
+    def build_sglang_serve_args(
```
FxxxxU: This function should be modified.

rlinf/workers/rollout/sglang/__init__.py
```diff
@@ -30,14 +30,30 @@ def get_version(pkg):
...
+    Engine = None  # noqa: F841
+    try:
+        from rlinf.hybrid_engines.sglang.common import io_struct  # noqa: F811
+        from rlinf.hybrid_engines.sglang.common.sgl_engine import (
+            Engine,
+        )
+    except Exception as _e:  # pragma: no cover - depends on sglang version
+        import logging as _logging
+
+        _logging.getLogger(__name__).warning(
```
Lin-xs: Please use rlinf's logger here. Also, could you clarify what this warning means?
Regarding the sglang version check, I noticed you disabled the exception when the version isn't in (0.4.4, 0.5.4) instead of expanding the supported range. Could you explain the reasoning behind this?

FxxxxU: The main reason for lifting the SGLang version restriction is that, currently, SGLang services that need to support the VLA model must use the latest version of SGLang, which is far beyond the restricted version 0.5.4.

FxxxxU: fix.

rlinf/workers/rollout/sglang/action_policies/dreamzero.py
```diff
@@ -0,0 +1,756 @@
...
+
+def _urlopen_no_proxy(request_or_url, *, timeout: float):
+    return _NO_PROXY_HTTP_OPENER.open(request_or_url, timeout=timeout)
+
+
+@dataclass
+class DreamZeroActionRequest:
+    """Request contract used by the DreamZero SGLang action endpoint.
+
+    ``normalized_input`` is the output of the local DreamZero data transform and
```
Lin-xs: For each attribute, add corresponding comments directly beneath the field, following the same style as rlinf/scheduler/manager/worker_manager.WorkerInfo.

rlinf/models/embodiment/dreamzero/sglang_convert_action.py
```diff
@@ -0,0 +1,756 @@
...
+        from sglang.multimodal_gen.runtime.entrypoints.vla.protocol import (
+            pack_msgpack,
+        )
+
+        return (
+            pack_msgpack(self._to_msgpackable(payload)),
+            "application/msgpack",
+        )
+
+    def _post_action_request(self, body: bytes, content_type: str) -> dict[str, Any]:
```
Lin-xs: Could we use `requests` package here to simplify codes and align with the implementation of sglang_server? BTW, `unpack_msgpack` can be imported at the top level.

Such as:

```python
def _post_action_request(self, body: bytes, content_type: str) -> dict[str, Any]:
    retry_statuses = {500, 502, 503, 504}
    url = f"{self._server_url}/v1/actions/generations"
    last_error: Exception | None = None
    for attempt in range(self._max_retries + 1):
        try:
            resp = requests.post(
                url,
                data=body,
                headers={"Content-Type": content_type, "Accept": content_type},
                timeout=self._timeout_s,
                proxies={"http": None, "https": None},
            )
        except requests.exceptions.RequestException as exc:
            # Connection / timeout / DNS — retry while attempts remain.
            if attempt < self._max_retries:
                last_error = exc
                self._sleep_before_retry(attempt)
                continue
            raise RuntimeError(
                "DreamZero SGLang action HTTP request failed after retries: "
                f"attempts={self._max_retries + 1}, body_bytes={len(body)}, "
                f"last_error={last_error or exc}"
            ) from exc
        # Retryable 5xx — back off and try again.
        if resp.status_code in retry_statuses and attempt < self._max_retries:
            last_error = RuntimeError(
                f"status={resp.status_code}, body={resp.text[:1000]}"
            )
            self._sleep_before_retry(attempt)
            continue
        # Non-retryable (4xx / other 5xx / final attempt) — surface the body.
        if not resp.ok:
            raise RuntimeError(
                "DreamZero SGLang action HTTP request failed: "
                f"status={resp.status_code}, body_bytes={len(body)}, "
                f"body={resp.text}"
            )
        # Success — response must be msgpack.
        if "msgpack" not in resp.headers.get("content-type", "").lower():
            raise RuntimeError(
                "DreamZero SGLang action HTTP response must be msgpack, "
                f"got content-type={resp.headers.get('content-type')!r}"
            )
        from sglang.multimodal_gen.runtime.entrypoints.vla.protocol import (
            unpack_msgpack,
        )
        return unpack_msgpack(resp.content)
    raise RuntimeError(
        "DreamZero SGLang action HTTP request failed after retries: "
        f"attempts={self._max_retries + 1}, body_bytes={len(body)}, "
        f"last_error={last_error}"
    )
```

rlinf/models/embodiment/dreamzero/sglang_convert_action.py
```diff
@@ -0,0 +1,756 @@
...
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from typing import Any, Literal
+from urllib import error as urllib_error
+from urllib import request as urllib_request
+
+import numpy as np
+import torch
+from omegaconf import DictConfig, open_dict
+from tianshou.data import Batch
```
Lin-xs: Is this package installed in current environment?

rlinf/workers/rollout/sglang/action_policies/dreamzero.py
```diff
@@ -0,0 +1,756 @@
...
+        ``False``; if RLinf later needs to explicitly clear a server session, the
+        corresponding row can be set to ``True`` while model/cache lifecycle
+        resets remain server-owned.
+        """
+
+        batch_size = self._infer_batch_size(normalized_input)
+        session_ids = self._session_ids_for_batch(batch_size, policy_context)
+        reset_mask = [False] * batch_size
+        if self._debug_batch_print:
+            print(
```
Lin-xs: Use logger

rlinf/workers/rollout/sglang/__init__.py
```diff
@@ -30,14 +30,30 @@ def get_version(pkg):
...
+    try:
+        from rlinf.hybrid_engines.sglang.common import io_struct  # noqa: F811
+        from rlinf.hybrid_engines.sglang.common.sgl_engine import (
+            Engine,
+        )
+    except Exception as _e:  # pragma: no cover - depends on sglang version
+        from rlinf.utils.logging import get_logger
+
+        get_logger().warning(
+            "SRT hybrid engine unavailable on sglang %s (%s). The legacy "
```
Lin-xs: This warning message looks abnormal.
Replace it with: "Failed to apply path to sglang due to version mismatch; certain hybrid engine features will be disabled. Evaluation tasks remain unaffected."

rlinf/workers/rollout/sglang/__init__.py
```diff
@@ -30,14 +30,30 @@ def get_version(pkg):
...
 
 if package_version is None:
-    raise ValueError(f"sglang version {package_version} not supported")
-elif package_version >= parse("0.4.4") and package_version <= parse("0.5.4"):
-    sglang_version = package_version
-    from rlinf.hybrid_engines.sglang.common import io_struct
-    from rlinf.hybrid_engines.sglang.common.sgl_engine import (
-        Engine,
+    raise ValueError(
+        "sglang is not installed. Install sglang_main (e.g. "
```
Lin-xs: Just notify that sglang is not installed is enough.

rlinf/workers/rollout/sglang_server/server_worker.py
```diff
@@ -222,3 +222,119 @@ def shutdown(self) -> None:
...
+
+class SGLangmmgenServerWorker(SGLangServerWorker):
+    """Worker that owns one multimodal ``sglang serve`` subprocess.
+
+    Lifecycle surface mirrors :class:`SGLangServerWorker`
+    (``init_server`` / ``is_healthy`` / ``shutdown`` / ``get_server_url``);
+    only the launch mechanism differs (multimodal in-process launcher).
+    """
+
+    # ------------------------------------------------------------------
```
Lin-xs: Delete this.

rlinf/workers/rollout/sglang_server/launcher.py
```diff
@@ -130,9 +130,14 @@ def launch_sglang_router_and_server(
                 node_group=rollout_node_group,
             )
 
+        # One worker class serves both sglang model families; ``server_type``
```
Lin-xs: delete

rlinf/workers/rollout/sglang_server/server_worker.py
```diff
@@ -143,30 +213,26 @@ def init_server(self) -> None:
...
-        sglang_kwargs = OmegaConf.to_container(self._sglang_cfg, resolve=True)
-        sglang_kwargs["host"] = self._bind_host
-        sglang_kwargs["port"] = http_port
-        sglang_kwargs["dist_init_addr"] = f"127.0.0.1:{dist_port}"
-        server_args = ServerArgs(**sglang_kwargs)
+        server_kwargs = OmegaConf.to_container(self._sglang_cfg, resolve=True) or {}
+        server_kwargs["host"] = self._bind_host
+        server_kwargs["port"] = http_port
+
+        launch_router = bool(self._cfg.rollout.sglang.get("launch_router", False))
```
Lin-xs: Remove launch_router; this validation logic should be handled by its caller instead.

rlinf/workers/rollout/sglang_server/server_worker.py
```diff
@@ -143,30 +213,26 @@ def init_server(self) -> None:
...
 
+        tp_size = server_kwargs.get("tp_size") or server_kwargs.get("tp-size")
```
Lin-xs: The format of attributes' name should be checked at first.

rlinf/workers/rollout/sglang_server/server_worker.py
```diff
@@ -19,58 +19,116 @@
...
+    _ensure_no_proxy_for_localhost()
+    os.environ.setdefault("FLASHINFER_DISABLE_VERSION_CHECK", "1")
+
+    if server_type not in ("srt", "embodied"):
+        raise ValueError(
+            f"Unsupported server_type {server_type!r}; "
+            "expected 'srt' (language model) or 'embodied' (VLA/diffusion)."
+        )
+
+    if server_type == "embodied" and launch_router:
```
Lin-xs: Delete.

rlinf/workers/rollout/sglang_server/server_worker.py
```diff
@@ -175,8 +241,19 @@ def init_server(self) -> None:
         if self._advertise_host is None:
             self._advertise_host = ray.util.get_node_ip_address()
 
+        # multimodal_gen (VLA) warmup is heavier than an LLM, so it gets a
+        # longer default. Overridable via rollout.sglang.spawn_timeout.
+        default_spawn_timeout = 1800.0 if self._server_type == "embodied" else 300.0
+        spawn_timeout = float(
+            self._cfg.rollout.sglang.get("spawn_timeout", default_spawn_timeout)
```
Lin-xs: Don't use `_cfg.rollout.sglang` here because this field may not exist.

rlinf/workers/rollout/sglang_server/server_worker.py
```diff
@@ -88,12 +146,15 @@ class SGLangServerWorker(Worker):
     """Worker that owns one sglang HTTP server process.
 
     Args:
-        config: Full RLinf ``DictConfig``. Kept for parity with other
-            workers / future use; the sglang server itself is configured
-            entirely from ``sglang_cfg``.
+        config: Full RLinf ``DictConfig``. The sglang server itself is
```
Lin-xs: Delete.

rlinf/workers/rollout/sglang_server/server_worker.py
```diff
@@ -19,58 +19,116 @@
...
+            server_args = ServerArgs.from_kwargs(**server_args_kwargs)
+            set_global_server_args(server_args)
+            dispatch_launch(server_args)
+        else:  # "srt"
+            from sglang.launch_server import run_server
+            from sglang.srt.server_args import ServerArgs
 
+            server_args_kwargs["dist_init_addr"] = f"127.0.0.1:{dist_port}"
+            server_args = ServerArgs(**server_args_kwargs)
+            run_server(server_args)
```
Lin-xs: from sglang.srt.entrypoints.http_server import launch_server

There is no `sglang.launch_server.run_server`

Lin-xs (approved):

docs/source-zh/rst_source/evaluations/guides/dreamzero_sglang.rst
```diff
@@ -0,0 +1,125 @@
...
+
+安装 embodied、LIBERO 和 DreamZero SGLang 依赖：
+
+.. code-block:: bash
+
+   cd /path/to/RLinf
+   bash requirements/install.sh embodied --env libero --model dreamzero-sglang \
+     --torch 2.11.0 --python 3.11.14 --venv /path/to/dreamzero_test
+
+安装包含 DreamZero 支持的 SGLang，并启用 ``diffusion`` extra：
```
Lin-xs: Please provide specific version requirements. For example: sglang >= 0.5.x, with version 0.5.y verified through testing.

rlinf/models/embodiment/dreamzero/sglang_convert_action.py
Zengwen2001: It is used by DreamZero action policy, like in ./rlinf/models/embodiment/dreamzero/dreamzero_policy.py line 21

rlinf/models/embodiment/dreamzero/sglang_convert_action.py
Zengwen2001: done

rlinf/workers/rollout/sglang/action_policies/dreamzero.py
Zengwen2001: done

docs/source-zh/rst_source/evaluations/guides/dreamzero_sglang.rst
Zengwen2001: done

Zengwen2001 (approved):

evaluations/eval_embodied_agent.py
```diff
@@ -41,17 +41,52 @@ def main(cfg) -> None:
...
+        )
+
+        server_group, _ = launch_sglang_router_and_server(
+            cfg,
+            cluster,
+            rollout_hardware_ranks=component_placement.get_hardware_ranks("rollout"),
+            router_server_args=cfg.rollout.sglang,
+        )
+        _server_urls = list(server_group.get_server_url().wait())
+        print(f"[eval] launched {len(_server_urls)} sglang server(s): {_server_urls}")
```
guozhen1997: use rlinf logger

guozhen1997 (approved):

evaluations/eval_embodied_agent.py
```diff
@@ -41,29 +42,66 @@ def main(cfg) -> None:
...
+        # Create rollout worker group
+        rollout_placement = component_placement.get_strategy("rollout")
+        rollout_group = rollout_worker_cls.create_group(cfg).launch(
+            cluster, name=cfg.rollout.group_name, placement_strategy=rollout_placement
+        )
+    else:
+        raise ValueError(f"Unsupported rollout backend: {rollout_backend}")
     # Create env worker group
     env_placement = component_placement.get_strategy("env")
     env_group = env_worker_cls.create_group(cfg).launch(
```
andylin-hao, line 84: **Blocker — `env_worker_cls` is undefined on the sglang path.**

`env_worker_cls` is only assigned inside the `elif rollout_backend == "huggingface"` branch (lines 68 and 71), but it is used unconditionally here. With `rollout_backend: sglang` — the very path this PR adds — this raises `NameError: name 'env_worker_cls' is not defined` before any worker launches.

Suggested fix: hoist the default above the backend branch, since RTC only overrides it on the HF path:

```python
env_worker_cls = EnvWorker
rollout_backend = cfg.rollout.get("rollout_backend", "huggingface")
if rollout_backend == "sglang":
    ...
```

rlinf/workers/rollout/utils.py
```diff
@@ -335,20 +335,24 @@ def get_rollout_backend_worker(cfg: DictConfig) -> Worker:
...
+                SGLangEmbodiedWorker,
+            )
+
+            return SGLangEmbodiedWorker
+        elif serving_mode == "worker_http":
+            from rlinf.workers.rollout.sglang.sglang_worker_server import (
+                SGLangWorkerWithHTTPServer,
+            )
+
+            return SGLangWorkerWithHTTPServer
```
andylin-hao: **Blocker — this import was broken by the rename in this same PR.**

`rlinf/workers/rollout/sglang/sglang_worker_server.py` is renamed to `sglang_agent_worker.py` and `SGLangWorkerWithHTTPServer` to `SGLangAgentWorkerWithHTTPServer` in this PR (and `agentlightning_runner.py` was updated accordingly), but this branch still references the old names. Any config with `rollout.sglang.serving_mode: worker_http` now fails with `ModuleNotFoundError: rlinf.workers.rollout.sglang.sglang_worker_server`.

```python
from rlinf.workers.rollout.sglang.sglang_agent_worker import (
    SGLangAgentWorkerWithHTTPServer,
)

return SGLangAgentWorkerWithHTTPServer
```

rlinf/workers/rollout/sglang/__init__.py
```diff
@@ -31,13 +31,26 @@ def get_version(pkg):
...
+    Engine = None  # noqa: F841
+    try:
+        from rlinf.hybrid_engines.sglang.common import io_struct  # noqa: F811
+        from rlinf.hybrid_engines.sglang.common.sgl_engine import (
+            Engine,
+        )
+    except Exception as _e:  # pragma: no cover - depends on sglang version
+        from rlinf.utils.logging import get_logger
+
+        get_logger().warning("the sglang hybrid engine is unavailable.")
```
andylin-hao: **The version gate now fails silently, and this affects every sglang user, not just the new embodied path.**

The previous `else` branch raised a clear `ValueError` naming the unsupported version. Now `Engine` stays `None` and `io_struct` stays an empty `types.ModuleType` stub, both of which are re-exported via `__all__`. `rlinf/workers/rollout/sglang/sglang_worker.py:40` imports both, so an incompatible sglang version surfaces much later as `TypeError: 'NoneType' object is not callable` or `AttributeError: module 'io_struct' has no attribute 'TaskMethodInput'` in the middle of a rollout.

Two smaller issues on top: the swallowed exception `_e` is never used, so the warning drops the actual reason; and the message doesn't say which version was rejected.

Suggested fix: keep the hard failure (the `<= 0.5.16` bump above already widens the supported range), or gate the soft fallback to the embodied path only — and at minimum log `repr(_e)` and `package_version`.

rlinf/models/embodiment/dreamzero/sglang_adapter.py
```diff
@@ -0,0 +1,386 @@
...
+        ]
+        return {
+            "model": self._model,
+            "input": {
+                "prompt": self._as_texts(prompts, batch_size, "prompt"),
+                "observation": observation,
+            },
+            "parameters": {
+                "session_ids": session_ids,
+                "reset_mask": [False] * batch_size,
```
andylin-hao: **`reset_mask` is hardcoded to all-`False`, so server-side session state is never reset at episode boundaries.**

`session_ids` above are stable per `(rank, stage, slot)`, and the eval config runs `auto_reset: True` — so an env slot silently begins a new episode while the server keeps the same session. DreamZero is a world model with per-session cached context (`dreamzero_max_sessions: 128` in the YAML), so the first steps after a reset would condition on the previous episode's frames. That would quietly bias the reported `eval/success_once`.

The signal needed is already on the wire: `EnvWorker._build_rollout_input_data` (`rlinf/workers/env/env_worker.py:923`) sends `final_obs` alongside `obs`, and a non-`None` entry marks a slot that just reset. It's currently dropped by `SGLangEmbodiedWorker._merge_obs_batches` before `predict()` ever sees it.

Suggested fix: thread `final_obs` through `_merge_obs_batches` and `_split_context` into `reset_mask`. If reset genuinely isn't needed here, please add a comment saying so — an all-`False` mask next to stable session ids reads as an unfinished stub.

rlinf/models/embodiment/dreamzero/sglang_adapter.py
```diff
@@ -0,0 +1,386 @@
...
+    format_training_prompt,
+    load_dreamzero_dataset_metadata,
+    normalize_instruction_text,
+    rollout_obs_layout_for_embodiment,
+)
+
+_RLINF_POLICY_CONTEXT_KEYS = ("_rlinf_stage_id",)
+
+
+class DreamzeroSglangAdapter:
```
andylin-hao: Naming nit on the rename in `6517e61`: `DreamzeroSglangAdapter` breaks the casing used everywhere else in the codebase — `DreamZeroPolicy`, `DreamZeroConfig`, `SGLangEmbodiedWorker`, `SGLangAgentWorkerWithHTTPServer`. `DreamZeroSGLangAdapter` would match.

rlinf/workers/rollout/sglang/sglang_embodied_worker.py
```diff
@@ -0,0 +1,219 @@
...
+        if not obs_batches:
+            return {}
+        obs_dicts = [b["obs"] if "obs" in b else b for b in obs_batches]
+        merged: dict[str, Any] = {}
+        for key in obs_dicts[0].keys():
+            values = [d[key] for d in obs_dicts]
+            first = next((v for v in values if v is not None), None)
+            if first is None:
+                merged[key] = None
+            elif isinstance(first, torch.Tensor):
```
andylin-hao, line 151: This drops every key except `obs` from the env payload. `final_obs` (needed for `reset_mask`, see my comment on `sglang_adapter.py`) is silently discarded here.

Separately, several fields set in `__init__` are never read: `self._placement`, `self.total_num_eval_envs`, `self.only_eval`, `self.env_decoupled_mode`, `self.collect_prev_infos`. And `self._cfg`/`self.cfg` plus `self._cfg_rollout`/`self.cfg_rollout` are four names for two objects. Worth trimming to keep this close to the HF sibling.

If `env_decoupled_mode` is meant to be supported (the HF worker branches on it at `huggingface_worker.py:783`), either implement it or assert it is `False`.

rlinf/workers/rollout/sglang_server/server_worker.py
```diff
@@ -211,7 +252,11 @@ def init_server(self) -> None:
             self._advertise_host = ray.util.get_node_ip_address()
 
         try:
-            _wait_for_http_health(self._advertise_host, http_port)
+            _wait_for_http_health(
+                host=self._advertise_host,
+                port=http_port,
+                is_alive=lambda: self._server_proc.is_alive(),
+            )
```
andylin-hao, line 259: **`spawn_timeout: 900` from the config never reaches here — the health wait stays at the hardcoded 300 s default.**

`_wait_for_http_health` defaults to `timeout: float = 300.0` and no timeout is passed. Meanwhile the guide added in this PR warns that "server first-time compilation + weight loading is slow, so allow enough time", and the new eval YAML asks for `spawn_timeout: 900`. A first DreamZero run will abort at 5 minutes.

Suggested fix: plumb `spawn_timeout` through `launch_sglang_router_and_server` into this call, or drop the key from the YAML if it isn't meant to be wired.

evaluations/libero/libero_spatial_dreamzero_eval_sglang.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+  sglang:
+    serving_mode: "embodied"
+    spawn_timeout: 900
+    http_timeout_s: 600
+    http_max_retries: 5
+    http_retry_backoff_s: 1.0
+    http_payload_format: "msgpack"
+    seed: 1140
+    debug_sessions: false
```
andylin-hao: **Four keys in this block are never read by any code in the PR** (`git grep` over `*.py` returns nothing for the last three):

- `spawn_timeout: 900` — the health wait is hardcoded to 300 s, see my comment on `server_worker.py`.
- `http_payload_format: "msgpack"` — `SGLangEmbodiedWorker.predict()` hardcodes `msgpack=True`. This one also gets a row in the user-facing config table in `dreamzero_sglang.rst`, so it reads as a supported knob.
- `debug_sessions: false`, `debug_batch: false` — no readers.

Suggested fix: wire `http_payload_format` in `predict()` (`msgpack=(fmt == "msgpack")`) or delete the key from the YAML and both doc tables; same for the two debug flags.

evaluations/libero/libero_spatial_dreamzero_eval_sglang.yaml
```diff
@@ -0,0 +1,127 @@
...
+    launch_router: false
+    server_type: embodied
+    group_name: SGLangServerGroup
+    router_group_name: SGLangRouterGroup
+    server:
+      model_path: ${rollout.model.model_path}
+      backend: sglang
+      pipeline_class_name: DreamZeroPipeline
+      tp_size: ${..tensor_parallel_size}
+      num_gpus: 1  # shoud be TP size * PP size * CFG degree * SP degree
```
andylin-hao: `num_gpus` is a launcher-only key sitting inside the `server` dict, which `docs/source-en/rst_source/guides/sglang_server.rst` — including the edit in this PR — documents as "Forwarded **verbatim** to the selected sglang `ServerArgs`".

`_num_accelerators_per_engine()` reads it for placement width, but it is *not* stripped from `server_kwargs` before `ServerArgs.from_kwargs(**server_kwargs)` in `server_worker.init_server()`. Every other launcher-owned knob (`tensor_parallel_size`, `pipeline_parallel_size`, `server_type`, `launch_server`) lives one level up under `rollout.sglang`.

Suggested fix: move it to `rollout.sglang.num_accelerators_per_engine`, or pop it from `server_kwargs` in `init_server()` and document the exception.

Also a typo in the inline comment: `shoud` → `should`.

rlinf/workers/rollout/sglang_server/launcher.py
```diff
@@ -224,3 +225,12 @@ def launch_sglang_api(
         router_node_rank=router_node_rank,
     )
     return get_sglang_api_url(server_group, router_group), server_group, router_group
+
+
+def _num_accelerators_per_engine(router_server_args: DictConfig) -> int:
+    server_args = router_server_args.get("server") or {}
+    if (num_gpus := int(server_args.get("num_gpus", 0) or 0)) > 0:
+        return num_gpus
```
andylin-hao, line 233: See my comment on the eval YAML: `server.num_gpus` is read here for placement, but it is also forwarded verbatim into the sglang `ServerArgs` constructor in `server_worker.init_server()`. If `sglang.multimodal_gen.runtime.server_args.ServerArgs.from_kwargs` doesn't accept a `num_gpus` field, this will fail at server launch — could you confirm against sgl-project/sglang#30679?

andylin-hao (commented):
Thanks for this — the `SGLangServerWorker` `server_type` dispatch, the launcher/placement rework, and the `InferenceHTTPClient.post` msgpack+retry path are clean and well documented, and splitting the converter (`build_request`/`parse_response`) from the worker's HTTP round-trip is a nice separation.

I've left inline comments. Summarizing what I think blocks merge:

**Blockers**
1. `evaluations/eval_embodied_agent.py:83` — `env_worker_cls` is only bound inside the `huggingface` branch, so the new sglang path raises `NameError` before any worker launches.
2. `rlinf/workers/rollout/utils.py:345` — still imports `sglang_worker_server` / `SGLangWorkerWithHTTPServer`, which this PR renamed. `serving_mode: worker_http` (the agentlightning path) now fails with `ModuleNotFoundError`.
3. `docs/{source-en,source-zh}/rst_source/extending/sglang_embodied_model.rst` — the ~900-line extension guide documents an API that doesn't exist in the PR (`action_policies/`, `EmbodiedActionPolicy`, `@register_action_policy`, `infer()`). 17 stale occurrences per language; neither file mentions the shipped `sglang_adapter` registry. Anyone following it produces code that cannot load.

**Should fix before merge**
4. `reset_mask` hardcoded all-`False` with stable session ids under `auto_reset: True` — server-side session state is never reset at episode boundaries. `final_obs` already carries the signal but is dropped in `_merge_obs_batches`.
5. `rlinf/workers/rollout/sglang/__init__.py` — the version gate's new soft fallback turns a clear `ValueError` into a `NoneType`/`AttributeError` deep in rollout, for all sglang users.
6. `spawn_timeout: 900` is never read (health wait stays at 300 s), plus three more dead keys in the new YAML: `http_payload_format` (documented as a supported knob), `debug_sessions`, `debug_batch`.
7. `dreamzero_config.py` `relative_action_keys` default change looks unrelated and unexplained; `peft` unpinned in a file shared with the existing SFT CI job; `install.sh` loses its executable bit.

**Process / coverage**
- No Dockerfile stage and no e2e/CI job for the new `dreamzero-sglang` model name; `install_dreamzero_sglang_model()` is currently a duplicate of the existing DreamZero path and doesn't install sglang at all.
- The feature needs SGLang built from an unmerged PR (sgl-project/sglang#30679) installed by hand, so nothing here is exercisable in CI and the eval numbers can't be independently reproduced. That's a maintainer call — either gate merge on #30679 landing plus a pinned sglang ref in `install.sh`, or mark the path experimental in the docs.
- The GPU-free parts (the convert-action registry, `http_client.post` retry/backoff) are easily unit-testable and currently have no tests.
- The branch is currently conflicting with `main`; 12 of the 38 commits lack `Signed-off-by` (`git rebase --signoff`), and several subjects aren't Conventional Commits ("add common args", "fix TP check", "modify args and docs", ...). Worth squashing the ~20 lint/style commits.

Verified as correct, for what it's worth: both HF repos referenced by the eval YAML resolve (200, no redirect) and the basenames match their URL comments; `dreamzero_5b.yaml` and `toolkits/lerobot/generate_dreamzero_metadata.py` exist; the `logs/<timestamp>-<config>/eval_embodiment.log` path matches `run_eval.sh`; the `SupportedModel.register` / `EMBODIED_MODEL` snippets are accurate; EN/ZH section counts match (34 each); and all new files carry the correct 2026 RLinf copyright header.

rlinf/models/embodiment/dreamzero/dreamzero_config.py
```diff
@@ -235,7 +235,7 @@ class DreamZeroConfig(VLAConfig):
         default=False, metadata={"help": "Relative action per horizon."}
     )
     relative_action_keys: list = field(
-        default_factory=list, metadata={"help": "Relative action keys."}
+        default=None, metadata={"help": "Relative action keys."}
```
andylin-hao, line 238: This change looks unrelated to the PR and is unexplained.

`default_factory=list` → `default=None` breaks the declared contract: the annotation still says `list`, but the field can now be `None`. Nothing in this PR reads `DreamZeroConfig.relative_action_keys` — the new sglang converter reads `cfg.get("relative_action_keys") or []` off the DictConfig (`sglang_adapter.py:104`) — so this only affects the existing SFT/HF path (`dreamzero_policy.py:153`), with no test covering it.

Suggested fix: revert, or annotate `list | None = None` and explain in the commit message what broke with `default_factory=list`.

requirements/embodied/models/dreamzero.txt
```diff
@@ -3,7 +3,8 @@ tianshou==0.5.1
 albumentations==1.4.18
 ftfy==6.3.1
 diffusers==0.37.1
-peft==0.17.0
+peft>=0.19.1
```
andylin-hao: This unpins a dependency that is shared beyond the new path. Every other line in this file is `==`-pinned (per CONTRIBUTING / the install conventions), and this file also feeds the existing `sft-dreamzero-droid-wan22-test` CI job and the `dreamzero` Docker stage — an open-ended floor lets a future peft release break the non-sglang DreamZero path.

Suggested fix: `peft==0.19.1` (or whichever exact version the sglang path needs). The stray blank line added after it can go too.

requirements/install.sh
```diff
@@ -1891,6 +1891,26 @@ install_dreamzero_model() {
     esac
 }
 
+install_dreamzero_sglang_model() {
```
andylin-hao: **This new `SUPPORTED_MODELS` entry doesn't add any install logic of its own.**

The body is byte-for-byte the `libero` / `maniskill_libero` / `""` subset of `install_dreamzero_model()`, calling the same `install_dreamzero_deps` (which clones the external DreamZero repo and `pip install -e`s it). And it does **not** install sglang or its `diffusion` extra — the docs tell the user to build that by hand from an unmerged PR. So `--model dreamzero-sglang` produces the same venv as `--model dreamzero --env libero`, while adding a name to maintain across `SUPPORTED_MODELS`, `docker/Dockerfile`, and CI.

Two follow-ons:
- There is no `docker/Dockerfile` build stage and no e2e/CI job for `dreamzero-sglang`. Per the add-install-docker-ci-e2e conventions a new model name needs both, plus an e2e config under `tests/e2e_tests/embodied/`.
- This file also loses its executable bit in this PR (`old mode 100755` → `new mode 100644`). Docs invoke it as `bash requirements/install.sh` so most flows survive, but it looks unintended — please `chmod +x` and amend.

Suggested fix: drop the separate model name and reuse `--model dreamzero`, or make the target meaningful by installing the sglang fork at a pinned git ref once #30679 lands.

rlinf/utils/http_client.py
```diff
@@ -207,17 +209,104 @@ def _generate_body(
...
+                if "msgpack" not in resp.headers.get("content-type", "").lower():
+                    raise RuntimeError(
+                        "expected a msgpack response, got content-type="
+                        f"{resp.headers.get('content-type')!r}"
+                    )
+                return unpack_msgpack(resp.content)
+            return resp.json()
+        raise RuntimeError(
+            f"POST {url} failed after {retries + 1} attempts: {last_error}"
         )
```
andylin-hao: This is unreachable: the `for` loop above always either returns or raises, so `last_error` is only ever written, never read. Worth restructuring so the final attempt's 5xx surfaces the accumulated context instead of a bare `raise_for_status()`.

Also, the `except Exception: pass` around the tianshou import in `_to_msgpackable` (line 299) hides real errors — `except ImportError` would be tighter.

Separately: the retry/backoff logic and the msgpack content-type rejection here are pure Python and need no GPU. Along with `register_sglang_adapter` / `get_sglang_adapter_cls`, these are the parts of the PR that could carry unit tests under `tests/`.

docs/source-en/rst_source/extending/sglang_embodied_model.rst
```diff
@@ -0,0 +1,901 @@
...
+Step 4: Import the Policy to Trigger Registration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Add the following in ``rlinf/workers/rollout/sglang/action_policies/__init__.py``:
+
+.. code-block:: python
+
+   from rlinf.workers.rollout.sglang.action_policies import your_model  # noqa: F401,E401
+
+The decorator executes only when the module is imported. If this step is omitted, Worker initialization reports that no
```
andylin-hao: **This guide documents an API that does not exist anywhere in the PR.**

Steps 3 and 4 tell contributors to create `rlinf/workers/rollout/sglang/action_policies/<model>.py`, subclass `EmbodiedActionPolicy`, decorate with `@register_action_policy("your_model")`, implement `infer(env_obs, mode)` with `__init__(cfg, server_url, rank)`, and import the module in `action_policies/__init__.py`. `git ls-tree` on the head commit finds **no `action_policies` directory at all**, and none of those symbols exist.

What the PR actually ships:

| Documented | Actual |
|---|---|
| `rlinf/workers/rollout/sglang/action_policies/<model>.py` | `rlinf/models/embodiment/<model>/sglang_convert_action.py` |
| `@register_action_policy("name")` decorator | `register_sglang_adapter(model_type, builder)` — not a decorator, takes a lazy zero-arg builder |
| base class `EmbodiedActionPolicy` | no base class; duck-typed |
| `infer(env_obs, mode) -> (actions, info)` | `build_request(env_obs, mode) -> (payload, state)` + `parse_response(resp, state) -> (actions, info)`; the HTTP POST lives in the worker |
| `__init__(cfg, server_url, rank)` | `__init__(cfg, rank)` |
| import in `action_policies/__init__.py` | add a builder to `_register_builtin_sglang_adapters()` |

There are 17 occurrences of the stale names in this file and 17 in the ZH counterpart; neither file mentions `sglang_adapter`, `build_request`, or `parse_response` even once. This reads like a guide written against an earlier refactor and never re-synced — anyone following it produces code that cannot load.

The same stale paths appear in the DreamZero file table further down (lines 262-268).

Suggested fix: rewrite Steps 3-4 and the file table against the shipped registry (`rlinf/models/embodiment/sglang_adapter.py:31,53,71`), and mirror the rewrite in `docs/source-zh/.../sglang_embodied_model.rst`.

docs/source-zh/rst_source/extending/sglang_embodied_model.rst
```diff
@@ -0,0 +1,881 @@
...
+步骤四：导入 Policy，触发注册
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+在 ``rlinf/workers/rollout/sglang/action_policies/__init__.py`` 中添加：
+
+.. code-block:: python
+
+   from rlinf.workers.rollout.sglang.action_policies import your_model  # noqa: F401,E401
+
+装饰器只有在模块被导入时才会执行。如果漏掉这一步，Worker 初始化时会报告没有为
```
andylin-hao: 同 EN 版本的问题：这一节描述的 API 在 PR 里并不存在（`action_policies/` 目录、`EmbodiedActionPolicy`、`register_action_policy`、`infer()`、`__init__(cfg, server_url, rank)` 全部查无此物）。实际实现是 `rlinf/models/embodiment/<model>/sglang_convert_action.py` + `register_sglang_adapter(model_type, builder)`（非装饰器，接受 lazy builder），接口是 `build_request()` / `parse_response()`，HTTP POST 由 worker 负责。

本文件有 17 处过时命名，与 EN 版本数量一致，两边都需要同步重写。详见我在 EN 页面上的评论。

docs/source-en/rst_source/evaluations/guides/dreamzero_sglang.rst
```diff
@@ -0,0 +1,123 @@
+DreamZero SGLang Evaluation
+===========================
+
+This guide runs DreamZero LIBERO evaluation through the RLinf SGLang embodied backend. Use this path for inference-only evaluation with an SGLang-native DreamZero checkpoint layout.
+
+Compared with the original DreamZero eval path in :doc:`../../examples/embodied/sft_dreamzero`, this backend does not require ``DREAMZERO_PATH`` or the external DreamZero Python package at runtime. The RLinf rollout worker starts and owns an SGLang action server, sends batched observations over the VLA action API, and denormalizes the returned action chunks before stepping LIBERO.
```
andylin-hao: Two factual errors in this paragraph:

**1. "The RLinf rollout worker starts and owns an SGLang action server" is backwards.** The code says the opposite — `SGLangEmbodiedWorker`'s own module docstring reads "no worker-owned HTTP server, no in-worker subprocess", and `evaluations/eval_embodied_agent.py` launches the server group via `launch_sglang_router_and_server` then pushes URLs down with `set_sglang_server_urls`. The extending guide gets this right (lines 29-33), so the two pages currently contradict each other. Same wording again at line 79 of this file, and in both places in the ZH page.

Suggested wording: "the eval driver launches the SGLang server group and pushes each server URL to the rollout workers; the worker is a thin client that posts observations to the URL assigned to its rank."

**2. "does not require ... the external DreamZero Python package at runtime" is false.** `sglang_adapter.py` imports `rlinf.data.datasets.dreamzero.data_transforms`, whose `__init__.py` begins with `from groot.vla.data.schema import DatasetMetadata`. `groot` is neither vendored in the repo nor listed in `requirements/embodied/models/dreamzero.txt` — it comes from the `pip install -e $dreamzero_path` inside `install_dreamzero_deps`, which the new `dreamzero-sglang` target still calls.

The accurate claim is narrower: the model *weights* live in the sglang process, but the transform stack still needs the package installed.

Minor: line 38 uses `huggingface-cli download`; most current pages use `hf download <org>/<repo> --local-dir <repo>` and `huggingface-cli` is deprecated upstream.

docs/source-zh/rst_source/evaluations/guides/dreamzero_sglang.rst
```diff
@@ -0,0 +1,123 @@
+DreamZero SGLang 评测
+=====================
+
+本文档说明如何通过 RLinf 的 SGLang embodied backend 运行 DreamZero LIBERO 评测。该路径用于只做推理评测的场景，要求 DreamZero checkpoint 已转换为 SGLang-native 的组件化目录。
+
+与 :doc:`../../examples/embodied/sft_dreamzero` 中的原始 DreamZero eval 路径相比，SGLang backend 在运行时不需要 ``DREAMZERO_PATH``，也不依赖外部 DreamZero Python 包。RLinf rollout worker 会启动并管理本地 SGLang action server，通过 VLA action API 发送 batched observation，并将返回的 action chunk 反归一化后送入 LIBERO 环境。
```
andylin-hao: 同 EN 版本的两个事实性错误：

1. **“RLinf rollout worker 会启动并管理本地 SGLang action server”与代码相反。** `SGLangEmbodiedWorker` 的模块 docstring 明确写着 “no worker-owned HTTP server, no in-worker subprocess”；实际由 `evaluations/eval_embodied_agent.py` 调用 `launch_sglang_router_and_server` 拉起 server group，再通过 `set_sglang_server_urls` 把 URL 下发给 worker。本文件第 79 行有同样的表述。

2. **“不依赖外部 DreamZero Python 包”不成立。** `sglang_adapter.py` 导入 `rlinf.data.datasets.dreamzero.data_transforms`，其 `__init__.py` 第一行就是 `from groot.vla.data.schema import DatasetMetadata`；`groot` 既没有 vendored 进仓库，也不在 `requirements/embodied/models/dreamzero.txt` 里，而是来自 `install_dreamzero_deps` 的 `pip install -e $dreamzero_path`。准确的说法是：模型权重跑在 sglang server 进程里，但 data transform 仍然需要该包。

另外第 38 行的 `huggingface-cli download` 建议改成仓库主流的 `hf download`。

rlinf/workers/rollout/sglang_server/server_worker.py
FxxxxU: fix.

rlinf/workers/rollout/sglang_server/server_worker.py
FxxxxU: fix.

rlinf/workers/rollout/sglang_server/launcher.py
FxxxxU: fix.

rlinf/workers/rollout/sglang_server/server_worker.py
FxxxxU: fix.

rlinf/workers/rollout/sglang_server/server_worker.py
FxxxxU: fix.

rlinf/workers/rollout/sglang_server/server_worker.py
FxxxxU: fix.

rlinf/workers/rollout/sglang_server/server_worker.py
FxxxxU: fix.

rlinf/workers/rollout/sglang_server/launcher.py
FxxxxU: The way `num_gpus` is passed in is accepted correctly by sglang

evaluations/eval_embodied_agent.py
FxxxxU: fix.

rlinf/workers/rollout/sglang/__init__.py
FxxxxU: fix.

rlinf/utils/http_client.py
FxxxxU: fix.

rlinf/models/embodiment/dreamzero/sglang_adapter.py
FxxxxU: fix.

rlinf/workers/rollout/sglang/sglang_embodied_worker.py
FxxxxU: fix.

rlinf/models/embodiment/dreamzero/sglang_adapter.py
FxxxxU: fix.

rlinf/workers/rollout/sglang_server/server_worker.py
FxxxxU: fix.

evaluations/libero/libero_spatial_dreamzero_eval_sglang.yaml
FxxxxU: fix.

evaluations/libero/libero_spatial_dreamzero_eval_sglang.yaml
FxxxxU: fix.

rlinf/models/embodiment/dreamzero/dreamzero_config.py
FxxxxU: fix.

requirements/install.sh
FxxxxU: fix.

docs/source-en/rst_source/extending/sglang_embodied_model.rst
FxxxxU: fix.

docs/source-zh/rst_source/extending/sglang_embodied_model.rst
FxxxxU: fix.

docs/source-en/rst_source/evaluations/guides/dreamzero_sglang.rst
FxxxxU: fix.

docs/source-zh/rst_source/evaluations/guides/dreamzero_sglang.rst
FxxxxU: fix.

rlinf/models/embodiment/dreamzero/dreamzero_config.py
FxxxxU: fix.

docs/source-en/rst_source/evaluations/guides/dreamzero_sglang.rst
```diff
@@ -0,0 +1,123 @@
...
+Install the Test Environment
+----------------------------
+
+Set up RLinf with the embodied, LIBERO, and DreamZero SGLang dependencies:
+
+.. code-block:: bash
+
+   cd /path/to/RLinf
+   bash requirements/install.sh embodied --env libero --model dreamzero \
+     --torch 2.11.0 --python 3.11.14 --venv /path/to/dreamzero_test
```
andylin-hao: Remove the --torch 2.11.0 and --python 3.11.14 as these are the default versions now.

andylin-hao (approved):

docs/source-en/rst_source/evaluations/guides/dreamzero_sglang.rst
FxxxxU: fix

PR #1395, opened 2026-07-20
feat(reward): add QwenTrend reward pipeline

NLC2004:
### Description
Add an end-to-end QwenTrend reward pipeline for embodied PPO that combines a sparse terminal-success LoRA reward with a dense potential-based shaping reward.

This includes:

- Uniform rollout collection across policy checkpoints.
- Sparse terminal-success dataset preprocessing and Qwen3-VL LoRA SFT.
- Dense potential/progress dataset preprocessing and Qwen3-VL LoRA SFT.
- Frozen feature extraction and scalar potential-head training.
- PPO training with combined Success + Potential rewards (env reward off).
- Peft LoRA checkpoint export/loading for FSDP-trained Qwen3-VL reward models.
- Example scripts/configs, unit tests, and English/Chinese documentation for the full workflow.

### Motivation and Context
Sparse environment success rewards provide little learning signal on long-horizon manipulation tasks, while raw VLM dense scores can be noisy across adjacent states.

This change introduces a reproducible Success + Potential reward stack: a sparse model detects task completion, and a dense potential model provides bounded progress shaping for PPO. The goal is a more informative reward signal without relying on ground-truth environment reward during training.

### How has this been tested?
Local environment: NVIDIA H100 GPUs; usable GPUs 1 and 2 (`PLACEMENT=1-2`); `NUM_ENVS=512` for collection and PPO.

Unit tests:

```bash
python -m pytest \
  tests/unit_tests/test_lora_loading.py \
  tests/unit_tests/test_qwentrend_success_reward.py \
  tests/unit_tests/test_qwentrend_success_dataset.py \
  -q
```
Result: passed.

We ran the full documented workflow with `EXIT=0` at every stage (local overrides: `PLACEMENT=1-2`, `NUM_ENVS=512`):

1. Uniform collection from checkpoints `0 50 100 150 200` (`run_collect_uniform_checkpoints.sh`)
2. Sparse success preprocess (`run_preprocess_sparse_success_dataset.sh`) + LoRA SFT 400 steps (`run_train_sparse_success_vlm.sh`)
3. Dense potential preprocess (`run_preprocess_dense_potential_dataset.sh`) + LoRA SFT 400 steps (`run_train_dense_potential_vlm.sh`)
4. Feature extraction + scalar-head training (`run_train_dense_scalar_head.sh`, `CUDA_DEVICES=1,2`, `FEATURE_WORLD_SIZE=2`)
5. Combined Success + Potential PPO (`run_qwentrend_success_reward.sh`, `MAX_STEPS=20`; checkpoint `global_step_20` saved)

Sparse success eval balanced accuracy: `0.994@100`, `0.983@200`, `0.980@300/400` (selected `global_step_100`). Dense potential selected at `global_step_400`.

Docs recommend the generic 4-GPU setup (`PLACEMENT=0-3`, `NUM_ENVS=1024`). Local testing used GPUs 1–2 because GPUs 0 and 3 on the test machine were unavailable for rendering.

Artifacts:

- Success: `logs/qwentrend_fulltest_success_train/.../global_step_100`
- Potential: `logs/qwentrend_fulltest_potential_train/.../global_step_400`
- Scalar: `logs/qwentrend_fulltest_scalar_head/best.pt`
- PPO: `logs/qwentrend_fulltest_ppo/`
- Summary: `logs/doc_fulltest_summary.txt`

### Types of changes

### Checklist:


rlinf/workers/sft/fsdp_vlm_sft_worker.py
```diff
@@ -203,16 +233,91 @@ def get_eval_model_output(self, batch: dict[str, Any]):
...
+            A dict with overall accuracy, positive-class recall, negative-class
+            accuracy, balanced accuracy and per-class totals.
+        """
+        positive_accuracy = counts["positive_correct"] / max(
+            1, counts["positive_total"]
+        )
+        negative_accuracy = counts["negative_correct"] / max(
+            1, counts["negative_total"]
+        )
+        return {
```
zhoutianxing7B: Returning different result types based on each individual batch can produce incorrect dataset-level metrics. In the 0–9 potential dataset, a batch containing only 0/1 labels returns counts, while other batches return an integer. run_eval() then enters the class_counts branch and excludes the integer results from the final metric. Could we always return one count dictionary, track binary_total separately, and compute balanced accuracy only when the entire eval dataset is binary? Please add a regression test with 0/1/2 labels across different batches.

rlinf/models/embodiment/reward/vlm_reward_model.py
```diff
@@ -245,42 +355,207 @@ def _slice_batch_value(self, value: Any, start: int, end: int) -> Any:
...
-            del batched_inputs
-            reward_chunk[valid_input_ids] = parsed_rewards.to(dtype=torch.float32)
             reward_chunks.append(reward_chunk)
+            valid_chunks.append(valid_chunk)
+            success_chunks.append(success_chunk)
 
         rewards = torch.cat(reward_chunks, dim=0)
+        if self.inference_mode == "scalar_head":
+            valid_mask = torch.cat(valid_chunks, dim=0)
+            dones = observations.get("dones")
```
zhoutianxing7B: HistoryVLMRewardModel now relies on reward_input["dones"] to reset _previous_potentials, _success_fired, and _success_streak. However, EnvOutput.dones is documented as [B], while env_worker.py only inserts it into reward_input inside the ndim > 1 branch. In the normal 1-D path, these states therefore leak across auto-reset episodes. Please always attach dones after the optional last-dimension slicing, and add an integration test that exercises reward computation across two episodes through the env-worker path.

rlinf/workers/sft/fsdp_vlm_sft_worker.py
```diff
@@ -42,6 +43,17 @@ def save_checkpoint(self, save_path: str, step: int = 0):
         super().save_checkpoint(save_path, step)
         if self._rank == 0:
             self._save_data_state(save_path)
+        # VLMRewardModel loads success/potential adapters from lora_* keys in
```
zhoutianxing7B: This rewrites the framework-defined model_state_dict/full_weights.pt with adapter-only tensors for every LoRA VLM SFT job, not only QwenTrend. Existing consumers may reasonably expect this path to contain the full model state. Please preserve full_weights.pt and export the adapter separately, together with its PEFT configuration, then make the QwenTrend loader consume the adapter artifact explicitly.

rlinf/workers/sft/lora_checkpoint.py
```diff
@@ -0,0 +1,98 @@
...
+        lora_state = export_lora_state_dict(model)
+    except ValueError as error:
+        if rank == 0 and log_warning is not None:
+            log_warning(f"Skip LoRA rewrite at {save_path}: {error}")
+        torch.distributed.barrier()
+        return False
+
+    torch.distributed.barrier()
+    wrote = False
+    if rank == 0:
```
zhoutianxing7B: If rank 0 raises during validation or torch.save(), every other rank will block indefinitely at the final barrier. Please catch the rank-0 I/O result, broadcast a success/error status to all ranks, and only then let every rank return or raise consistently. A distributed regression test for a simulated save failure would be useful.

docs/source-zh/rst_source/extending/reward_model.rst
```diff
@@ -7,6 +7,177 @@ reward，也包括 QwenTrend / ``HistoryVLMRewardModel`` 这类 VLM reward。
...
+
+.. code-block:: bash
+
+   export QWEN_MODEL_PATH=/path/to/Qwen3-VL-4B-Instruct
+   export QWENTREND_POTENTIAL_CHECKPOINT="$(cat /path/to/qwentrend_potential_vlm_train/SELECTED_CKPT.txt)"
+   export QWENTREND_SCALAR_HEAD=/path/to/qwentrend_scalar_head/best.pt
+   export QWENTREND_SUCCESS_CHECKPOINT=/path/to/qwentrend_sparse_success_sft/.../global_step_300
+   export POLICY_CHECKPOINT=/path/to/policy/full_weights.pt
+   export PPO_OUTPUT_ROOT=/path/to/ppo-output
+   CUDA_VISIBLE_DEVICES=0,1,2,3 PLACEMENT=0-3 NUM_ENVS=1024 MAX_STEPS=160 \
```
zhoutianxing7B: The docs say NUM_ENVS=1024 matches the script defaults, but the PPO launcher defaults to 128. Please distinguish the collection default from the PPO default, or align the scripts and both EN/ZH docs.

examples/sft/config/vlm_trend_sft_success.yaml
```diff
@@ -0,0 +1,120 @@
+defaults:
```
zhoutianxing7B, line 1: Please run pre-commit run --all-files and add the standard 2026 RLinf license header to the newly added shell scripts and YAML config. The current setup_processor() indentation/whitespace also appears not to be Ruff-formatted.

zhoutianxing7B (commented):
Thanks for the contribution. I left several inline comments. Could you please address or clarify the correctness concerns before merging?

rlinf/workers/sft/fsdp_vlm_sft_worker.py
NLC2004: Thanks — good catch. Fixed as suggested:

- `get_eval_model_output()` now always returns a single count dict (never an int), and tracks `binary_total` separately.
- Per-class counts are only updated for gold labels in `{0, 1}`.
- `compute_eval_metrics()` reports balanced accuracy / class metrics only when `binary_total == total` (i.e. the entire eval set is binary); otherwise it returns overall `eval_accuracy` only.

Added a regression test with labels `0/1` in one batch and `2/0` in another, going through the real `get_eval_model_output` → `run_eval` path, to ensure no batches are dropped and balanced accuracy is not reported for mixed potential labels.

rlinf/models/embodiment/reward/vlm_reward_model.py
NLC2004: Thanks — agreed. Fixed on the env-worker side:

- `get_reward_model_output()` now always attaches `dones` to `reward_input` when present.
- For 2-D chunk dones we reduce with `.any(dim=-1)` (instead of `[:, -1]`) so mid-chunk terminations are not missed, consistent with the eval / terminal-reward paths. 1-D `[B]` dones are attached as-is.

Added an integration test that builds `reward_input` via the env-worker path (including 1-D dones) and feeds it into `HistoryVLMRewardModel.compute_reward` across two episodes, verifying that `_previous_potentials` / `_success_fired` / `_success_streak` reset on done and do not leak across auto-reset.

rlinf/workers/sft/fsdp_vlm_sft_worker.py
NLC2004: Thanks for the catch — agreed.

`full_weights.pt` is now left untouched as the framework full-model checkpoint. On LoRA VLM SFT saves we export a separate Peft adapter artifact under `actor/lora_adapter/` (`adapter_model.bin` + `adapter_config.json`). The QwenTrend reward loader (`lora_loading.py`) resolves and loads that adapter artifact explicitly instead of scraping `lora_*` keys from `full_weights.pt`.

rlinf/workers/sft/lora_checkpoint.py
NLC2004: Good point — fixed.

Rank-0 validation / `torch.save` are now wrapped so a failure cannot leave other ranks waiting on the final barrier. The result is broadcast to all ranks (`ok` / `skip` / `error`), and every rank then returns or raises consistently. Added a distributed regression test that simulates a rank-0 save failure and checks that all ranks exit with an error instead of hanging.

docs/source-zh/rst_source/extending/reward_model.rst
NLC2004: Aligned the PPO launcher with the docs by changing the default in `examples/embodiment/run_qwentrend_success_reward.sh` from `128` to `1024`. Collection (`run_collect_uniform_checkpoints.sh`) and PPO now both default to `NUM_ENVS=1024`, so the EN/ZH docs statement that the examples match the script defaults is accurate without further doc edits.

examples/sft/config/vlm_trend_sft_success.yaml
NLC2004: Addressed in this commit:

- Added the standard 2026 RLinf license header to the newly added QwenTrend Success shell scripts and `examples/sft/config/qwen3vl_sft_qwentrend_success.yaml`.
- Ran `pre-commit run --all-files` (ruff, ruff-format, and signoff checks all passed).
- Applied Ruff formatting fixes, including cleanup around `setup_processor()` in `vlm_reward_model.py`.

rlinf/models/embodiment/reward/vlm_reward_utils/lora_loading.py
```diff
@@ -0,0 +1,131 @@
+# Copyright 2026 The RLinf Authors.
```
zhoutianxing7B: Extracting LoRA load logic out of vlm_reward_model is good, but a separate lora_loading.py under vlm_reward_utils/ duplicates lora_checkpoint.py and splits save/load across worker vs reward packages. Please merge load + export + legacy full_weights.pt fallback into one shared module (extend lora_checkpoint.py or add rlinf/utils/lora_adapter.py). Reward model should call that shared API instead of maintaining a second loader. Also restore backward-compatible loading from full_weights.pt in that single place.

rlinf/workers/sft/lora_checkpoint.py
```diff
@@ -0,0 +1,195 @@
...
+
+def get_active_peft_config(peft_model: torch.nn.Module) -> Any:
+    """Return the active adapter's Peft config from a Peft-wrapped module."""
+    adapter_name = getattr(peft_model, "active_adapter", "default")
+    if isinstance(adapter_name, (list, tuple)):
+        adapter_name = adapter_name[0]
+    return peft_model.peft_config[adapter_name]
+
+
+def save_lora_adapter_files(
```
zhoutianxing7B: Manual torch.save(adapter_model.bin) + handwritten adapter_config.json reinvents Peft’s standard artifact format. RLinf already loads adapters via PeftModel.from_pretrained in rlinf/models/__init__.py:apply_lora.

examples/reward/vlm_trend/extract_potential_features.py
```diff
@@ -0,0 +1,280 @@
...
+            raise ValueError(
+                f"Missing dual-view image at {row['source_episode_path']}:{index}"
+            )
+        main_frames.append(_to_uint8_rgb(main))
+        extra_frames.append(_to_uint8_rgb(extra))
+    return main_frames, extra_frames
+
+
+@torch.no_grad()
+def encode(
```
zhoutianxing7B: encode() reimplements last-token hidden-state pooling that VLMRewardModel.compute_scalar_potential already does. Training and online inference can silently diverge.

Suggested fix: Add something like VLMRewardModel.extract_prompt_features(batched_inputs) -> Tensor (or a small util next to the reward model) and call it from both the scalar-head path and this script. Do not keep a second copy of the pooling math in scripts/.

examples/reward/vlm_trend/preprocess_terminal_success.py
```diff
@@ -0,0 +1,336 @@
+# Copyright 2026 The RLinf Authors.
```
zhoutianxing7B: Success/Potential preprocessing reintroduces episode I/O and dual-view helpers already present in preprocess_qwentrend_reward_dataset.py. Worse, scripts/extract_qwentrend_potential_features.py imports private helpers (_to_uint8_rgb, _extract_extra_view_image) across example modules — that coupling will break easily.

Suggested fix: Factor shared helpers into one place (e.g. examples/reward/qwentrend_data_utils.py or rlinf/data/datasets/qwentrend_io.py): episode inspect, dual-view extract, uint8 convert, split hashing. All preprocess / feature scripts should import the shared API; no private cross-imports, no copy-paste of _to_uint8_rgb.

rlinf/models/embodiment/reward/vlm_reward_utils/input_builder.py
```diff
@@ -318,3 +376,91 @@ def process_inputs(self, prepared_inputs: dict[str, Any]):
             answer_text=None,
         )
         return processed_inputs
+
+
+@register_input_builder("qwentrend_terminal_success_input_builder")
```
zhoutianxing7B: qwentrend_terminal_success_input_builder and qwentrend_potential_input_builder largely copy QwentrendInputBuilder.prepare_inputs, differing only in prompt text. That is registry bloat.

Suggested fix: Keep one video/history builder and pass the prompt via input_builder_params (template / prompt_fn / num_bins). Register new builders only when the data shape or packing changes, not for string variants. Align names with #1418 (vlm_trend_*) after rebase.

NLC2004:
###  Run before the second round of reviewer fixes. Config for this `from68` QwenTrend Success PPO:
- **Start policy:** baseline MLP `global_step_68` (`eval/success_at_end` ~ 0.42)
- **Envs:** `NUM_ENVS=1024` (train + eval)
- **Reward:** `success_threshold=0.25`
- **PPO:**`lr=3e-5`, `value_lr=5e-5`, `clip=0.2`
- **Other:** `env_reward_weight=0.0`, `MAX_STEPS=160`, 2xA800
`eval/success_at_end` curve below (step 4-54, peak ~0.75):
from68_success_curve_step4_54


### Run before the second round of reviewer fixes
Config for this VLM-Trend Success PPO: `global_step_82` mid-policy start
- **Start policy:** baseline MLP `global_step_82` (~0.58 `eval/success_at_end`)
- **Envs:** `NUM_ENVS=1024` (train + eval)
- **Reward:** `success_threshold=0.5`
- **PPO:** `lr=1e-5`, `value_lr=3e-5`, `clip=0.1`
- **Other:** `MAX_STEPS=160`, 2xA800
`eval/success_at_end` curve below (start → peak, step 5–20, peak **0.710**):

success_curve_start_to_peak_v2

rlinf/models/embodiment/reward/vlm_reward_utils/lora_loading.py
NLC2004: Thanks — addressed.

I removed the separate `vlm_reward_utils/lora_loading.py` and consolidated load + export + legacy `full_weights.pt` fallback into one shared module: `rlinf/utils/lora_adapter.py`.

The reward model now calls that shared API (`load_adapter_onto_model` / `export_lora_adapter`) instead of maintaining a second loader. Backward-compatible loading from legacy `full_weights.pt` is handled only in this single place.

rlinf/workers/sft/lora_checkpoint.py
NLC2004: Thanks — addressed.

I dropped the manual `torch.save(adapter_model.bin)` + handwritten `adapter_config.json` path. Adapter export now uses `PeftModel.save_pretrained`, which matches how RLinf already loads adapters via `PeftModel.from_pretrained` / `apply_lora`.

SFT still keeps the framework `full_weights.pt`; the Peft adapter is exported beside it under `actor/lora_adapter/`.

examples/reward/vlm_trend/extract_potential_features.py
NLC2004: Thanks — addressed.

I added `VLMRewardModel.extract_prompt_features(batched_inputs) -> Tensor` for last-attended-token pooling.

Both the online scalar-head path (`compute_scalar_potential`) and `scripts/extract_vlm_trend_potential_features.py` now call this shared method. The duplicated `encode()` pooling logic in the script has been removed, so training and online inference stay aligned.

examples/reward/vlm_trend/preprocess_terminal_success.py
NLC2004: Thanks — addressed.

I factored the shared episode helpers into `rlinf/data/datasets/vlm_trend_io.py`:
- episode inspect
- dual-view frame extraction
- uint8 conversion
- split hashing
- shared prompts

All preprocess / feature scripts now import this public API. Private cross-module imports (e.g. `scripts/` importing `_to_uint8_rgb` / helpers from `examples/`) have been removed, and the duplicated I/O copy-paste is gone.

rlinf/models/embodiment/reward/vlm_reward_utils/input_builder.py
NLC2004: Thanks — addressed.

I kept a single video/history builder: `vlm_trend_reward_input_builder`. Prompt variants are passed via `input_builder_params` (`prompt_template` / `num_bins`, etc.), instead of registering new builders for string-only differences.

New builders are only intended when data shape / packing changes. Naming is also aligned with #1418 (`vlm_trend_*`) after rebase/merge.

NLC2004:
Thanks for the review — addressed the multi-entrypoint feedback as follows:

**Preprocess / train**
- Single public preprocess entry: `examples/reward/preprocess_vlm_trend_dataset.py --mode {terminal_success,potential,progress,trend_reward}`.
- Mode implementations live under `examples/reward/vlm_trend/` (helpers stay in `vlm_trend_io` / `vlm_trend_success`).
- Teacher training: `examples/reward/run_vlm_trend.py --stage train_teacher` (not folded into `train_reward_model.py` — different stack: argparse MLP on state histories vs Hydra/FSDP ResNet).
- `--mode progress` is kept for compatibility only and documented as **legacy**; the Success pipeline uses `--mode potential` (which already emits potential + progress windows).

**Shells / scripts**
- Removed `examples/embodiment/vlm_trend_success/` and `run_vlm_trend_success_reward.sh`.
- Removed top-level `scripts/extract_vlm_trend_*` / `train_vlm_trend_scalar_head.py`.
- Scalar head: `examples/reward/run_vlm_trend_scalar_head.py --stage extract|train|all`.
- PPO: `bash examples/embodiment/run_embodiment.sh maniskill_ppo_mlp_vlm_trend_success` (+ env vars for checkpoints).
- SFT unchanged: `run_vlm_sft.sh` + config-name.
- Collection is documented as an `eval_embodied_agent.py` loop with the existing collect YAML (no branded shell wrapper).

**Tests / placement**
- Merged potential/success reward unit tests into `test_vlm_trend_reward.py`.
- Kept framework-contract tests (LoRA load/export, env dones→reward, video collate, small dataset helpers).
- Moved `StateSuccessValue` teacher to `rlinf/utils/state_success_value.py`.
- `export_lora_adapter` remains opt-in via `actor.model.is_lora`.

EN/ZH `reward_model.rst` and SFT YAML comments updated to the new entrypoints.

examples/reward/vlm_trend/__init__.py
zhoutianxing7B, line 1: Why is Success/Potential a whole examples/reward/vlm_trend/ package with ~10 scripts? The existing trend recipe is one flat file (preprocess_vlm_trend_reward_dataset.py) plus shared runners/YAMLs. Please follow that: one (or at most two) flat scripts under examples/reward/, with --mode / --stage branches inside the file. Delete the vlm_trend/ package of parallel preprocess/train/extract modules and the importlib dispatchers. Shared I/O can stay in rlinf/data/datasets/vlm_trend_io.py; recipe code should not become a mini-library.

examples/reward/preprocess_vlm_trend_success_dataset.py
zhoutianxing7B, line 1: preprocess_potential.py (~575 lines) and preprocess_progress.py (~448 lines) still duplicate the episode→window→manifest pipeline. Merge into one implementation with --mode branches. If progress is legacy-only, keep a thin compatibility path or remove it — do not maintain two near-copy scripts.

examples/reward/run_vlm_trend.py
zhoutianxing7B, line 1: We still have three launchers (preprocess_vlm_trend_dataset.py, run_vlm_trend.py, run_vlm_trend_scalar_head.py). Please collapse to a single user-facing entry (or fold stages into the one flat recipe script). One recipe should not need three dispatchers.

rlinf/workers/sft/fsdp_vlm_sft_worker.py
```diff
@@ -42,6 +43,16 @@ def save_checkpoint(self, save_path: str, step: int = 0):
         super().save_checkpoint(save_path, step)
         if self._rank == 0:
             self._save_data_state(save_path)
+        # Preserve framework full_weights.pt (merged full state). Export Peft
+        # adapters separately so VLMRewardModel can load them explicitly.
+        if bool(self.cfg.actor.model.get("is_lora", False)):
```
zhoutianxing7B: export_lora_adapter still runs for every VLM SFT save when is_lora: true. That is not Success-only opt-in and adds barriers/I/O for existing LoRA SFT jobs. Please gate with an explicit flag (e.g. actor.model.export_lora_adapter, default false) and enable it only in Success YAMLs.

rlinf/models/__init__.py
```diff
@@ -293,6 +294,89 @@ def _build_steam_value_model(cfg: DictConfig, torch_dtype):
 _register_builtin_models()
 
 
+def apply_lora(model: torch.nn.Module, cfg: DictConfig) -> torch.nn.Module:
```
zhoutianxing7B: Two apply_lora concerns: (1) Qwen3-VL target_modules dropping bare "proj" changes all Qwen3-VL LoRA training, not just Success — please make it config-gated or call out as breaking. (2) PeftModel.from_pretrained(cfg.lora_path) does not use resolve_lora_adapter_dir / legacy fallback from rlinf/utils/lora_adapter.py, so train vs reward checkpoint paths diverge. Please share one resolver.

examples/reward/vlm_trend/__init__.py
NLC2004: Done. The `examples/reward/vlm_trend/` package is deleted entirely, along with all three importlib dispatchers.

The Success recipe is now two flat scripts under `examples/reward/`, matching the existing trend recipe layout:

- `preprocess_vlm_trend_success_dataset.py` — `--mode {terminal_success,potential}`
- `train_vlm_trend_success_model.py` — `--stage {teacher,extract,scalar_head}`

Both dispatch through a plain dict of handlers inside the file, no dynamic imports. Script count for this recipe went from 10 to 2 (plus the pre-existing flat `preprocess_vlm_trend_reward_dataset.py` for classic trend).

Shared I/O stays in `rlinf/data/datasets/vlm_trend_io.py` as you suggested, and the state-success teacher MLP moved to `rlinf/utils/state_success_value.py` so the recipe scripts are CLI-only.

examples/reward/preprocess_vlm_trend_success_dataset.py
NLC2004: Merged. `preprocess_potential.py` and `preprocess_progress.py` are both gone; the episode→window→manifest pipeline now exists once, inside `preprocess_vlm_trend_success_dataset.py --mode potential`.

Progress is not a second script anymore — it is emitted as an additional sample type from the same pass over the episode (`sample_type: "progress"` rows sharing the same reservoir/window machinery), so there is no near-copy to maintain. It is documented as legacy-only in the reward model docs.

examples/reward/run_vlm_trend.py
NLC2004: Done. All three launchers (`preprocess_vlm_trend_dataset.py`, `run_vlm_trend.py`, `run_vlm_trend_scalar_head.py`) are deleted.

Stages are folded into the two flat recipe scripts as `--mode` / `--stage` branches, so there is one user-facing entry per phase:

    python examples/reward/preprocess_vlm_trend_success_dataset.py --mode potential ...
    python examples/reward/train_vlm_trend_success_model.py --stage teacher ...

`--help` without a mode/stage prints the top-level usage; with one, it forwards to that branch's parser. Docs (EN/ZH) are updated to the new commands.

rlinf/workers/sft/fsdp_vlm_sft_worker.py
NLC2004: Fixed. The export is now gated on an explicit flag that defaults to false, so existing LoRA SFT jobs pay no extra barriers or I/O:

    # Opt-in Peft adapter export beside preserved full_weights.pt.
    # Default off so ordinary LoRA SFT jobs are not charged extra I/O/barriers.
    if bool(self.cfg.actor.model.get("export_lora_adapter", False)):
        export_lora_adapter(...)

`export_lora_adapter: true` is set only in `vlm_trend_sft_success.yaml` and `vlm_trend_sft_potential.yaml`. Classic trend reward SFT (`vlm_trend_sft_reward.yaml`) does not set it and keeps relying on the existing `full_weights.pt` fallback.

rlinf/models/__init__.py
NLC2004: Both addressed.

**(1) `target_modules` is now config-gated, not a silent behaviour change.** The framework default is unchanged from `main` and still includes bare `"proj"`. Qwen3-VL recipes opt out explicitly in their own YAML rather than the framework special-casing them:

    def resolve_lora_target_modules(cfg: DictConfig) -> list[str]:
        """Resolve LoRA ``target_modules`` from ``cfg`` or the framework default."""

It accepts either a list or a comma-separated string, so the existing DreamZero-style `lora_target_modules: q,k,v,o,ffn.0,ffn.2` spelling keeps working. The Success/potential SFT YAMLs carry an explicit Qwen3-VL-safe list that omits bare `"proj"`. No existing LoRA training path changes behaviour.

**(2) Train and reward now share one resolver.** `apply_lora` routes through `load_adapter_onto_model` from `rlinf/utils/lora_adapter.py`, so checkpoint-layout resolution and the legacy fallback are identical on both sides. Hub repo IDs are preserved via a narrow fallback, since the shared loader only resolves local checkpoint layouts:

    try:
        model = load_adapter_onto_model(model, cfg.lora_path, adapter_name="default", is_trainable=True)
    except FileNotFoundError:
        # Shared loader only resolves local checkpoint layouts. Hub repo
        # IDs (and other Peft-remote sources) keep the historical contract.
        model = PeftModel.from_pretrained(model, cfg.lora_path, is_trainable=True)

Covered by new unit tests in `tests/unit_tests/test_lora_loading.py` (default target modules, list/CSV override, empty-override rejection, shared-loader path, and the Hub-ID fallback).

rlinf/models/embodiment/reward/vlm_reward_utils/input_builder.py
```diff
@@ -266,6 +266,19 @@ class VLMTrendRewardInputBuilder(VideoVLMInputBuilder):
...
+        task = str(task or self.default_task_description).strip()
+        return (
+            f"You are currently performing the task: {task}. "
+            "You are given two synchronized 5-frame videos from different camera "
+            "views (main view and third-person view) of the same robot action "
+            "window. Judge whether the action trend is positive, negative, or "
+            "unclear. Answer with exactly one word: positive, negative, or unclear."
+        )
+
+    def _video_fps(self) -> float:
```
Lin-xs: Please see the following comment.

rlinf/models/embodiment/reward/vlm_reward_utils/input_builder.py
```diff
@@ -311,5 +317,30 @@ def process_inputs(self, prepared_inputs: dict[str, Any]):
...
+        task = str(task or self.default_task_description).strip()
+        task_text = f" Task: {task}." if self.include_task and task else ""
+        return self.prompt_template.format(
+            task=task,
+            task_text=task_text,
+            num_bins=self.num_bins,
+            num_bins_max=self.num_bins - 1,
+        )
+
+    def _video_fps(self) -> float:
```
Lin-xs: This private method just returns a public attribute, which is unnecessary. Please use `self.video_fps` directly, and keep the same pattern in the base class / subclasses.

rlinf/models/embodiment/reward/vlm_trend_success_potential_reward_model.py
```diff
@@ -0,0 +1,428 @@
...
+    BufferedVLMInputBuilder,
+    get_input_builder,
+)
+from rlinf.models.embodiment.reward.vlm_reward_utils.reward_parser import (
+    BaseRewardParser,
+    get_reward_parser,
+)
+
+
+def _rlinf_checkpoint_state(path: str) -> dict[str, torch.Tensor] | None:
```
Lin-xs: Please ask users to provide full path for `full_weights.pt`. It's not necessary to detect automatically.

rlinf/models/embodiment/reward/vlm_trend_success_potential_reward_model.py
```diff
@@ -0,0 +1,428 @@
...
+            candidate
+            for candidate in (
+                Path(path),
+                Path(path) / "lora_adapter",
+                Path(path) / "actor" / "lora_adapter",
+            )
+            if (candidate / "adapter_config.json").is_file()
+        ),
+        None,
+    )
```
Lin-xs: Same issue, please ask users to provide full path of `lora_adapter` directly.

examples/reward/preprocess_vlm_trend_reward_dataset.py
```diff
@@ -772,5 +849,891 @@ def main() -> None:
...
+    pre.add_argument("-h", "--help", action="store_true")
+    known, remaining = pre.parse_known_args(argv)
+    if known.help:
+        remaining = ["--help"]
+    if known.mode == "trend":
+        run_trend(remaining)
+        return
+    parser = argparse.ArgumentParser(description=f"preprocess --mode {known.mode}")
+    _MODE_ADDERS[known.mode](parser)
+    _MODE_RUNNERS[known.mode](parser.parse_args(remaining))
```
Lin-xs: If these modes don’t depend on each other, please put each mode in its own file instead of keeping them together.

rlinf/models/embodiment/reward/vlm_reward_utils/input_builder.py
NLC2004: OK,I will.

rlinf/models/embodiment/reward/vlm_reward_utils/input_builder.py
NLC2004: Done. `_video_fps()` is removed from the base class and both subclasses. `VideoVLMInputBuilder` now exposes `video_fps` (default `24.0`), and `process_inputs` uses `self.video_fps` directly.

rlinf/models/embodiment/reward/vlm_trend_success_potential_reward_model.py
NLC2004: Done. Checkpoint-root search is removed. `load_lora_adapter()` now requires either the full path to `full_weights.pt` (typically `.../actor/model_state_dict/full_weights.pt` from VLM SFT) or a PEFT adapter directory that itself contains `adapter_config.json`. Passing a checkpoint root raises `FileNotFoundError`. Docs, YAML comments, and unit tests are updated.

rlinf/models/embodiment/reward/vlm_trend_success_potential_reward_model.py
NLC2004: Done, same contract as above. We no longer walk `lora_adapter` / `actor/lora_adapter`. The path must be the adapter directory itself (with `adapter_config.json`) or the `full_weights.pt` file. The documented default example is the SFT output file, since current VLM SFT writes `full_weights.pt` rather than exporting a PEFT directory.

examples/reward/preprocess_vlm_trend_reward_dataset.py
NLC2004: Done. Independent modes now live in their own scripts:

- Trend: `examples/reward/preprocess_vlm_trend_reward_dataset.py`
- Terminal success: `examples/reward/preprocess_vlm_trend_terminal_success.py`
- Potential: `examples/reward/preprocess_vlm_trend_potential.py`
- Feature extraction: `examples/reward/extract_vlm_trend_potential_features.py`

Shared episode I/O is in `examples/reward/vlm_trend_data.py`. The old `--mode` dispatcher is gone; the Trend script errors with a pointer if `--mode` is still passed. EN/ZH docs and tests cover the split.

examples/reward/preprocess_vlm_trend_reward_dataset.py
```diff
@@ -772,5 +720,37 @@ def main() -> None:
     print("=" * 80)
 
 
+_MOVED_MODES = {
```
Lin-xs: Delete these outdated error handler.

tests/unit_tests/models/test_vlm_trend_reward.py
```diff
@@ -0,0 +1,203 @@
+# Copyright 2026 The RLinf Authors.
```
Lin-xs: Please delete all unit_tests.

docs/source-zh/rst_source/extending/reward_model.rst
```diff
@@ -395,10 +491,52 @@ VLM Trend reward 在线推理共用以下核心字段（ ``model_type`` 始终
...
+该配方基于现有 VLM Trend 配方，使用 1,024 个环境，每 5 个 PPO step 评估一次，
+关闭环境奖励，默认训练 160 个 step：
+
+.. code-block:: bash
+
+   export POLICY_CHECKPOINT=/path/to/evaluated_approximately_5pct/actor/model_state_dict/full_weights.pt
+   export VLM_MODEL_PATH=/path/to/Qwen3-VL-4B-Instruct
+   export VLM_TREND_SUCCESS_CHECKPOINT=/path/to/success/selected_global_step/actor/model_state_dict/full_weights.pt
+   export VLM_TREND_POTENTIAL_CHECKPOINT=/path/to/potential/selected_global_step/actor/model_state_dict/full_weights.pt
+   export VLM_TREND_SCALAR_HEAD=/path/to/scalar_head/best.pt
```
Lin-xs: Please set these fields in the yaml instead of via environment variables.

docs/source-zh/rst_source/extending/reward_model.rst
```diff
@@ -236,6 +239,99 @@ RLinf 支持两条 reward 训练路径。``examples/reward/run_reward_training.s
...
+           --manifest "$PIPELINE_ROOT/potential_data/$split/segments.jsonl" \
+           --sample-type "$sample_type" --rank "$rank" --world-size 4 \
+           --device cuda:0 \
+           --output "$PIPELINE_ROOT/features/${split}_${sample_type}_${rank}.pt" &
+       done
+       wait
+     done
+   done
+
+   python examples/reward/train_vlm_trend_auxiliary.py \
```
Lin-xs: Please set these fields in the yaml instead of via environment variables.
Yaml also support relative reference.

examples/reward/vlm_trend/extract_potential_features.py
```diff
@@ -0,0 +1,240 @@
+# Copyright 2026 The RLinf Authors.
```
Lin-xs, line 1: Could you move all `vlm_trend`-related files into a single folder under `reward/`?

docs/source-zh/rst_source/extending/reward_model.rst
```diff
@@ -236,6 +239,99 @@ RLinf 支持两条 reward 训练路径。``examples/reward/run_reward_training.s
...
+   bash examples/sft/run_vlm_sft.sh vlm_trend_success_sft
+
+   export VLM_TREND_POTENTIAL_DATA_ROOT="$PIPELINE_ROOT/potential_data"
+   export OUTPUT_ROOT="$PIPELINE_ROOT/potential_sft"
+   bash examples/sft/run_vlm_sft.sh vlm_trend_potential_sft
+
+将 ``POTENTIAL_CHECKPOINT`` 指向选中的 Potential VLM SFT 权重文件
+（通常是 ``.../global_step_*/actor/model_state_dict/full_weights.pt``）。
+如果你自行导出过 PEFT adapter，也可以传包含 ``adapter_config.json`` 的目录。
+在四张 GPU 上提取冻结 feature，再训练 scalar head：
```
Lin-xs: Is 4 GPUs required for feature extraction? Could we use RLinf’s cluster config to handle the distributed setup automatically?

docs/source-zh/rst_source/extending/reward_model.rst
```diff
@@ -236,6 +239,99 @@ RLinf 支持两条 reward 训练路径。``examples/reward/run_reward_training.s
...
+
+.. code-block:: bash
+
+   export POTENTIAL_CHECKPOINT=/path/to/potential/selected_global_step/actor/model_state_dict/full_weights.pt
+   mkdir -p "$PIPELINE_ROOT/features"
+
+   for split in train eval; do
+     for sample_type in potential progress; do
```
Lin-xs: Could you move this nested loop into a Python file and drive it with RLinf’s cluster config? This will make it more clear.

examples/reward/preprocess_vlm_trend_reward_dataset.py
zhoutianxing7B: Done. I removed the outdated --mode compatibility/error handling. The entrypoint now directly runs Trend preprocessing.

tests/unit_tests/models/test_vlm_trend_reward.py
zhoutianxing7B: Done. I removed all unit-test changes introduced by this PR, including the added history-manager test. The final diff against main no longer contains unit-test changes.

docs/source-zh/rst_source/extending/reward_model.rst
zhoutianxing7B: Done. The policy checkpoint, base VLM, Success/Potential checkpoints, and scalar head are now configured under vlm_trend_paths in the YAML and referenced by the runtime fields.

docs/source-zh/rst_source/extending/reward_model.rst
zhoutianxing7B: Done. The pipeline and SFT paths are now defined in YAML, with data and output paths derived through OmegaConf references. The documentation no longer requires exporting these path variables.

examples/reward/vlm_trend/extract_potential_features.py
zhoutianxing7B: Done. I grouped the VLM Trend reward utilities, shared data helpers, and pipeline config under examples/reward/vlm_trend/. Runner-owned SFT and embodiment configs remain in their standard config directories.

docs/source-zh/rst_source/extending/reward_model.rst
zhoutianxing7B: Four GPUs are no longer required. Feature extraction now uses RLinf’s cluster placement and defaults to all available accelerators. It can run on one GPU by setting feature_extractor: 0.

docs/source-zh/rst_source/extending/reward_model.rst
zhoutianxing7B: Done. The nested shell loop has been replaced by a single Python entrypoint backed by an RLinf WorkerGroup. Worker rank and world size are derived automatically from the cluster placement.

docs/source-zh/rst_source/extending/reward_model.rst
```diff
@@ -236,6 +239,87 @@ RLinf 支持两条 reward 训练路径。``examples/reward/run_reward_training.s
...
+     --raw-data-path /path/to/vlm_trend_uniform_collection/step20 \
+     --value-checkpoint /path/to/vlm_trend_success_potential/teacher/best.pt \
+     --output-dir /path/to/vlm_trend_success_potential/potential_data
+
+这些命令会保留与在线采样一致的 ``0``/``1`` terminal-success 窗口，不执行
+balanced sampling。两条分支都在首次成功 transition 截断，与在线环境的终止边界一致。
+Potential 预处理使用 teacher 生成绝对 potential 数字，以及
+``up``/``same``/``down`` progress pair。
+
+两个薄配置继承现有 Trend 配方，路径已经在 YAML 中配置，无需再导出路径环境变量：
```
Lin-xs: Please re-check the Chinese version and replace leftover machine-translated phrasing.
"薄配置" -> “配置”
“配方” -> “配置”

Please delete all description about the deleted legacy implementation, such as "路径已经在 YAML 中配置，无需再导出路径环境变量", "不要求必须使用四张 GPU", "替代原来的嵌套 shell 循环", etc.

examples/reward/vlm_trend/config/pipeline.yaml
```diff
@@ -0,0 +1,108 @@
...
+  num_bins: 10
+  video_fps: 24.0
+  max_samples: null
+
+auxiliary:
+  stage: teacher  # teacher or scalar_head
+
+  teacher:
+    raw_data_paths:
+      - ${paths.raw_data_root}/step0
```
Lin-xs, line 56: Please note in the docs that users should set `raw_data_paths` to their actual paths.

docs/source-zh/rst_source/extending/reward_model.rst
zhoutianxing7B: Addressed in c3efbe3d: rechecked the Chinese reward-model section, replaced the leftover machine-translated wording, and removed descriptions that referred to deleted legacy implementation details.

examples/reward/vlm_trend/config/pipeline.yaml
zhoutianxing7B: Addressed in c3efbe3d: the docs now tell users to replace auxiliary.teacher.raw_data_paths with their actual collected checkpoint directories, and pipeline.yaml includes the same reminder next to the example list.

Lin-xs (approved):

andylin-hao (approved):

guozhen1997 (approved):

PR #1292, opened 2026-06-17
feat(embodied): add FastWAM world-action model (LIBERO eval + SFT)

wzii:
Wrap the upstream FastWAM (Wan2.2 video-diffusion world-action model) behind RLinf's embodied interfaces for LIBERO eval and SFT (predict_action_batch -> infer_action, sft_forward -> training_loss).

### Description
This PR integrates FastWAM into RLinf.

### How has this been tested?
Tested with LIBERO-spatial evaluation (SR meets the paper's results) and a smoke run of sft

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (Document-only update)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1398, opened 2026-07-22
feat(embodied): add FastWAM evaluation and FSDP SFT support

ligediaomao:
### Description

This PR adds FastWAM support to RLinf, including:

- Batched FastWAM inference for LIBERO and LIBERO-Plus evaluation.
- FastWAM supervised fine-tuning through the RLinf FSDP2 SFT pipeline.
- FastWAM-specific installation dependencies and Docker integration.
- English and Chinese documentation.
- Mock unit tests for preprocessing, normalization, batched inference, and SFT dispatch.
- CI smoke coverage for real checkpoint loading, two-sample GPU batch inference, and one FSDP forward/backward optimizer step.

### Motivation and Context

This PR supersedes [#1292](https://github.com/RLinf/RLinf/pull/1292) and carries forward the original FastWAM integration while addressing the previous review comments.

Compared with #1292, this PR:

- removes duplicated evaluation configurations,
- aligns FastWAM prompts and evaluation settings with upstream,
- fixes batched inference and configuration handling,
- corrects the FSDP2 wrap policy,
- adds FastWAM SFT support,
- adds mock unit tests and GPU CI smoke jobs,
- and validates the RLinf-native evaluation pipeline against upstream FastWAM on the standard LIBERO suites.

### LIBERO Evaluation Results

In addition to CI smoke tests, the RLinf-native FastWAM evaluation pipeline was evaluated on the standard LIBERO suites after aligning the prompts, reset states, gripper wait steps, model/processor setup, and evaluation configuration with upstream FastWAM.

| Suite | Upstream FastWAM | RLinf FastWAM | Difference |
| --- | ---: | ---: | ---: |
| LIBERO-Spatial | 98.2% | 97.2% | -1.0 pp |
| LIBERO-Object | 100.0% | 100.0% | 0.0 pp |
| LIBERO-Goal | 97.0% | 97.6% | +0.6 pp |
| LIBERO-Long / LIBERO-10 | 95.2% | 93.6% | -1.6 pp |
| **Average** | **97.6%** | **97.1%** | **-0.5 pp** |

#### Training Curves
PixPin_2026-08-02_14-37-35

#### Eval Results
FastWam official result shows as below photo.
PixPin_2026-08-02_14-31-24

Our results evaluaed from RLinf:
1. Libero10: 93.6%
PixPin_2026-08-02_14-33-46

2. Libero object: 100%
PixPin_2026-08-02_14-36-34

3. Libero spatial: 97.2%
PixPin_2026-08-02_14-35-09

4. Libero goal: 97.6%
PixPin_2026-08-02_14-35-55

The RLinf-native evaluation reproduces upstream FastWAM performance closely across all four suites, with an average difference of 0.5 percentage points.

The full LIBERO benchmark results above were obtained from a dedicated GPU evaluation run and are separate from the CI jobs, which are intentionally smoke-oriented.

### How has this been tested?

- Mock FastWAM unit tests cover:
  - preprocessing,
  - action normalization / denormalization,
  - batched inference,
  - and SFT dispatch,
  without requiring the full FastWAM checkpoint.

- Local checks cover:
  - workflow YAML parsing,
  - `git diff --check`,
  - CI install-job resolution,
  - and e2e configuration invariants.

- GPU CI smoke jobs cover:
  - loading a real FastWAM checkpoint,
  - two-sample batched GPU inference,
  - and one FSDP forward/backward optimizer step.

- A dedicated GPU run was used to evaluate all four standard LIBERO suites, producing a 97.1% average success rate compared with 97.6% reported by upstream FastWAM.

The GPU CI jobs require the `run-ci` label and the following shared runner assets:

```text
/workspace/dataset/FastWAM/checkpoints/libero_uncond_2cam224.pt
/workspace/dataset/FastWAM/checkpoints/libero_uncond_2cam224_dataset_stats.json
/workspace/dataset/FastWAM/models/Wan-AI/Wan2.2-TI2V-5B
/workspace/dataset/FastWAM/models/Wan-AI/Wan2.1-T2V-1.3B
/workspace/dataset/FastWAM/text_embeds_cache/libero
/workspace/dataset/data/sft-ci-data/fastwam-libero
```
Maintainers: please help confirm or stage these large shared assets before enabling the GPU CI jobs.

### Additional information

The upstream FastWAM dependency is pinned to commit:

45d8e1458921d83f8ad6cf9ce993d371208dabd0

The full LIBERO evaluation results were obtained from a dedicated GPU run using the RLinf-native FastWAM evaluation pipeline. CI remains smoke-oriented and is intended to validate checkpoint loading, batched GPU inference, and the FSDP training path rather than reproduce the full benchmark.

### Types of changes

- [x] Documentation update
- [ ] Breaking change

### Checklist

- [x] Relevant local tests and evaluation runs passed.


wzii:
Thank you very much for moving this forward. Previously I can't reproduce fastwam's eval results as there seems to be some misalignment on the LIBERO tasks (e.g. task prompt). Does this commit fix that?

ligediaomao:
> Thank you very much for moving this forward. Previously I can't reproduce fastwam's eval results as there seems to be some misalignment on the LIBERO tasks (e.g. task prompt). Does this commit fix that?

The task prompt is now aligned with upstream FastWAM. However, I haven’t reproduced the full LIBERO evaluation results yet—the new CI is only a smoke test. I’ll continue working on reproducing the results and aligning the evaluation setup.

zoeyuchao:
@wzii Hi, we fixed the performance mismatch issue in this pr, could you double-check?

wzii:
> @wzii Hi, we fixed the performance mismatch issue in this pr, could you double-check?

Hi, thank you very much for the work. Could you please share the results?

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -88,6 +95,87 @@ def init_worker(self):
...
+            total_steps = steps_per_epoch * max_epochs
+            if max_steps >= 0:
+                total_steps = min(total_steps, max_steps)
+        elif max_steps >= 0:
+            total_steps = max_steps
+        else:
+            total_steps = steps_per_epoch
+
+        total_steps = max(int(total_steps), 1)
+        self._cfg.optim.total_training_steps = total_steps
```
F9rozen: This unconditionally overwrites `total_training_steps` for every SFT model, not just FastWAM. Existing recipes that set an explicit `actor.optim.total_training_steps` will silently get a different LR warmup/cosine horizon.

Please gate this to FastWAM only (or only when `total_training_steps` is null), or move the override into a FastWAM-specific SFT worker subclass.

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -101,7 +189,21 @@ def set_global_step(self, global_step):
...
+                    math.ceil(
+                        dataset_size / (self.micro_batch_size * self._world_size)
+                    ),
+                    1,
+                )
+                return max(
+                    math.ceil(global_micro_steps / self.gradient_accumulation),
+                    1,
+                )
+            return max(1, math.ceil(len(self.data_loader) / self.gradient_accumulation))
```
F9rozen: Non-FastWAM path changed from floor division (`len(data_loader) // grad_accum`) to `math.ceil(...)`. That silently changes epoch length / scheduler horizon for all existing SFT recipes.

Please keep the previous `//` semantics for non-FastWAM models in this PR. If the ceil change is intentional, land it separately with an explicit callout.

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -88,6 +95,87 @@ def init_worker(self):
...
+            # Match FastWAM's set_global_seed(seed) after model construction:
+            # each process uses seed + its distributed rank.
+            from fastwam.utils.pytorch_utils import set_global_seed
+
+            set_global_seed(int(self.cfg.actor.seed))
+            self._set_training_mode()
+            if self.data_loader is not None:
+                self.data_iter = iter(self.data_loader)
+
+    def _set_training_mode(self) -> None:
```
F9rozen: `_set_training_mode()` is FastWAM-specific but is now on the shared base class and called from `run_training()` for all models. Please keep the FastWAM `model.eval()` + `dit.train()` split in a FastWAM subclass / override.

evaluations/libero/libero_spatial_fastwam_eval.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+  eval:
+    # Minimal eval: a handful of parallel envs, one rollout epoch. FastWAM runs a
+    # 5B+1B flow-matching inference every `num_action_chunks` sim steps, so keep
+    # the env count / horizon small for a quick real-world sanity number.
+    rollout_epoch: 1
+    total_num_envs: 4
+    auto_reset: True
+    ignore_terminations: True
+    max_episode_steps: 240
```
F9rozen: `libero_fastwam_full_eval.yaml` sets `num_steps_wait: 30`, but this spatial eval still relies on LiberoEnv's default of 15. The PR notes that FastWAM's official evaluator uses 30 open-gripper no-ops.

Please set `num_steps_wait: 30` here (and in the e2e FastWAM eval yaml) so the smoke path matches upstream start-of-episode setup.

tests/e2e_tests/embodied/libero_spatial_eval_fastwam.yaml
```diff
@@ -0,0 +1,78 @@
...
+  ckpt_path: null
+
+env:
+  group_name: "EnvGroup"
+  eval:
+    rollout_epoch: 1
+    total_num_envs: 2
+    auto_reset: true
+    ignore_terminations: true
+    max_episode_steps: 10
```
F9rozen: Same as the spatial eval config: please add `num_steps_wait: 30` so CI smoke does not diverge from the official FastWAM wait protocol (even if the episode horizon is short).

requirements/embodied/models/fastwam.txt
```diff
@@ -0,0 +1,31 @@
+# FastWAM runtime dependencies from upstream commit
+# 45d8e1458921d83f8ad6cf9ce993d371208dabd0. Torch and torchvision are
+# intentionally omitted so requirements/install.sh keeps RLinf's platform stack.
+accelerate==1.12.0
+av==16.0.1
+boto3==1.35.99
+datasets==3.6.0
+deepspeed==0.18.5
```
F9rozen: Why pin `deepspeed` when the RLinf path uses FSDP2? If this is only carried over from upstream and unused, please drop it to reduce dependency conflicts. If some FastWAM script still needs it, a one-line comment here would help.

F9rozen:
Per the thread: prompts are aligned, but full LIBERO numbers are not reproduced yet and CI is smoke-only. Before merging as “evaluation support”, please post at least a spatial/suite comparison against upstream FastWAM, or explicitly mark this as experimental/smoke-only in the PR summary.

ligediaomao:
> Thank you very much for moving this forward. Previously I can't reproduce fastwam's eval results as there seems to be some misalignment on the LIBERO tasks (e.g. task prompt). Does this commit fix that?

Thanks! We have now completed the RLinf-native LIBERO evaluation after aligning the evaluation setup with upstream FastWAM.

Results:

| Suite | Upstream FastWAM | RLinf FastWAM |
| --- | ---: | ---: |
| LIBERO-Spatial | 98.2% | 97.2% |
| LIBERO-Object | 100.0% | 100.0% |
| LIBERO-Goal | 97.0% | 97.6% |
| LIBERO-Long / LIBERO-10 | 95.2% | 93.6% |
| **Average** | **97.6%** | **97.1%** |

ligediaomao:
> Per the thread: prompts are aligned, but full LIBERO numbers are not reproduced yet and CI is smoke-only. Before merging as “evaluation support”, please post at least a spatial/suite comparison against upstream FastWAM, or explicitly mark this as experimental/smoke-only in the PR summary.

Thanks for pointing this out. We have now completed a dedicated LIBERO evaluation and added the suite-level comparison against upstream FastWAM to the PR summary.

RLinf FastWAM reproduces an average success rate of 97.1%, compared with 97.6% upstream:

- Spatial: 97.2% vs 98.2%
- Object: 100.0% vs 100.0%
- Goal: 97.6% vs 97.0%
- Long / LIBERO-10: 93.6% vs 95.2%

I also clarified in the PR summary that CI remains smoke-oriented; the full LIBERO results above come from a separate dedicated GPU evaluation run.

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > This unconditionally overwrites `total_training_steps` for every SFT model, not just FastWAM. Existing recipes that set an explicit `actor.optim.total_training_steps` will silently get a different LR warmup/cosine horizon.
> 
> Please gate this to FastWAM only (or only when `total_training_steps` is null), or move the override into a FastWAM-specific SFT worker subclass.

Thanks for pointing this out. The scheduler horizon derivation is now gated to FastWAM only.

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > Non-FastWAM path changed from floor division (`len(data_loader) // grad_accum`) to `math.ceil(...)`. That silently changes epoch length / scheduler horizon for all existing SFT recipes.
> 
> Please keep the previous `//` semantics for non-FastWAM models in this PR. If the ceil change is intentional, land it separately with an explicit callout.

Thanks for catching this. I restored the floor-division semantics for non-FastWAM models.

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > `_set_training_mode()` is FastWAM-specific but is now on the shared base class and called from `run_training()` for all models. Please keep the FastWAM `model.eval()` + `dit.train()` split in a FastWAM subclass / override.

Thanks for the suggestion. I moved the FastWAM-specific mode split to the VLA worker override.

evaluations/libero/libero_spatial_fastwam_eval.yaml
ligediaomao: > `libero_fastwam_full_eval.yaml` sets `num_steps_wait: 30`, but this spatial eval still relies on LiberoEnv's default of 15. The PR notes that FastWAM's official evaluator uses 30 open-gripper no-ops.
> 
> Please set `num_steps_wait: 30` here (and in the e2e FastWAM eval yaml) so the smoke path matches upstream start-of-episode setup.

Good catch, thanks. I added num_steps_wait: 30 to the spatial eval config.

tests/e2e_tests/embodied/libero_spatial_eval_fastwam.yaml
ligediaomao: > Same as the spatial eval config: please add `num_steps_wait: 30` so CI smoke does not diverge from the official FastWAM wait protocol (even if the episode horizon is short).

Thanks for noting this. I also added num_steps_wait: 30 to the E2E eval config.

requirements/embodied/models/fastwam.txt
ligediaomao: > Why pin `deepspeed` when the RLinf path uses FSDP2? If this is only carried over from upstream and unused, please drop it to reduce dependency conflicts. If some FastWAM script still needs it, a one-line comment here would help.

Thanks for pointing this out. I removed the unused DeepSpeed dependency since the RLinf path uses FSDP2.

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -88,6 +95,77 @@ def init_worker(self):
...
```
thereAreDemonsNearby: Now this method seems unnecessary. Using self.model.train() is simpler.

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -88,6 +95,77 @@ def init_worker(self):
...
+        """Set the model to training mode."""
+        self.model.train()
+
+    def setup_model_and_optimizer(self) -> None:
+        """Build the optimizer schedule from the runner's actual SFT horizon.
```
thereAreDemonsNearby: These modifications should be considered for moving to a more common location, such as the FSDP model manager. The logic could be: if the user does not specify cfg.optim.total_training_steps, then infer it using your method here.

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -88,6 +95,77 @@ def init_worker(self):
...
+            total_steps,
+        )
+        super().setup_model_and_optimizer()
+        self._set_training_mode()
+
+        # FSDPModelManager pre-warms AdamW state with a zero-lr optimizer step
+        # so checkpoint loading has materialized state entries. Official
+        # FastWAM starts from an untouched optimizer, where the first real
+        # update has AdamW step=1. Restore that state only for fresh FastWAM
+        # runs; a resumed run must keep the checkpoint's optimizer step.
```
thereAreDemonsNearby: If this modification addresses a common issue, it would be better to move it to a common location as well.

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -67,7 +71,10 @@ def __init__(self, cfg: DictConfig):
             self.data_loader, self.data_config = self.build_dataloader(
                 self.cfg.data.train_data_paths, eval_dataset=False
             )
-            self.data_iter = iter(self.data_loader)
+            # FastWAM seeds the process after model construction and before the
+            # first iterator is created. Defer iterator construction until
+            # init_worker() so its worker/RNG ordering matches upstream.
+            self.data_iter = None if self._is_fastwam else iter(self.data_loader)
```
thereAreDemonsNearby: If your purpose for doing this (adding special handling code) is to make the dataloader fully align with the upstream random seed in order to match the curve, I don't think it's worth the effort. Generally, changing the data random seed shouldn't drastically affect the training results.

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -88,6 +95,77 @@ def init_worker(self):
             self.offload_param_and_grad()
             self.offload_optimizer()
 
+        if self._is_fastwam:
+            # Match FastWAM's set_global_seed(seed) after model construction:
```
thereAreDemonsNearby: If your purpose for doing this (adding special handling code) is to make the dataloader fully align with the upstream random seed in order to match the curve, I don't think it's worth the effort. Generally, changing the data random seed shouldn't drastically affect the training results. I suggest you remove this modification and see if it noticeably affects the training results.

thereAreDemonsNearby: If your concern is that set_global_seed is not being called for the dataset, I suggest calling it during dataset initialization (which may require a thin wrapper around the dataset).

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -101,6 +179,20 @@ def set_global_step(self, global_step):
 
     def get_max_steps_per_epoch(self):
         if self.data_loader is not None:
+            if self._is_fastwam:
+                # Match Wan22Trainer._estimate_total_train_steps exactly:
+                # first count global micro-batches, then apply accumulation.
+                dataset_size = len(self.data_loader.dataset)
```
thereAreDemonsNearby: Here, isn't len(dataset) / (mb * ws) equal to len(dataloader)?

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -101,6 +179,20 @@ def set_global_step(self, global_step):
...
+                global_micro_steps = max(
+                    math.ceil(
+                        dataset_size / (self.micro_batch_size * self._world_size)
+                    ),
+                    1,
+                )
+                return max(
+                    math.ceil(global_micro_steps / self.gradient_accumulation),
```
thereAreDemonsNearby: Here, you're using ceiling division instead of the existing floor division in rlinf because the Fast WAM dataset doesn't drop the last batch, right? I suggest adding attribute to the dataloader (whether to drop last) to determine which division method to use, rather than judging by the model name.

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -29,6 +29,22 @@ class FSDPVlaSftWorker(FSDPSftWorker):
     def __init__(self, cfg: DictConfig):
         super().__init__(cfg)
 
+    def _set_training_mode(self) -> None:
+        """Apply the official FastWAM train/eval mode split."""
```
thereAreDemonsNearby: You forgot to remove this

wzii:
> > Thank you very much for moving this forward. Previously I can't reproduce fastwam's eval results as there seems to be some misalignment on the LIBERO tasks (e.g. task prompt). Does this commit fix that?
> 
> Thanks! We have now completed the RLinf-native LIBERO evaluation after aligning the evaluation setup with upstream FastWAM.
> 
> Results:
> 
> Suite	Upstream FastWAM	RLinf FastWAM
> LIBERO-Spatial	98.2%	97.2%
> LIBERO-Object	100.0%	100.0%
> LIBERO-Goal	97.0%	97.6%
> LIBERO-Long / LIBERO-10	95.2%	93.6%
> **Average**	**97.6%**	**97.1%**

Thank you very much. The numbers look good to me.

F9rozen (approved):

F9rozen (changes requested):
Simplify fsdp_sft_worker,Delete unnecessary changes

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > Now this method seems unnecessary. Using self.model.train() is simpler.

Agreed. I removed _set_training_mode() .Thank you.

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > These modifications should be considered for moving to a more common location, such as the FSDP model manager. The logic could be: if the user does not specify cfg.optim.total_training_steps, then infer it using your method here.

Thanks for the suggestion. I moved the fallback inference to FSDPModelManager. An explicit total_training_steps value is preserved, while SFT infers it when unset.

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > If this modification addresses a common issue, it would be better to move it to a common location as well.

Good point. warmup_optimizer_state() already handles the step reset in the common path, so I removed the duplicate FastWAM-specific logic.

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > If your purpose for doing this (adding special handling code) is to make the dataloader fully align with the upstream random seed in order to match the curve, I don't think it's worth the effort. Generally, changing the data random seed shouldn't drastically affect the training results.

Agreed. I removed the FastWAM-specific global seed handling and restored normal DataLoader iterator initialization.

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > If your concern is that set_global_seed is not being called for the dataset, I suggest calling it during dataset initialization (which may require a thin wrapper around the dataset).

Thanks. The FastWAM sampler already uses a dedicated generator seeded with cfg.actor.seed + epoch, so no additional dataset wrapper is needed.

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > Here, isn't len(dataset) / (mb * ws) equal to len(dataloader)?

Yes, that is equivalent with the current sampler. I simplified the calculation to use len(self.data_loader) directly.

rlinf/workers/sft/fsdp_sft_worker.py
ligediaomao: > Here, you're using ceiling division instead of the existing floor division in rlinf because the Fast WAM dataset doesn't drop the last batch, right? I suggest adding attribute to the dataloader (whether to drop last) to determine which division method to use, rather than judging by the model name.

Agreed. The calculation now uses data_loader.drop_last to choose floor or ceiling division, without checking the model name.

rlinf/workers/sft/fsdp_vla_sft_worker.py
ligediaomao: > You forgot to remove this

Thanks for catching this. I removed the leftover _set_training_mode() override from FSDPVlaSftWorker.

ligediaomao:
> Simplify fsdp_sft_worker,Delete unnecessary changes

Thanks for the review. I simplified fsdp_sft_worker.py and removed the unnecessary FastWAM-specific changes as suggested. The updates have been pushed in commit d7c0b6d3.

rlinf/envs/libero/libero_env.py
F9rozen: _eval_active changes LiberoEnv.step() for every LIBERO eval, not just FastWAM. Training rollouts are untouched and the all-active path is the old step, which is good. Please confirm other models (openpi / dreamzero / …) keep the same success rate after init-state exhaustion, and that a fully inactive batch cannot spin forever. This is the largest shared-path behavior change

docs/source-en/rst_source/examples/embodied/sft_fastwam.rst
```diff
@@ -0,0 +1,415 @@
...
+prepares ActionDiT, downloads and extracts all four LIBERO suites, and
+precomputes T5 text embeddings. Activate the environment, then paste it into a
+terminal from the repository root; use tmux for long downloads and training:
+
+.. code-block:: bash
+
+   #!/usr/bin/env bash
+   set -euo pipefail
+
+   SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
```
F9rozen: The prose says paste from the repo root, but the snippet sets SCRIPT_DIR from BASH_SOURCE[0] and REPO_PATH=$SCRIPT_DIR/../... In an interactive paste BASH_SOURCE is empty, so REPO_PATH walks two levels above the repo. Use REPO_PATH="$(git rev-parse --show-toplevel)" or "$(pwd)" and drop SCRIPT_DIR/../...

docs/source-en/rst_source/examples/embodied/sft_fastwam.rst
```diff
@@ -0,0 +1,415 @@
...
+   message does not indicate a FastWAM configuration error; do not switch to
+   fp32 merely to silence it.
+
+Use only ``model_path`` for the FastWAM checkpoint. ``checkpoint_path`` is not a
+supported alias.
+
+Evaluate
+--------
+
+Use evaluations/libero/libero_fastwam_full_eval.yaml to evaluate an SFT
```
F9rozen: The section first says use libero_fastwam_full_eval.yaml for an SFT ckpt, then says the single spatial YAML replaced everything, then all commands use libero_spatial_fastwam_eval. Split smoke (spatial YAML) vs full-suite (full_eval YAML) and show run_eval.sh libero libero_fastwam_full_eval.

docs/source-en/rst_source/examples/embodied/sft_fastwam.rst
```diff
@@ -0,0 +1,415 @@
...
+   dataset_stats_path: /workspace/checkpoints/fastwam/libero_uncond_2cam224_dataset_stats.json
+
+FastWAM and RLinf Configuration
+-------------------------------
+
+The standard evaluation YAML reads examples/embodiment/config/model/fastwam.yaml.
+Before use, replace the model_path and dataset_stats_path placeholders in that
+file with local model files:
+
+.. code-block:: yaml
```
F9rozen: Asking users to edit placeholders in examples/embodiment/config/model/fastwam.yaml mutates the shared rollout preset. full_eval.yaml already has its own model_path placeholders (L69–L70).   full eval → the two fields in full_eval.yaml.

docs/source-zh/rst_source/evaluations/get_started/installation.rst
```diff
@@ -18,7 +18,7 @@
      - 推荐 ``--model``
      - 推荐 ``--env``
    * - LIBERO
-     - ``openpi`` / ``openvla-oft`` / ``starvla`` / ``dreamzero``
+     - ``openpi`` / ``openvla-oft`` / ``starvla`` / ``dreamzero`` / ``fastwam``
```
F9rozen: FastWAM is grouped with openpi under “maniskill_libero or libero”, but install_fastwam_model only accepts libero / liberoplus. --env maniskill_libero will fail.

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
```diff
@@ -267,10 +267,36 @@ def _optimize_with_liger_kernel(self, model: torch.nn.Module) -> None:
...
         Setup model, lr_scheduler, optimizer and grad_scaler.
         """
+        if self._cfg.optim.get("total_training_steps") is None:
+            total_training_steps = self.infer_total_training_steps()
+            self._cfg.optim.total_training_steps = total_training_steps
+            self._logger.info(
+                "[FSDP] Inferred optim.total_training_steps=%d",
+                total_training_steps,
+            )
+
```
thereAreDemonsNearby: These lines of code should be put in fsdp_sft_worker.py (write a small function and maybe put it inside init_worker) since they're only related to SFT.  fsdp model manager is also used by RL.

Iron-Wph:
### Review follow-up

The latest commits address the remaining review points:

- Removed `stop_when_eval_exhausted` and `_eval_active`; the shared LIBERO environment behavior is unchanged. Full evaluation now uses 4 workers and 20 environments, covering exactly 500 trajectories without padding or duplication.
- Restored the generic SFT floor division. FastWAM infers `optim.total_training_steps` only when it is unset; explicit values and other models are unaffected.
- Aligned the evaluation base seed to 42 and updated the smoke/full-evaluation and throughput documentation.

Using the [official released checkpoint and dataset statistics](https://huggingface.co/yuanty/fastwam), seed 42, inference batch size 5, and 500 trajectories per suite:

| Suite | Official FastWAM | RLinf | Successful trajectories | Difference |
| --- | ---: | ---: | ---: | ---: |
| LIBERO-Spatial | 98.2% | 96.2% | 481/500 | -2.0 pp |
| LIBERO-Object | 100.0% | 99.4% | 497/500 | -0.6 pp |
| LIBERO-Goal | 97.0% | 96.6% | 483/500 | -0.4 pp |
| LIBERO-Long | 95.2% | 94.6% | 473/500 | -0.6 pp |
| **Average** | **97.6%** | **96.7%** | **1934/2000** | **-0.9 pp** |

Validation: 9 relevant unit tests passed; `ruff check`, `ruff format --check`, and `git diff --check` also passed.

rlinf/envs/libero/libero_env.py
```diff
@@ -839,7 +839,9 @@ def reset(
             reset_state_ids = self._get_random_reset_state_ids(num_reset_states)
 
         self._reconfigure(reset_state_ids, env_idx)
-        for _ in range(15):
+        # FastWAM's official evaluator waits for 30 open-gripper no-op steps.
+        num_steps_wait = int(self.cfg.get("num_steps_wait", 0))
```
FxxxxU: The default RLinf configuration requires waiting for 15 iterations. Why does it start from 0 when I don't set 'num_steps_wait' ?

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -28,6 +29,41 @@ class FSDPVlaSftWorker(FSDPSftWorker):
     def __init__(self, cfg: DictConfig):
         super().__init__(cfg)
 
+    def init_worker(self):
+        self._set_fastwam_total_training_steps_if_missing()
+        super().init_worker()
+
+    def _set_fastwam_total_training_steps_if_missing(self) -> None:
```
thereAreDemonsNearby: why change this method from  general  one to one limited to fastwam only?

tests/e2e_tests/embodied/maniskill_sac_mlp_async_decoupled_with_delay.yaml
```diff
@@ -62,7 +62,7 @@ algorithm:
   adv_type: embodied_sac
   loss_type: embodied_sac
   loss_agg_func: "token-mean"
-
```
FxxxxU, line 65: Revert the unnecessary changes made here.

tests/e2e_tests/embodied/libero_spatial_eval_fastwam.yaml
```diff
@@ -0,0 +1,79 @@
+# Copyright 2026 The RLinf Authors.
```
FxxxxU: This “open source” tag should not be included in the YAML file.

tests/e2e_tests/embodied/libero_spatial_eval_fastwam.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+rollout:
+  group_name: "RolloutGroup"
+  generation_backend: "huggingface"
+  enable_offload: false
+  pipeline_stage_num: 1
+
+  model:
+    model_type: "fastwam"
+    model_path: /workspace/dataset/FastWAM/checkpoints/libero_uncond_2cam224.pt
```
FxxxxU: Change the file path to 'path/to/xxx'

FxxxxU: Add appropriate comments, and include any HTTP links for open-source downloads, if available.

tests/e2e_tests/embodied/libero_spatial_eval_fastwam.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+cluster:
+  num_nodes: 1
+  component_placement:
+    env,rollout: 0
+
+runner:
+  task_type: embodied_eval
+  logger:
+    log_path: "/workspace/results/fastwam_eval"
```
FxxxxU: Replace with "../results"

tests/e2e_tests/sft/libero_sft_fastwam.yaml
```diff
@@ -0,0 +1,99 @@
+# Copyright 2026 The RLinf Authors.
```
FxxxxU: Same as above.

tests/e2e_tests/sft/libero_sft_fastwam.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+cluster:
+  num_nodes: 1
+  component_placement:
+    actor: 0-1
+
+runner:
+  task_type: sft
+  logger:
+    log_path: "/workspace/results/fastwam_sft"
```
FxxxxU: Same as above.

tests/e2e_tests/sft/libero_sft_fastwam.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  max_epochs: -1
+  max_steps: 1
+  val_check_interval: -1
+  save_interval: -1
+  log_interval: 1
+  resume_dir: null
+
+data:
+  train_data_paths: /workspace/dataset/data/sft-ci-data/fastwam-libero
```
FxxxxU: Same as above.

tests/e2e_tests/sft/libero_sft_fastwam.yaml
```diff
@@ -0,0 +1,99 @@
...
+actor:
+  group_name: "ActorGroup"
+  training_backend: "fsdp"
+  micro_batch_size: 1
+  global_batch_size: 2
+  seed: 42
+
+  model:
+    model_type: "fastwam"
+    model_path: /workspace/dataset/FastWAM/checkpoints/libero_uncond_2cam224.pt
```
FxxxxU: Same as above.

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -28,6 +29,41 @@ class FSDPVlaSftWorker(FSDPSftWorker):
     def __init__(self, cfg: DictConfig):
         super().__init__(cfg)
 
+    def init_worker(self):
+        self._set_fastwam_total_training_steps_if_missing()
```
FxxxxU: This approach to making changes is difficult to scale and also reduces the code's readability.
I have a suggested modification:
1. Here, you're checking `model_type == SupportedModel.FASTWAM` in a single line, and then setting the necessary parameters within that line.
2. You can move the necessary parameter settings into a function named `fastwam_sft_init()` and place it in the corresponding model folder to handle initialization, ensuring the readability of the current `FSDPVlaSftWorker`.
3. You can move the function `fastwam_sft_init()` to `models/embodiement/fastwam/__init__.py` file.

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -162,6 +204,6 @@ def get_max_steps_per_epoch(self):
                 )
 
                 num_batches = get_official_openpi_sft_num_batches(self.data_loader)
-        else:
-            return super().get_max_steps_per_epoch()
-        return max(1, num_batches // self.gradient_accumulation)
+            return max(1, num_batches // self.gradient_accumulation)
```
FxxxxU: This is really bothering me. 
Either I should add an ‘else:’ at the end, or I shouldn't make any changes.

rlinf/models/embodiment/fastwam/fastwam_policy.py
```diff
@@ -0,0 +1,584 @@
...
+
+        from fastwam.utils.video_io import save_mp4
+
+        out_dir = self.policy_cfg.future_video_dir
+        os.makedirs(out_dir, exist_ok=True)
+        path = os.path.join(out_dir, f"future_{self._video_saves:04d}.mp4")
+        save_mp4(frames, path, fps=8)
+        self._video_saves += 1
+
+    # ------------------------------------------------------------------ utils
```
FxxxxU: delete.

rlinf/models/embodiment/fastwam/fastwam_policy.py
```diff
@@ -0,0 +1,584 @@
...
+            )
+
+        return {
+            "action": latents_action.detach().to(
+                device="cpu",
+                dtype=torch.float32,
+            )
+        }
+
+    # ------------------------------------------------------------------ rollout
```
FxxxxU: delete.

rlinf/models/embodiment/fastwam/fastwam_policy.py
```diff
@@ -0,0 +1,584 @@
...
+        # LIBERO's -1=open, +1=close convention before execution.
+        actions[..., -1] = actions[..., -1] * 2 - 1
+        actions = _invert_gripper_action(actions)
+        if cfg.binarize_gripper:
+            actions[..., -1] = np.sign(actions[..., -1])
+
+        actions = actions[:, : cfg.num_action_chunks].astype(np.float32)
+        return actions, {}
+
+    # ------------------------------------------------------------------ SFT
```
FxxxxU: delete.

docs/source-zh/rst_source/examples/embodied/sft_fastwam.rst
```diff
@@ -0,0 +1,395 @@
...
+   这条建议不适用于 FastWAM：其官方 SFT 路径就是 bf16 模型参数加 bf16 autocast，
+   且 FSDP2 不应再额外进行参数 dtype cast。看到该通用提示不表示 FastWAM 配置错误，
+   不需要为消除提示而改为 fp32。
+
+FastWAM 检查点只使用 ``model_path``，不支持 ``checkpoint_path`` 别名。
+
+评测
+----
+
+完整 SFT checkpoint 的评测使用 evaluations/libero/libero_fastwam_full_eval.yaml。
```
F9rozen: splits smoke (libero_spatial_fastwam_eval) vs full (libero_fastwam_full_eval)

docs/source-zh/rst_source/examples/embodied/sft_fastwam.rst
```diff
@@ -0,0 +1,395 @@
...
+下载并解压四套 LIBERO 数据，并预计算 T5 text embedding。先激活环境，在仓库根目录
+将代码逐行粘贴到终端执行；长时间下载和训练建议在 tmux 中进行：
+
+.. code-block:: bash
+
+   #!/usr/bin/env bash
+   set -euo pipefail
+
+   SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+   REPO_PATH="$(cd -- "${SCRIPT_DIR}/../.." && pwd)"
```
F9rozen: uses REPO_PATH="$(git rev-parse --show-toplevel)"
copy the EN structure.

thereAreDemonsNearby (approved):

rlinf/workers/sft/fsdp_vla_sft_worker.py
Iron-Wph: Addressed. I removed the FastWAM-specific scheduler initialization from
`FSDPVlaSftWorker`. Training-step inference is now a generic opt-in behavior in
the base SFT worker and only runs when `optim.total_training_steps` is explicitly
set to `null`. Missing keys and explicit numeric values remain unchanged.

rlinf/workers/sft/fsdp_vla_sft_worker.py
Iron-Wph: The method now has an explicit final `else` that delegates to the
parent implementation. FastWAM uses this generic path.

rlinf/models/embodiment/fastwam/fastwam_policy.py
Iron-Wph: fixed

rlinf/models/embodiment/fastwam/fastwam_policy.py
Iron-Wph: fixed

rlinf/models/embodiment/fastwam/fastwam_policy.py
Iron-Wph: fixed

rlinf/envs/libero/libero_env.py
Iron-Wph: Fixed. The default remains 15 to preserve the existing LIBERO behavior for
other models.

docs/source-zh/rst_source/examples/embodied/sft_fastwam.rst
Iron-Wph: Addressed by replacing the ambiguous `full_eval` setup with four suite-specific
evaluation configs: Spatial, Object, Goal, and Long. Each config performs the
complete 500-state evaluation. CI reuses the Spatial config with explicit
small-scale overrides for smoke testing.

docs/source-zh/rst_source/examples/embodied/sft_fastwam.rst
Iron-Wph: Addressed. The unnecessary path-discovery snippet was removed, and the commands
now assume they are run from the repository root. The English and Chinese
guides use the same structure.

rlinf/workers/sft/fsdp_vla_sft_worker.py
Iron-Wph: I removed the FastWAM-specific
`_set_fastwam_total_training_steps_if_missing()` method from
`FSDPVlaSftWorker`.

The training-step inference is now implemented generically in the base
`FSDPSftWorker`. It is only enabled when a configuration explicitly sets
`optim.total_training_steps: null`; missing keys and explicit numeric values
remain unchanged. FastWAM opts into this behavior through its SFT YAML.

Iron-Wph:
### Intentional differences from upstream FastWAM

- Upstream uses Accelerate/DeepSpeed, while RLinf uses its native FSDP2
  backend. This changes parameter sharding, communication, and checkpointing,
  so training is not expected to be bitwise identical, while the model, loss,
  and effective batch size remain aligned.
- The `ceil`/`floor` difference depends on the dataloader length and gradient
  accumulation. They are equivalent for the current LIBERO configuration
  because gradient accumulation is 1; other datasets or future configurations
  should be checked separately.
- The base seed is aligned to 42, but RLinf adds a per-worker `seed_offset` to
  keep environment streams and reset-state partitions distinct. This shared
  LIBERO behavior is intentionally preserved.

This comment records these intentional differences for future maintenance.

F9rozen (approved):

rlinf/workers/sft/fsdp_sft_worker.py
```diff
@@ -82,12 +82,35 @@ def __init__(self, cfg: DictConfig):
...
 
     def init_worker(self):
+        self._set_total_training_steps_if_missing()
         self.setup_model_and_optimizer()
 
         if self.cfg.actor.get("enable_offload", False):
             self.offload_param_and_grad()
             self.offload_optimizer()
 
+    def _set_total_training_steps_if_missing(self) -> None:
```
FxxxxU: I suggest changing `_set_total_training_steps_if_missing` here to `_infer_total_training_steps` to make it easier to understand.

examples/embodiment/config/model/fastwam.yaml
```diff
@@ -0,0 +1,31 @@
+model_type: "fastwam"
+precision: bf16
+is_lora: false
+
+# Fine-tuned FastWAM checkpoint.
+model_path: /your_path_to/fastwam/libero_uncond_2cam224.pt
```
FxxxxU: It would be best to standardize this format with the other YAML files in RLinf; replace `/your_path_to/` with `path/to/`.

FxxxxU (approved):
There are no other serious bugs, but I suggest fixing two naming issues.

examples/embodiment/config/model/fastwam.yaml
FxxxxU: Follow RLinf's rules for files.

Iron-Wph: fixed, rename to "/path/to/..."

rlinf/workers/sft/fsdp_sft_worker.py
Iron-Wph: fixed, rename to  "_infer_total_training_steps"

andylin-hao (approved):

guozhen1997 (approved):

PR #1400, opened 2026-07-22
feat(agent): add multi-agent FSDP (MA-FSDP) actor support

qweadzsc:
### Description

Add multi-agent FSDP (MA-FSDP) actor support for SearchR1 dynamic rollout training.

- Add `MAFSDPActor` for dynamic rollout batches, trajectory packing, DP load balancing, dynamic GRPO advantages, and collocated actor/rollout execution.
- Select MA-FSDP from `examples/agent/searchr1/train.py` when `agentloop.is_dynamic_rollout_batch` is enabled.
- Add `train_qwen2.5_fsdp.yaml` as a separate FSDP SearchR1 example without changing the existing Megatron configuration.
- Align the SearchR1 FSDP e2e configuration with the MA-FSDP example and run it in the existing agent e2e workflow.
- Keep FSDP reference-policy swapping and CPU state-dict handling compatible with the MA-FSDP path.

### Motivation and Context

SearchR1 dynamic multi-agent rollouts previously had a Megatron actor implementation only. This change provides the corresponding FSDP actor path while preserving the existing Megatron workflow.

### How has this been tested?

Hardware: 8*NVIDIA A800-SXM

Ran the Qwen2.5 SearchR1 workflow on NVIDIA with both FSDP and Megatron. The resulting training curves are closely aligned

To test the fsdp path, run

```
bash examples/agent/searchr1/run_train.sh train_qwen2.5_fsdp
```

NV-Megatron

NV-FSDP

The first figure is the curve of Megatron. The second is FSDP.

### Types of changes

### Checklist:


examples/agent/searchr1/config/train_qwen2.5.yaml
```diff
@@ -34,7 +34,7 @@ runner:
   max_tokens_per_mbs: 11008
 
   resume_dir: null
-  experiment_name: searchr1-grpo-qwen2.5-3b-ins
+  experiment_name: searchr1-grpo-deepseek-r1-distill-qwen-1.5b-fsdp
```
qurakchin: deepseek-r1 model need to use deepseek-r1 style tool parser.
the tokenizer cat method in the agentloop now is only for non-thinking model. the tokenizer cat method need to be fix when using deepseek-r1 model.

examples/agent/searchr1/config/train_qwen2.5.yaml
```diff
@@ -58,14 +58,17 @@ algorithm:
   # GRPO loss params
   loss_type: actor
   loss_agg_func: "token-mean"
-  kl_beta: 0.001 # 0.001
+  # kl_beta: 0.001 # 0.001
+  kl_beta: 0.0
```
qurakchin: the original search r1 uses kl_beta. what's the reason to close it?

examples/agent/searchr1/config/train_qwen2.5.yaml
```diff
@@ -74,6 +77,8 @@ algorithm:
   early_stop_imp_ratio: 5.0
   use_valid_token_scale: False
 
+  entropy_op_type: torch
```
qurakchin: this entropy_op calculator use fp16 and it may lower the reward curve

examples/agent/searchr1/config/train_qwen2.5.yaml
```diff
@@ -99,7 +104,8 @@ data:
   shuffle: True
   validation_shuffle: True
   seed: 1234
-  train_data_paths: ["/path/to/data/searchr1_train.jsonl"]
+  train_data_paths: ["/inspire/sj-ssd/project/inference-chip/cengshulin-p-zengshulin/datasets/Search-R1-Data/train.jsonl"]
```
qurakchin: do not use your own path

examples/agent/searchr1/config/train_qwen2.5.yaml
```diff
@@ -123,9 +129,9 @@ rollout:
   group_name: "RolloutGroup"
   gpu_memory_utilization: 0.75
   model:
-    model_path: /path/to/model/Qwen2.5-3B-Instruct  # https://huggingface.co/Qwen/Qwen2.5-3B-Instruct
+    model_path: /inspire/sj-ssd/project/inference-chip/cengshulin-p-zengshulin/models/deepseek-r1-distill-qwen-1.5b
```
qurakchin: do not use your own path

examples/agent/searchr1/config/train_qwen2.5.yaml
```diff
@@ -136,7 +142,7 @@ rollout:
 
   rollout_backend: sglang       # here choose which backend to rollout,support [sglang, vllm]
   sglang:
-    attention_backend: triton   # [flashinfer, triton] for more, see sglang's doc
+    attention_backend: ascend   # [flashinfer, triton] for more, see sglang's doc
```
qurakchin: add a new yaml. do not change the exist train_qwen2.5.yaml for nvidia

examples/agent/searchr1/config/train_qwen2.5.yaml
```diff
@@ -174,146 +180,88 @@ reward:
...
-  offload_weight: True
-  offload_grad: True
+  global_batch_size: 8
+  micro_batch_size: 1
 
   model:
-    precision: fp16
+    model_type: ${rollout.model.model_type}
+    model_path: ${rollout.model.model_path}
+    precision: bf16
```
qurakchin: fsdp use fp32 to enable mixed precision

examples/agent/searchr1/config/train_qwen2.5.yaml
```diff
@@ -174,146 +180,88 @@ reward:
...
-    batch_p2p_comm: False
-    variable_seq_lengths: True
-    gradient_accumulation_fusion: False
-    moe_token_dispatcher_type: alltoall
-    use_cpu_initialization: False
-
   optim:
     optimizer: adam
-    lr: 1.0e-06
+    bf16: True
```
qurakchin: many config is not needed in fsdp

examples/agent/searchr1/run_train_ascend.sh
```diff
@@ -0,0 +1,29 @@
...
+export TOKENIZERS_PARALLELISM=false
+export RAY_DEDUP_LOGS=0
+export HCCL_INTRA_ROCE_ENABLE=1
+
+CONFIG_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+REPO_PATH=$(dirname $(dirname $(dirname "$CONFIG_PATH")))
+MEGATRON_PATH=/opt/Megatron-LM
+export PYTHONPATH=${REPO_PATH}:${MEGATRON_PATH}:${REPO_PATH}/examples:$PYTHONPATH
+
+#支持NPU单卡多进程
```
qurakchin: use english

examples/agent/tools/search_local_server_qdrant/launch_local_server.sh
```diff
@@ -3,12 +3,12 @@
 set -ex
 
 # Launch step 1: set your wiki_dataset pages path
-WIKI2018_DIR=/your/wiki_dataset/path
+WIKI2018_DIR=/inspire/sj-ssd/project/inference-chip/cengshulin-p-zengshulin/datasets/local_wiki
```
qurakchin: do not use your own path

examples/agent/tools/search_local_server_qdrant/launch_local_server.sh
```diff
@@ -3,12 +3,12 @@
...
 pages_file=$WIKI2018_DIR/wiki_webpages.jsonl
 
 # Launch step 2: set your retriever model path
 retriever_name=e5
-retriever_path=/your/retriever/model/path
+retriever_path=/inspire/sj-ssd/project/inference-chip/cengshulin-p-zengshulin/models/e5-base-v2
```
qurakchin: do not use your own path

examples/agent/tools/search_local_server_qdrant/local_retrieval_server.py
```diff
@@ -12,64 +12,324 @@
...
+# These globals are intentionally initialized once per encoder worker process.
+_GLOBAL_ENCODER: Optional[Encoder] = None
+_GLOBAL_ENCODER_DEVICE: Optional[str] = None
+
+
+@dataclass(frozen=True)
+class DeviceSpec:
+    """A normalized device descriptor that is safe to pass to child processes."""
+
+    backend: str  # "cuda", "npu", or "cpu"
```
qurakchin: cpu is not need

examples/agent/tools/search_local_server_qdrant/local_retrieval_server.py
```diff
@@ -12,64 +12,324 @@
...
-        assert "global_encoder" in globals()
-        encoder: Encoder = globals()["global_encoder"]
-        return encoder.encode(*args, **kwargs)
+# These globals are intentionally initialized once per encoder worker process.
+_GLOBAL_ENCODER: Optional[Encoder] = None
+_GLOBAL_ENCODER_DEVICE: Optional[str] = None
+
+
+@dataclass(frozen=True)
+class DeviceSpec:
```
qurakchin: too long for the device check

examples/reasoning/config/hybrid_engines/fsdp.yaml
```diff
@@ -6,12 +6,12 @@ strategy: "fsdp"            # FSDP strategy: ["fsdp", "fsdp2"]
 # "no_shard" (no sharding)
 sharding_strategy: "full_shard" 
 
-gradient_checkpointing: False # if True, gradient checkpointing will be used for FSDP/FSDP2 training, which will save memory but increase computation time.
-cpu_offload: False            # whether to offload parameters and gradients to CPU when not in use，if True, actor's enable_offload should be True too.
+gradient_checkpointing: True  # if True, gradient checkpointing will be used for FSDP/FSDP2 training, which will save memory but increase computation time.
```
qurakchin: do not change the public config.

rlinf/hybrid_engines/fsdp/strategy/fsdp.py
```diff
@@ -39,6 +39,49 @@
...
+
+def get_local_npu_device():
+    local_rank = int(os.environ["LOCAL_RANK"])
+    torch.npu.set_device(local_rank)
+    return torch.device(f"npu:{local_rank}")
+
+
+def make_meta_init_fn(target_device: torch.device, root_model: nn.Module = None):
+    """
+    FSDP1 会在 flatten/shard 前对 meta module 调用 param_init_fn。
```
qurakchin: use english

qurakchin:
the pr is dirty and not ready. convert to draft.

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: Added a new DeepSeek-R1-style tool parser in `rlinf/algorithms/toolcall_parsers.py`

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: I forgot to restore kl_beta after testing; it is now back to the original value and has been tested.

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: FP32 caused NaNs during training, so we use BF16 instead.

examples/agent/tools/search_local_server_qdrant/build_index_ascend.py
```diff
@@ -0,0 +1,321 @@
+# Copyright 2025 The RLinf Authors.
```
qurakchin: Is this file need to be change? what's the difference to the nvidia version?

rlinf/agents/searchr1/searchr1_agent_loop.py
```diff
@@ -167,10 +170,21 @@ async def generate_tool_response(
             message = {"role": "tool", "content": tool_response.text}
             tool_messages.append(message)
 
-        # Tokenize tool responses
-        tool_response_ids: list[int] = self.tokenizer.encode(
-            tool_messages[0]["content"], add_special_tokens=False
-        )
+        # DeepSeek-R1 encodes tool outputs with dedicated chat-template tokens.
```
qurakchin: Create and use a new function of `encode_tool_resp`.

rlinf/hybrid_engines/fsdp/strategy/fsdp.py
```diff
@@ -185,6 +228,23 @@ def wrap_model(self, model: nn.Module, device_mesh: DeviceMesh) -> FSDP:
             use_orig_params=self.cfg.fsdp_config.use_orig_params,
             cpu_offload=cpu_offload,
         )
+
```
qurakchin: delete unused comment.

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
```diff
@@ -289,6 +309,16 @@ def setup_model_and_optimizer(self) -> None:
         self.model = self._strategy.wrap_model(
             model=module, device_mesh=self._device_mesh
         )
+
+        if self._cfg.fsdp_config.get("init_model_with_meta_device", False):
+            self._strategy.load_hf_checkpoint_to_fsdp2_model(
```
qurakchin: why it's not need for fsdp1?

rlinf/scheduler/collective/multi_channel_pg.py
```diff
@@ -856,7 +857,11 @@ def _new_process_group_helper(
                     dist_backend_opts.group_id = group_name
                     dist_backend_opts.global_ranks_in_group = global_ranks_in_group
 
-                    backend_class = creator_fn(dist_backend_opts, pg_options)
+                    try:
+                        backend_class = creator_fn(dist_backend_opts, pg_options)
+                    except BaseException as e:
+                        traceback.print_exc()
```
qurakchin: why use traceback.print_exc

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -862,8 +1358,53 @@ def iterator_wrapper():
             # scale loss for gradient accumulation and backprop
             final_loss_metric = loss.detach()
             loss = loss / self.gradient_accumulation
-            with backward_ctx:
-                self.grad_scaler.scale(loss).backward()
+
```
qurakchin: do not add debug code.

rlinf/workers/rollout/sglang/sglang_worker.py
```diff
@@ -170,7 +175,7 @@ def _init_engine(self):
             ),
             tp_size=self._cfg_rollout.tensor_parallel_size,
             mem_fraction_static=self._cfg_rollout.gpu_memory_utilization,
-            enable_memory_saver=use_cudagraph,
+            enable_memory_saver=True,  # TODO: =use_cudagraph
```
qurakchin: solve this

MAFSDPActor_work_summary.md
```diff
@@ -0,0 +1,127 @@
+# MAFSDPActor 阶段工作说明
```
qurakchin: strange file

mp_wrapper.tz
qurakchin, line 1: strange file

examples/agent/searchr1/config/train_qwen2.5.yaml
```diff
@@ -178,7 +178,7 @@ actor:
   enable_dp_load_balance: True
```
qurakchin, line 178: add a train_qwen2.5_fsdp.yaml and add the curve to the pr comment

examples/agent/tools/search_local_server_qdrant/retriever_server_accel.py
```diff
@@ -0,0 +1,659 @@
+# Copyright 2025 The RLinf Authors.
```
qurakchin: this file can also be merged

examples/agent/searchr1/config/train_qwen2.5_ascend.yaml
```diff
@@ -0,0 +1,237 @@
...
+  training_backend: fsdp
+  enable_offload: True
+  enable_dp_load_balance: True
+  seed: 1234
+  pack_traj: True
+
+  model:
+    model_type: ${rollout.model.model_type}
+    model_path: ${rollout.model.model_path}
+    precision: bf16
```
qurakchin: use fp32 to enable mix precision

examples/agent/searchr1/config/train_qwen2.5_ascend.yaml
```diff
@@ -0,0 +1,237 @@
...
+  shuffle: True
+  validation_shuffle: True
+  seed: 1234
+  train_data_paths: ["/path/to/data/searchr1_train.jsonl"]
+  val_data_paths: ["/path/to/data/searchr1_val.jsonl"]
+
+
+agentloop:
+  group_name: "AgentLoopGroup"
+  toolcall_parser: "searchr1-deepseek-r1"
```
qurakchin: use qwen not deepseek-r1

examples/agent/searchr1/config/train_qwen2.5_fsdp.yaml
```diff
@@ -0,0 +1,237 @@
...
+  pack_traj: True
+
+  model:
+    model_type: ${rollout.model.model_type}
+    model_path: ${rollout.model.model_path}
+    precision: bf16
+    attn_implementation: eager
+    encoder_seq_length: ${runner.seq_length}
+
+  optim:
```
qurakchin, line 194: use same config in train_qwen2.5

rlinf/agents/searchr1/searchr1_agent_loop.py
```diff
@@ -46,6 +46,9 @@ def __init__(
         assert self.toolcall_parser is not None, (
             "toolcall_parser must be set in searchr1"
         )
+        self.use_native_tool_messages = cfg.agentloop.get(
```
qurakchin: use a attr like `toolcall_parser` such as `toolresp_encoder`.

rlinf/agents/searchr1/searchr1_agent_loop.py
```diff
@@ -141,6 +144,23 @@ async def generate_llm_response(
 
         return True, llm_response_ids, llm_response_text, llm_output
 
+    def encode_tool_resp(self, tool_messages: list[dict[str, str]]) -> list[int]:
+        """Encode tool responses using the configured model protocol."""
+        if not tool_messages:
+            raise ValueError("At least one tool response is required")
+
+        if not self.use_native_tool_messages:
```
qurakchin: use `toolresp_encoder` to differentiate branches

rlinf/algorithms/toolcall_parsers.py
```diff
@@ -91,6 +91,51 @@ async def __call__(self, response_text: str) -> tuple[str, list[ToolRequest]]:
         return content, function_calls
 
 
+@register_toolcall_parser("searchr1-deepseek-r1")
```
qurakchin: `searchr1-` is not needed.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -519,6 +521,34 @@ def _load_weight_and_optimizer(self) -> None:
             if self.is_optimizer_offloaded:
                 self.load_optimizer(self.device)
 
+    @contextmanager
+    def _swap_to_ref_policy(self):
```
qurakchin: what's the advantage to the cpu_weight_swap?

rlinf/utils/utils.py
```diff
@@ -36,7 +36,10 @@ def clear_memory(sync=True):
     if sync:
         Worker.torch_platform.synchronize()
     gc.collect()
-    Worker.torch_platform.ipc_collect()
+    try:
+        Worker.torch_platform.ipc_collect()
+    except Exception:
+        pass
```
qurakchin: why to pass the failing process

rlinf/utils/utils.py
```diff
@@ -203,14 +206,14 @@ def retrieve_model_state_dict_in_cpu(model, offloaded_buffer=None):
...
                 offloaded_buffer[name].copy_(item.detach(), non_blocking=True)
             else:
-                item = (
-                    item.detach()
-                    .to(device="cpu", non_blocking=True, copy=True)
-                    .pin_memory()
-                )
+                item = item.detach().to(device="cpu", non_blocking=True, copy=True)
+                if not isinstance(item, DTensor):
+                    item = item.pin_memory()
```
qurakchin: why does it no pin when item is a DTensor

qurakchin:
i think it's better to split this pr into 2 pr:
1. agentic support in fsdp2, such as ma fsdp actor / deepseek-r1 support. align the reward curve between fsdp actor and megatron actor on nvidia chip.
2. ascend support in ascend, such as your accelerate utils and changes to sglang and fsdp workers. align the reward curve between ascend and nvidia.

andylin-hao (commented):
Thanks for this — the Ascend-specific work is well isolated in places (the NPU weight-sync branch is explicitly gated and the CUDA path kept verbatim, and `init_model_with_meta_device` is opt-in and validated). My concerns are mostly about collateral on the **shared NVIDIA path**, plus packaging/CI structure. Reviewed statically against `main`; I have no Ascend hardware and ran nothing.

## Blocking

**1. A deliberate Adam bias-correction fix was deleted.** `warmup_optimizer_state` ends on `main` by resetting each param's `step` counter to 0 after the fake `optimizer.step()`, with a comment explaining why ([`main:rlinf/utils/utils.py#L635-L648`](https://github.com/RLinf/RLinf/blob/main/rlinf/utils/utils.py#L635-L648)). That loop and its comment are gone here. The fake step now leaves `step == 1`, so the first real update applies bias correction for t=2 instead of t=1 — with β₁=0.9 the t=1 factor is ~10×, so the first update's effective step size is materially wrong. This is silent and hits **every FSDP run on every platform**. See the inline note: the replacement `init_adamw_optimizer_state` is never called, so this looks like a half-finished swap.

**2. An NVML init/shutdown was added to per-step paths.** `AcceleratorUtil.get_accelerator_type()` is not cached — it calls `Accelerator.enumerate()` → `get_accelerator_model()` → `pynvml.nvmlInit()/nvmlShutdown()` every call. Inline notes mark the four sites; the worst is inside the per-parameter loop of `batch_load_hf_weight` (thousands of driver round-trips per weight sync on a 32B model). NVIDIA users pay all of this cost for a branch that is otherwise unchanged.

**3. Packaging should extend `install.sh`, not fork it.** `requirements/install.sh` already has `SUPPORTED_PLATFORMS=("nvidia" "amd" "ascend")`, an `install_ascend_extras()` that installs a torch-matched `torch-npu` and sources the CANN `set_env.sh`, and an `install_platform_extras()` dispatcher. This PR touches neither `install.sh` nor `docker/`, and instead adds a parallel flat script that re-implements `install_uv()` and `create_or_reuse_venv()` (duplicating `create_and_sync_venv`). It also can't compose with `--target embodied/agentic` or `--engine sglang/vllm`. Please fold the wheel/SGLang-NPU logic into `install_ascend_extras()` and the engine install path so `bash requirements/install.sh agentic --platform ascend --engine sglang` works, and add the matching Dockerfile stage.

**4. No CI.** `.github/` is untouched, so nothing exercises this — not the Ascend path, and not the shared-code changes on CUDA, which all run on existing hardware. A unit test asserting `warmup_optimizer_state` leaves `step == 0` would have caught (1) outright. The only test added is `test_data_iter_utils.py`, which covers the one shared change that was implemented correctly.

**5. No documentation.** Zero files under `docs/`. This adds a new supported accelerator, an installer, 4 configs, 3 launchers, a retriever server, and two new config keys (`fsdp_config.init_model_with_meta_device`, `model.attn_implementation`) — none documented in EN or ZH. Please add an Ascend installation section (both languages), document the new config keys, and register the examples in the gallery index.

## Non-blocking

- **Three parallel device-resolution helpers** are introduced: `_get_metric_compute_device` (`distributed.py`) and `_get_runtime_device` (`data_iter_utils.py`) are ~90% identical, alongside the `AcceleratorUtil.get_device_type(get_accelerator_type())` idiom — while `Worker.torch_platform` / `Worker.torch_device_type` already existed and is what this PR *removes* from `distributed.py:222`. Worth consolidating into one helper. (`data_iter_utils.py` gets it right: hoisted once per call, cheap idiom.)
- **The reference-policy swap was rewritten for all platforms** (`fsdp_actor_worker.py`): `cpu_weight_swap` → a local `_swap_to_ref_policy`, and full → sharded state dict. The `try/finally` restore is correct, but this changes the KL path on CUDA and holds two CPU copies during the swap — worth calling out in the description as a deliberate change.
- `clear_memory` now swallows `ipc_collect()` failures with a bare `except Exception: pass`.
- Dead code: `init_adamw_optimizer_state` (64 lines, unreferenced) and a commented-out block in `_execute_with_device_injection`.

## Verified clean

Example-config model paths all follow the `/path/to/model/<repo>` convention with matching `# https://huggingface.co/...` comments, and all three referenced repos return `200` with no rename redirect. `.gitignore` additions are appropriate and it fixes the missing trailing newline. New files carry correct license headers. The description is unusually thorough and reproducible.

rlinf/utils/utils.py
```diff
@@ -632,11 +635,75 @@ def zero_grad_like(p):
...
         p.grad = saved_grads[p]
 
-    # The empty step above advances each Adam param's step counter to 1, which biases
-    # the first real update through bias correction and diverges from a freshly-built
-    # optimizer. Reset the step counter to 0 so this state initialization is a true
-    # no-op for training dynamics, while preserving the already-zeroed exp_avg/exp_avg_sq
-    # entries so load_state_dict still finds initialized state.
+
+@torch.no_grad()
+def init_adamw_optimizer_state(optimizer):
```
andylin-hao: **Blocking — silent training regression on all platforms.**

This new function is never called anywhere in the PR (`git grep init_adamw_optimizer_state` returns only this definition), while `fsdp_model_manager.py:606` still calls `warmup_optimizer_state`. In the process, `warmup_optimizer_state` lost the step-reset loop it has on `main`:

```python
# main:rlinf/utils/utils.py:635-648
# The empty step above advances each Adam param's step counter to 1, which biases
# the first real update through bias correction ... Reset the step counter to 0 ...
for p in all_params:
    st = optimizer.state.get(p, {})
    step = st.get("step", None)
    ...
```

So the fake `optimizer.step()` now leaves `step == 1` and the first real update is bias-corrected as t=2. Either restore that loop, or wire this new function into `fsdp_model_manager.py:606` and delete the old one — and please add a test asserting `state["step"] == 0` after warmup.

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
```diff
@@ -121,11 +202,18 @@ def batch_load_hf_weight(self, state_dict: dict[str, Any]) -> Any:
         if rollout_sync_mode_collocated:
             for name, handle in state_dict.items():
                 func, args = handle
-                list_args = list(args)
-                # NOTE: the key is to change device id to the current device id
-                # in case two processes have different CUDA_VISIBLE_DEVICES
-                list_args[6] = torch.cuda.current_device()
-                new_weight = func(*list_args)
+                if AcceleratorUtil.get_accelerator_type() == AcceleratorType.NPU:
```
andylin-hao: **Blocking — this gate runs per parameter tensor.**

This is inside `for name, handle in state_dict.items()`, and `AcceleratorUtil.get_accelerator_type()` is uncached: it calls `Accelerator.enumerate()` → `NvidiaGPUManager.get_accelerator_model()` → `pynvml.nvmlInit()` / `nvmlDeviceGetHandleByIndex` / `nvmlDeviceGetName` / `nvmlShutdown()`.

For a 32B model that's thousands of driver round-trips on **every weight sync**, and NVIDIA pays it even though the `else` branch below is unchanged. Please hoist the accelerator-type check out of the loop (or cache it at module level / with `functools.lru_cache`).

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
```diff
@@ -109,6 +131,65 @@ def resume_memory_occupation(self, recv_req: ResumeMemoryOccupationReqInput):
...
+                    platform.device(inject_value)
+                    if hasattr(platform, "device") and device_type != "cpu"
+                    else nullcontext()
+                )
+                with device_context:
+                    new_weight = func(**kwargs)
+                # if hasattr(new_weight, 'device') and new_weight.device != inject_value:
+                #     new_weight = new_weight.to(target_device)
+                return new_weight
+            except (ValueError, TypeError, AttributeError):
```
andylin-hao: **Silent fallback can send wrong-device weights.**

This swallows `TypeError` raised *inside* `func(**kwargs)` — i.e. a genuine weight-reconstruction failure — and then falls through to `return func(*args)` at the end of the function, with the **original, un-redirected** device argument. That's precisely the failure this helper exists to prevent, reached silently.

Please narrow the catch to the signature-inspection call only, and let errors from `func` itself propagate. The commented-out `# if hasattr(new_weight, 'device')...` block below should also go.

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
```diff
@@ -262,6 +350,12 @@ def init_rlinf_worker(
             validate_weight_first_sync = self.cfg.rollout.get(
                 "validate_weight_first_sync", False
             )
+            if self.cfg.actor.training_backend == "fsdp":
+                # FSDP rollout sync rebuilds tensors from actor-provided payloads rather
+                # than comparing two identical HF initialization paths. The first-sync
+                # norm check is therefore prone to false positives and is only reliable
+                # for the Megatron path.
+                validate_weight_first_sync = False
```
andylin-hao: **Blocking — this silently overrides user config on the existing CUDA path.**

The condition is `training_backend == "fsdp"`, not Ascend, so any current NVIDIA FSDP user who sets `validate_weight_first_sync: True` now has it ignored without a warning and loses a weight-sync correctness check. It also conflicts with the project rule that code must not overwrite user-facing config fields.

If the check genuinely can't work under FSDP, please reject the combination in `validate_cfg` with a clear message, or at minimum `log_warning` when overriding.

rlinf/data/io_struct.py
```diff
@@ -765,29 +766,46 @@ def to_actor_batch(
 
         max_response_len = training_seq_length - data_seq_length
 
+        target_device = AcceleratorUtil.get_device_type(
```
andylin-hao: Same uncached-NVML concern as the weight-sync loop: `get_accelerator_type()` does a full `nvmlInit()/nvmlShutdown()` cycle, and `to_actor_batch` runs per training batch (also at line ~1186).

`Worker.torch_platform.current_device()` is already imported in this file and is free — or resolve `target_device` once at module scope.

rlinf/envs/action_utils.py
```diff
@@ -54,9 +55,12 @@ def prepare_actions_for_maniskill(
     actions["terminate_episode"] = np.array([0.0] * batch_size).reshape(-1, 1)  # [B, 1]
 
     actions = {k: torch.tensor(v, dtype=torch.float32) for k, v in actions.items()}
+    target_device = AcceleratorUtil.get_device_type(
```
andylin-hao: Hot path: `prepare_actions_for_maniskill` is called every env chunk step, and `get_accelerator_type()` does an NVML init/shutdown per call. Please resolve this once (module-level or `functools.lru_cache`) rather than per step.

rlinf/utils/distributed.py
```diff
@@ -893,7 +925,10 @@ def masked_normalization(
             Normalized x, with the same shape as x.
     """
     dtype = torch.float64 if high_precision else torch.float32
-    x = x.to(dtype=dtype).cuda()
+    target_device = AcceleratorUtil.get_device_type(
```
andylin-hao: Same per-call NVML cost in `masked_normalization`, which runs on every advantage normalization.

Worth noting this file already uses the cheap idiom — and this PR *removes* one such usage at line 222 (`Worker.torch_platform.current_device()`) in favor of the new helper. Consolidating on `Worker.torch_platform` would fix both the cost and the inconsistency.

rlinf/hybrid_engines/fsdp/strategy/fsdp2.py
```diff
@@ -201,3 +207,507 @@ def before_micro_batch(
...
+                                )
+
+                                # Some implementations also keep attention_scaling as an attr.
+                                if hasattr(module, "attention_scaling"):
+                                    module.attention_scaling = attention_scaling
+                            else:
+                                inv_freq = _build_rope_inv_freq(module, root_config)
+
+                        except Exception:
+                            # Fallback to standard RoPE formula.
```
andylin-hao: Preferring `module.rope_init_fn` above is the right call — it handles `linear`/`dynamic`/`yarn`/`llama3`/`longrope` and sets `attention_scaling`.

But this `except Exception` falls back to `_build_rope_inv_freq`, which implements only the **default** formula and ignores `config.rope_scaling` entirely. For a scaled-rope model that silently yields wrong positional encoding — no error, no warning, just degraded results that look like a bad hyperparameter.

Suggest logging a warning here, and re-raising when `config.rope_scaling` is set and non-trivial, since the fallback is known-wrong in that case.

requirements/install_ascend.sh
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env bash
```
andylin-hao: **Blocking — this deviates from the existing install design.**

`requirements/install.sh` already treats Ascend as a first-class platform:

- `SUPPORTED_PLATFORMS=("nvidia" "amd" "ascend")`
- `install_ascend_extras()` — installs a torch-matched `torch-npu`, adds `pyyaml`/`decorator`, and sources `/usr/local/Ascend/ascend-toolkit/set_env.sh`
- `install_platform_extras()` dispatching on `$PLATFORM`

This PR touches neither `install.sh` nor `docker/`, and this script re-implements `install_uv()` and `create_or_reuse_venv()` (duplicating `create_and_sync_venv`). Being flat, it also can't compose with `--target embodied/agentic` or `--engine sglang/vllm`, so the two installers will drift.

Please fold the wheel download + SGLang-NPU logic into `install_ascend_extras()` and the engine install path so the documented entrypoint works:

```bash
bash requirements/install.sh agentic --platform ascend --engine sglang
```

and add the matching Dockerfile stage + CI job, per the `add-install-docker-ci-e2e` conventions.

rlinf/algorithms/toolcall_parsers.py
```diff
@@ -18,13 +18,37 @@
...
+    register_toolcall_parser,
+    register_toolresp_encoder,
+)
 from rlinf.data.tool_call.tool_io_struct import (
     ToolRequest,
     ToolResponse,
 )
 
 
+@register_toolresp_encoder("qwen")
```
qurakchin: add to a split file.

qurakchin (commented):
have you tested thinking tag is correct in prompt id in deepseek r1 model on search r1?

rlinf/workers/actor/fsdp_actor_worker.py
qweadzsc: cpu_weight_swap assumes a regular module state dict, while FSDP requires its strategy-aware state-dict APIs for sharded/DTensor weights. CPU offload only avoids a second GPU-resident policy copy.

qurakchin: It's better to add fsdp / fsdp2 support in cpu_weight_swap and retrieve_model_state_dict_in_cpu by adding an `is_fsdp=False` arg.

rlinf/workers/actor/ma_fsdp_actor_worker.py
```diff
@@ -0,0 +1,655 @@
...
+
+            def iterator_wrapper():
+                for _ in range(micro_batch_cnt):
+                    yield next(batch)
+
+            micro_batches_iter = iterator_wrapper()
+        self.optimizer.zero_grad()
+        mbs_metrics_list = {}
+        for idx, m_batch in enumerate(micro_batches_iter):
+            backward_ctx = self.before_micro_batch(
```
qurakchin: It's better to refactor the body in the for loop to a function in FSDPActor, such as a 'forward_backward_batch' with (idx, m_batch) input and (mbs_metrics_data) output. Then MAFSDPActor only need to override this.

rlinf/workers/actor/fsdp_actor_worker.py
qweadzsc: Refactored as suggested: cpu_weight_swap() and retrieve_model_state_dict_in_cpu() now support FSDP/FSDP2 through an is_fsdp flag, and the _swap_to_ref_policy() implementation has been removed.

rlinf/workers/actor/ma_fsdp_actor_worker.py
qweadzsc: Moved the per-micro-batch forward/loss/backward logic into overridable FSDPActor.forward_backward_batch(); MAFSDPActor now only overrides this method for dynamic-rollout-specific loss scaling and masks.

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5.yaml
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/run_train_ascend.sh
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/tools/search_local_server_qdrant/launch_local_server.sh
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/tools/search_local_server_qdrant/launch_local_server.sh
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/tools/search_local_server_qdrant/local_retrieval_server.py
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/tools/search_local_server_qdrant/local_retrieval_server.py
qweadzsc: Moved to the follow-up Ascend PR

examples/reasoning/config/hybrid_engines/fsdp.yaml
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/fsdp/strategy/fsdp.py
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/tools/search_local_server_qdrant/build_index_ascend.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/agents/searchr1/searchr1_agent_loop.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/fsdp/strategy/fsdp.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/scheduler/collective/multi_channel_pg.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/workers/actor/fsdp_actor_worker.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/workers/rollout/sglang/sglang_worker.py
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/tools/search_local_server_qdrant/retriever_server_accel.py
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5_ascend.yaml
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5_ascend.yaml
qweadzsc: Moved to the follow-up Ascend PR

rlinf/agents/searchr1/searchr1_agent_loop.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/agents/searchr1/searchr1_agent_loop.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/algorithms/toolcall_parsers.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/utils/utils.py
qweadzsc: Moved to the follow-up Ascend PR

examples/agent/searchr1/config/train_qwen2.5_ascend.yaml
qweadzsc: Moved to the follow-up Ascend PR

rlinf/utils/utils.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/agents/searchr1/searchr1_agent_loop.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/utils/utils.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/agents/searchr1/searchr1_agent_loop.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/algorithms/toolcall_parsers.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/utils/utils.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/utils/utils.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/data/io_struct.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/utils/utils.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/envs/action_utils.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/utils/distributed.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/fsdp/strategy/fsdp2.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
qweadzsc: Moved to the follow-up Ascend PR

requirements/install_ascend.sh
qweadzsc: Moved to the follow-up Ascend PR

rlinf/data/io_struct.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/algorithms/toolcall_parsers.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/envs/action_utils.py
qweadzsc: Moved to the follow-up Ascend PR

rlinf/utils/distributed.py
qweadzsc: Moved to the follow-up Ascend PR

requirements/install_ascend.sh
qweadzsc: Moved to the follow-up Ascend PR

qweadzsc: Moved to the follow-up Ascend PR

rlinf/algorithms/toolcall_parsers.py
qweadzsc: Moved to the follow-up Ascend PR

qweadzsc: Moved to the follow-up Ascend PR

MAFSDPActor_work_summary.md
qweadzsc: Removed.

mp_wrapper.tz
qweadzsc: Removed.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -640,11 +642,7 @@ def inference_step(
             assert self.ref_policy_state_dict is not None, (
                 "Reference policy state dict is None but compute_ref_logprobs is True"
             )
-            with cpu_weight_swap(
-                self.model,
-                self.ref_policy_state_dict,
-                self.offload_model_buffer,
-            ):
+            with cpu_weight_swap(self, self.ref_policy_state_dict, is_fsdp=True):
```
qurakchin, line 551: Have you tested that would it be slower after remove the offload_model_buffer.

rlinf/workers/actor/ma_fsdp_actor_worker.py
```diff
@@ -0,0 +1,664 @@
...
+
+        clip_ratio = self.cfg.algorithm.ratio_clip_eps
+        clip_ratio_low = self.cfg.algorithm.get("clip_ratio_low", None)
+        clip_ratio_high = self.cfg.algorithm.get("clip_ratio_high", None)
+        clip_ratio_low = clip_ratio_low if clip_ratio_low is not None else clip_ratio
+        clip_ratio_high = clip_ratio_high if clip_ratio_high is not None else clip_ratio
+        clip_ratio_c = self.cfg.algorithm.get("clip_ratio_c", 3.0)
+
+        if self.cfg.algorithm.get("importance_sampling_fix", False):
+            raise AssertionError(
```
qurakchin: If not support,  raise at the init function

rlinf/workers/actor/ma_fsdp_actor_worker.py
```diff
@@ -0,0 +1,664 @@
...
+            {
+                "actor/final_loss": final_loss_metric,
+                "actor/entropy_loss": entropy_loss.detach(),
+                "actor/kl_loss": kl_loss.detach(),
+            }
+        )
+
+        return mbs_metrics_data
+
+    def training_step(
```
qurakchin: @Worker.timer("training_step")

rlinf/workers/actor/fsdp_actor_worker.py
qweadzsc: I tested this with SearchR1 Qwen2.5-3B on 8xA800 using collocated MA-FSDP for 10 steps. I also evaluated a strategy-aware CPU buffer-reuse path that preserves FSDP1 ShardedTensor/FSDP2 DTensor metadata and reuses only local tensor storage. Excluding initialization, the current implementation achieved 27.83s inference time and 107.75s step time, versus 28.41s and 108.63s with buffer reuse. There is no measurable regression after removing offload_model_buffer, so I kept the simpler strategy-aware path.

rlinf/workers/actor/ma_fsdp_actor_worker.py
qweadzsc: Done. importance_sampling_fix is now rejected during MAFSDPActor initialization with a clear ValueError, rather than failing only after the first micro-batch reaches the training path.

rlinf/workers/actor/ma_fsdp_actor_worker.py
qweadzsc: Done. Added @Worker.timer("training_step") to MAFSDPActor.training_step, matching the base FSDPActor timing metric.

qurakchin (approved):

rlinf/workers/actor/ma_fsdp_actor_worker.py
```diff
@@ -0,0 +1,664 @@
...
+            {
+                "actor/final_loss": final_loss_metric,
+                "actor/entropy_loss": entropy_loss.detach(),
+                "actor/kl_loss": kl_loss.detach(),
+            }
+        )
+
+        return mbs_metrics_data
+
+    @Worker.timer("training_step")
```
andylin-hao: This is a line-for-line copy of `FSDPActor.training_step` (`fsdp_actor_worker.py:849`). The only thing it changes is leaving out two metrics: `rollout_train_kl` and the critic explained-variance stats.

Neither needs to be removed by hand. `compute_rollout_train_kl` already returns `None` when `rollout_logprobs`/`recomputed_logprobs` are missing (`fsdp_actor_worker.py:75`), and the critic block is already behind `if explained_variance_stats:`. So the parent version would behave the same here.

Fix: delete this method and inherit the parent. If the parent's `loss_mask = m_batch["response_mask"][:, -self.response_len:]` is the real reason for the copy, move just that one line into a small method and override that instead.

rlinf/workers/actor/ma_fsdp_actor_worker.py
```diff
@@ -0,0 +1,664 @@
...
+
+        entropy_loss = torch.tensor(0.0, device=Worker.torch_platform.current_device())
+        if self.calculate_entropy:
+            entropy_loss = self.loss_agg_func(entropy, mask=loss_mask)
+            if self.calculate_entropy_loss:
+                loss = loss - self.cfg.algorithm.entropy_bonus * entropy_loss
+
+        kl_loss = torch.tensor(0.0, device=Worker.torch_platform.current_device())
+        if self.kl_beta > 0 and ref_logprobs is not None:
+            kld = kl_penalty(logprobs, ref_logprobs, self.kl_penalty_type)
```
andylin-hao, line 418: Argument order here is `(logprobs, ref_logprobs)`, which matches the signature `kl_penalty(logprob, ref_logprob, kl_penalty)` and both Megatron actors. `FSDPActor` passes them the other way round at `fsdp_actor_worker.py:827`.

One of the two is wrong. For `kl_penalty_type: kl` / `k1` the result is `logprob - ref_logprob`, so swapping the arguments flips the sign and the KL penalty becomes a bonus.

Fix: this line is correct — change `fsdp_actor_worker.py:827` to match it.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -749,6 +747,104 @@ def run_inference(
...
+
+        entropy_loss = torch.tensor(0.0, device=Worker.torch_platform.current_device())
+        if self.calculate_entropy:
+            entropy_loss = self.loss_agg_func(entropy, mask=loss_mask)
+            if self.calculate_entropy_loss:
+                loss = loss - self.cfg.algorithm.entropy_bonus * entropy_loss
+
+        kl_loss = torch.tensor(0.0, device=Worker.torch_platform.current_device())
+        if self.kl_beta > 0 and ref_logprobs is not None:
+            kld = kl_penalty(ref_logprobs, logprobs, self.kl_penalty_type)
```
andylin-hao: The signature is `kl_penalty(logprob, ref_logprob, kl_penalty)`, so this passes the reference logprobs where the policy logprobs are expected. Every other caller uses the opposite order: `megatron_actor_worker.py:254`, `ma_megatron_actor_worker.py:249`, and the new `ma_fsdp_actor_worker.py:413`.

This predates the PR, but it matters now: every existing FSDP e2e config sets `kl_beta: 0.0`, and the config this PR adds to CI sets `kl_beta: 0.001`, so this line starts running in CI with this PR.

Fix: `kld = kl_penalty(logprobs, ref_logprobs, self.kl_penalty_type)`.

rlinf/workers/actor/ma_fsdp_actor_worker.py
```diff
@@ -0,0 +1,664 @@
...
+
+        For multi-agent dynamic rollout, each channel item corresponds to one trajectory
+        group, while `num_sequence` inside that item can vary across turns. Therefore the
+        receive loop should be counted in rollout-result units, not sequence units.
+        """
+        return self.cfg.data.rollout_batch_size // torch.distributed.get_world_size()
+
+    def get_rollout_metrics_group(self, batch: dict[str, torch.Tensor]):
+        """Return the process group used to aggregate rollout metrics."""
+        if batch["input_ids"].shape[0] == 0:
```
andylin-hao: Returning `None` makes `_compute_rollout_metrics` return early and skip `compute_rollout_metrics_dynamic`, which runs `all_gather_object` on the group (`rlinf/utils/distributed.py:77`).

That decision is made from local data only. If one rank gets an empty batch and the others don't, the others enter the collective and this rank never does, so the job hangs until the NCCL timeout instead of failing.

The Megatron version never returns `None` unless the auto-scheduler is on, and in that case it agrees on the participating ranks with an `all_reduce` first (`rlinf/workers/megatron_worker.py:914`).

Fix: always return `torch.distributed.group.WORLD`, or all-reduce the "is my batch empty" flag first so every rank takes the same branch.

rlinf/utils/utils.py
```diff
@@ -214,43 +234,62 @@ def retrieve_model_state_dict_in_cpu(model, offloaded_buffer=None):
...
         offloaded_buffer = retrieve_model_state_dict_in_cpu(
-            resident_model, offloaded_buffer
+            resident_model,
+            offloaded_buffer,
+            is_fsdp=is_fsdp,
         )
 
-    resident_model.load_state_dict(cpu_weights)
+    if is_fsdp:
+        resident_model._strategy.load_model_with_state_dict(
```
andylin-hao: `_strategy` is a private attribute of `FSDPModelManager`, and this is a generic helper module that otherwise knows nothing about FSDP. Two lines up, the read side correctly goes through the public `get_model_state_dict()` wrapper.

Fix: add a public `load_model_with_state_dict()` wrapper on `FSDPModelManager`, right next to `get_model_state_dict` at `fsdp_model_manager.py:325`, and call that instead of `resident_model._strategy`.

rlinf/utils/utils.py
```diff
@@ -196,13 +196,33 @@ def seed_everything(seed: int) -> int:
...
+        resident_model: A regular ``torch.nn.Module`` or an FSDP actor when
+            ``is_fsdp`` is set.
+        offloaded_buffer: Optional buffer to reuse for the CPU state dict.
+        is_fsdp: Whether to retrieve local FSDP/FSDP2 state through its strategy.
+    """
     if offloaded_buffer is None:
         offloaded_buffer = {}
 
-    for name, item in model.state_dict().items():
+    if is_fsdp:
```
andylin-hao, line 211: On this path the function returns before `offloaded_buffer` is ever used, so a buffer passed by the caller is silently dropped and a new CPU copy is allocated every call.

Together with removing `self.offload_model_buffer` at `fsdp_actor_worker.py:275`, this means each `inference_step` allocates a full, non-pinned CPU copy of the model where it used to copy into a reused pinned buffer.

Fix: either fill `offloaded_buffer` on this path too, or raise a `ValueError` when `offloaded_buffer` is passed with `is_fsdp=True`, so it can't be passed by mistake and ignored.

rlinf/workers/actor/fsdp_actor_worker.py
```diff
@@ -275,8 +275,10 @@ def init_worker(self) -> None:
         if (
             self.kl_beta > 0 or self.reinpp_kl_beta > 0
         ) and self.combine_reference_model:
-            self.ref_policy_state_dict = retrieve_model_state_dict_in_cpu(self.model)
-            self.offload_model_buffer = {}
+            self.ref_policy_state_dict = retrieve_model_state_dict_in_cpu(
```
andylin-hao, line 184: This call is updated for the new FSDP path, but `rlinf/workers/inference/fsdp_inference_worker.py:71` still calls `retrieve_model_state_dict_in_cpu(self.model[0])` without `is_fsdp=True`. It hits exactly the DTensor problem this change fixes here.

(`self.model` is assigned the wrapped module at `fsdp_inference_worker.py:63`, so `self.model[0]` looks wrong on its own as well.)

Fix: update that call to `retrieve_model_state_dict_in_cpu(self, is_fsdp=True)` in this PR, or note in the description why the inference worker is left for later.

examples/agent/searchr1/train.py
```diff
@@ -80,7 +92,8 @@ def main(cfg) -> None:
         and cfg.algorithm.recompute_logprobs
     ):
         inference_placement_strategy = component_placement.get_strategy("inference")
-        inference_group = MegatronInference.create_group(
+        inference_worker_cls = get_inference_backend_worker(cfg, "actor")
```
andylin-hao, line 95: This branch only runs under `DISAGGREGATED` placement, but `MAFSDPActor.__init__` asserts collocated placement (`ma_fsdp_actor_worker.py:79`). So `training_backend: fsdp` with disaggregated placement launches all the workers and then dies on a bare assert inside the actor.

Fix: reject that combination in `validate_cfg` in `rlinf/config.py`, so it fails at startup with a message that says what to change.

tests/e2e_tests/agent/searchr1/qwen2.5-1.5b-fsdp.yaml
```diff
@@ -33,7 +33,7 @@ runner:
   enable_dynamic_batch_size: True
   max_tokens_per_mbs: 11008
 
-  resume_dir: auto
+  resume_dir: null
```
andylin-hao: This config was already in the repo but nothing referenced it — this PR is what puts it into CI. Switching `auto` to `null` means the new job never exercises resume.

Fix: keep `resume_dir: auto` if it works, or mention in the PR description why the job can't resume.

tests/e2e_tests/agent/searchr1/qwen2.5-1.5b-fsdp.yaml
```diff
@@ -232,11 +238,11 @@ actor:
     limit_all_gathers: False            # if True, FSDP will synchronizes CPU threads to limit the number of concurrent all-gathers. Only affects strategies that schedule all-gathers
     backward_prefetch: null             # options are null, 'pre', 'post'. if 'pre', FSDP will prefetch the next upcoming all-gather while computing gradients. if 'post', FSDP will prefetch the next all-gather until current gradient is computed.
     use_orig_params: False              # if True, FSDP will use module's original parameters, it means it will expose nn.Module.named_parameters rather than FlatParameter
-    use_liger_kernel: True              # if True, liger_kernel will be used for FSDP, note that currently supported models in RLinf are [qwen2.5, qwen2.5-vl], for more details, see liger_kernel's doc
+    use_liger_kernel: False              # if True, liger_kernel will be used for FSDP, note that currently supported models in RLinf are [qwen2.5, qwen2.5-vl], for more details, see liger_kernel's doc
```
andylin-hao: Every other FSDP e2e config sets `use_liger_kernel: True`, so turning it off here leaves liger + FSDP untested in the new job.

Fix: set it back to `True` if it passes, otherwise add a short comment saying what broke.

rlinf/workers/actor/ma_fsdp_actor_worker.py
```diff
@@ -0,0 +1,664 @@
...
+class MAFSDPActor(FSDPActor):
+    def __init__(
+        self,
+        cfg: DictConfig,
+        placement: ModelParallelComponentPlacement,
+        cfg_fsdp: Optional[DictConfig] = None,
+    ) -> None:
+        super().__init__(cfg, placement, cfg_fsdp)
+        self.is_dynamic_rollout_batch = self.cfg.agentloop.is_dynamic_rollout_batch
+        assert self.is_dynamic_rollout_batch
```
andylin-hao: Small style points for this class: `MAMegatronActor` has a class docstring and this one has none, and this assert has no message while the two below it do.

Fix: add a one-line class docstring and a message here, e.g. `"MAFSDPActor requires agentloop.is_dynamic_rollout_batch=True"`.

andylin-hao (commented):
Re-reviewed at `d3dd930`. The PR has been rescoped since my last pass — the Ascend work is gone, and this is now MA-FSDP actor support (7 files, +1112/-125). The blockers I raised before about `warmup_optimizer_state`, NVML calls, `install.sh`, and missing CI no longer apply.

What looks good: commits are signed off and follow Conventional Commits, `pre-commit` passes, the license header year is right, it merges cleanly with `main`, and the e2e config is wired into the agent workflow. Worth calling out that `retrieve_model_state_dict_in_cpu` on `main` calls `.pin_memory()` on FSDP2 DTensors and loads them back with `load_state_dict`, which cannot have worked — and no FSDP e2e config exercised it, because they all set `kl_beta: 0.0`. This PR fixes that path and adds the first CI job that actually runs it.

Inline comments cover the rest. The main one is that `MAFSDPActor.training_step` is a copy of its parent that doesn't need to exist.

One thing that has no diff line to attach to — **documentation**. `docs/source-en/rst_source/examples/agentic/searchr1.rst` and the matching `docs/source-zh/...` page still describe only `train_qwen2.5.yaml` and `bash examples/agent/searchr1/run_train.sh` (lines 161, 170, 201), and the checkpoint conversion section at lines 215-223 is Megatron-only, which will send FSDP users down the wrong path. The new `train_qwen2.5_fsdp.yaml` and the new `actor.pack_traj` key aren't mentioned anywhere. Both languages need updating, and the "My change requires a change to the documentation" checkbox should be ticked.

rlinf/workers/actor/ma_fsdp_actor_worker.py
qweadzsc: Fixed in 623af9c4. Removed the duplicate training_step() and introduced get_response_loss_mask(); MAFSDPActor now inherits the parent training-step lifecycle and metrics.

rlinf/workers/actor/ma_fsdp_actor_worker.py
qweadzsc: Fixed in 623af9c4. The shared FSDP path now uses kl_penalty(logprobs, ref_logprobs, ...), matching this MA-FSDP call and the Megatron actors.

rlinf/workers/actor/fsdp_actor_worker.py
qweadzsc: Fixed in 623af9c4. FSDP now passes policy logprobs before reference logprobs, consistent with the function signature and all other callers.

rlinf/workers/actor/ma_fsdp_actor_worker.py
qweadzsc: Fixed in 623af9c4. MA-FSDP now always returns the WORLD group for dynamic rollout metrics, so every rank takes the same collective path even for an empty local batch.

rlinf/utils/utils.py
qweadzsc: Fixed in 623af9c4. Added the public FSDPModelManager.load_model_with_state_dict() wrapper and changed the generic helper to use it.

rlinf/utils/utils.py
qweadzsc: Fixed in 623af9c4. FSDP state dicts preserve DTensor metadata and cannot safely reuse the regular-module buffer; passing offloaded_buffer with is_fsdp=True now raises ValueError instead of silently allocating and ignoring it.

rlinf/workers/actor/fsdp_actor_worker.py
qweadzsc: Fixed in 623af9c4. FSDPInference now retrieves the reference state through retrieve_model_state_dict_in_cpu(self, is_fsdp=True).

tests/e2e_tests/agent/searchr1/qwen2.5-1.5b-fsdp.yaml
qweadzsc: Fixed in 623af9c4. The MA-FSDP e2e config now keeps resume_dir: auto.

tests/e2e_tests/agent/searchr1/qwen2.5-1.5b-fsdp.yaml
qweadzsc: Fixed in 623af9c4. The MA-FSDP e2e config now enables use_liger_kernel: True, so the new CI path covers Liger with FSDP.

rlinf/workers/actor/ma_fsdp_actor_worker.py
qweadzsc: Fixed in 623af9c4. Added a class docstring and a descriptive assertion message for is_dynamic_rollout_batch.

examples/agent/searchr1/train.py
qweadzsc: Addressed the late worker-startup failure in 623af9c4: SearchR1 now validates collocated placement before any worker group launches and emits a clear ValueError. I will move this check into the shared validate_cfg path as the final configuration-validation cleanup.

andylin-hao (approved):

guozhen1997 (approved):

PR #1425, opened 2026-08-03
fix: update the logic of reconstruction in RLT

tiny-xie:
### Types of changes

### Checklist:


rlinf/models/embodiment/modules/rlt_token_transformer.py
```diff
@@ -225,55 +223,88 @@ def __init__(
...
-        self.output_proj = (
-            nn.Linear(self.embed_dim, self.input_dim)
-            if self.input_dim != self.embed_dim
-            else nn.Identity()
-        )
+        self.output_proj = nn.Linear(self.embed_dim, self.input_dim)
 
-    def forward(self, rl_tokens: torch.Tensor, target_seq_len: int | None = None):
-        target_seq_len = int(target_seq_len or self.prefix_seq_len)
+    def forward(
```
F9rozen, line 228: `RLTTokenDecoder.forward` is now a teacher-forced AR decoder and changes the public `decode(...)` contract (needs `target_embeddings` instead of `target_seq_len`).

This is incompatible with existing stage-1 RLT checkpoints (removed `prefix_token_embed` / old decoder PE, and `output_proj` is always `nn.Linear` even when `input_dim == embed_dim`, whereas it used to be `Identity`).

Please document clearly that stage-1 must be retrained before stage-2, and that old checkpoints cannot be loaded. If loading old ckpts is still required, add an explicit migration / version guard instead of failing opaquely at `load_state_dict`.

rlinf/models/embodiment/modules/rlt_token_transformer.py
```diff
@@ -225,55 +223,88 @@ def __init__(
                 for _ in range(num_layers)
             ]
         )
-        self.output_proj = (
-            nn.Linear(self.embed_dim, self.input_dim)
-            if self.input_dim != self.embed_dim
-            else nn.Identity()
-        )
+        self.output_proj = nn.Linear(self.embed_dim, self.input_dim)
```
F9rozen, line 226: `output_proj` is now unconditionally `nn.Linear(embed_dim, input_dim)`. Previously it was `Identity` when dims matched (the default 2048/2048 case).

Even ignoring the broader AR rewrite, this alone breaks weight compatibility for default configs. Is the Linear necessary for the AR objective, or can we keep `Identity` when `input_dim == embed_dim`?

rlinf/models/embodiment/modules/rlt_token_transformer.py
```diff
@@ -314,15 +343,19 @@ def encode_flat(
...
-        self, rl_tokens: torch.Tensor, target_seq_len: int | None = None
+        self,
+        rl_tokens: torch.Tensor,
+        target_embeddings: torch.Tensor,
+        mask: torch.Tensor | None = None,
     ) -> torch.Tensor:
-        return self.decoder(rl_tokens, target_seq_len)
+        return self.decoder(rl_tokens, target_embeddings, mask)
 
     def reconstruct(
```
F9rozen, line 353: `reconstruct()` now detaches the prefix and runs teacher-forced decoding. Please add unit tests covering:
1) causal mask: position `t` cannot attend to `> t`
2) loss is masked correctly when `mask` has trailing pads
3) output shape equals `prefix_embs` shape
4) gradients flow into encoder/decoder params but not into the detached teacher targets

This path is the core of the PR and currently has no direct tests.

rlinf/models/embodiment/modules/rlt_token_transformer.py
F9rozen: Docstring still says "a small set of RL tokens", but the module is hard-coded to a single token. Please update the encoder docstring to match.

rlinf/algorithms/rlt/transition.py
```diff
@@ -64,8 +66,26 @@ def update_rlt_transitions(
...
     cache_current: bool,
+    intervene_actions: torch.Tensor | None = None,
+    intervene_flags: torch.Tensor | None = None,
 ) -> None:
     if pending_obs[stage_id] is not None:
+        if intervene_actions is not None and intervene_flags is not None:
+            current_obs = pending_obs[stage_id]
+            ref_chunk = current_obs["ref_chunk"]
+            batch_size = ref_chunk.shape[0]
+            flags = intervene_flags.reshape(batch_size, -1, 1).to(
```
F9rozen, line 77: `intervene_flags.reshape(batch_size, -1, 1)` assumes flags can be broadcast over the action chunk.

If `intervene_flags` is shaped `[B]` (env-level) this only patches `ref_actions[:, :1]`, not the full chunk. If it is `[B, chunk_len]` it works.

rlinf/algorithms/rlt/transition.py
```diff
@@ -64,8 +66,26 @@ def update_rlt_transitions(
...
+                device=ref_chunk.device, dtype=torch.bool
+            )
+            human_actions = intervene_actions.reshape(batch_size, flags.shape[1], -1)
+            action_dim = human_actions.shape[-1]
+            ref_actions = ref_chunk.reshape(batch_size, -1, action_dim).clone()
+            ref_actions[:, : flags.shape[1]] = torch.where(
```
F9rozen, line 83: `torch.where(flags, human_actions, ref_actions[:, : flags.shape[1]])` silently truncates when `flags.shape[1] != ref_chunk` horizon.

If a mismatch is unexpected, please raise; if partial override is intentional (e.g. only executed sub-steps), add a comment explaining why truncating the leading axis is correct.

F9rozen:
Please fix the title typo ("upadte" → "update") and fill in the PR template:
- why AR reconstruction replaces the old decoder
- why single RL token is now mandatory
- checkpoint compatibility (stage-1 retrain required?)
- how intervention→ref_chunk was tested (ManiSkill expert takeover and/or real-world teleop)

guozhen1997 (approved):

andylin-hao (approved):

PR #1430, opened 2026-08-04
fix(scheduler): introduce serialization for PG

Lin-xs:
### Description
Serialize ProcessGroup create/destroy/split within a process to avoid a known torch/pybind race on concurrent GLOO rendezvous / TCPStore.

Changes:
- Add `RLINF_SERIALIZE_PG_LIFECYCLE` (`ClusterEnvVar.SERIALIZE_PG_LIFECYCLE`):
  - unset: enabled for `torch < 2.7.0`, disabled for `torch >= 2.7.0`
  - `1`/`true`/`on`/`yes`: force enable
  - `0`/`false`/`off`/`no`: force disable
- Wrap `MultiChannelProcessGroup._create_process_group` / `_split_process_group` with a process-wide lifecycle lock when serialization is enabled
- Add a unit test that stresses concurrent GLOO PG create/destroy in a subprocess (so a pybind `SIGABRT` is reported via exit code instead of killing pytest)

### Motivation and Context
Concurrent GLOO ProcessGroup create/destroy from multiple threads can abort with:
```text
pybind11_object_dealloc(): Tried to deallocate unregistered instance!
```
This is a known torch/pybind race (see [pytorch#143598](https://github.com/pytorch/pytorch/pull/143598), [pybind11#5473](https://github.com/pybind/pybind11/issues/5473)). Prefer correctness over parallel PG init on older torch builds; keep the opt-out for newer torch and for explicit overrides.

### How has this been tested?
- Added `tests/unit_tests/test_concurrent_pg_init_safety.py`
- Ran locally with torch `2.6.0+cu124`:
  - default behavior (serialization on): test **PASSED**
  - forced `RLINF_SERIALIZE_PG_LIFECYCLE=0`: child aborts and the harness reports a clear failure (expected)

Default policy matches the torch cutoff where the upstream race is expected to be fixed (`>= 2.7.0`).

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/scheduler/cluster/cluster.py
+++ b/rlinf/scheduler/cluster/cluster.py
@@ -100,6 +100,19 @@ class ClusterEnvVar(str, Enum):
     Set explicitly when workers do not share a filesystem with the launch node.
     """
 
+    SERIALIZE_PG_LIFECYCLE = "SERIALIZE_PG_LIFECYCLE"
+    """Serialize ProcessGroup create/destroy within a process.
+
+    Concurrent GLOO rendezvous / TCPStore from multiple threads can abort with
+    ``pybind11_object_dealloc`` on older torch builds. Env FT recovery hits this
+    when many send-queue threads rebuild peer groups at once.
+
+    Values (``RLINF_SERIALIZE_PG_LIFECYCLE``):
+        - Unset: enabled when ``torch < 2.7.0``, disabled when ``torch >= 2.7.0``.
+        - ``1`` / ``true`` / ``on`` / ``yes``: force enable.
+        - ``0`` / ``false`` / ``off`` / ``no``: force disable.
+    """
+
 
 class PathEnvMergeMode(str, Enum):
     """Merge mode for path-like worker env vars."""
@@ -128,6 +141,7 @@ class Cluster:
         ClusterEnvVar.EXT_MODULE: None,
         ClusterEnvVar.PATH_ENV_MERGE_MODE: PathEnvMergeMode.APPEND.value,
         ClusterEnvVar.CODE_WORKING_DIR: "0",
+        ClusterEnvVar.SERIALIZE_PG_LIFECYCLE: None,
     }
     PATH_LIKE_ENV_VARS = {
         "PYTHONPATH",
--- a/rlinf/scheduler/collective/multi_channel_pg.py
+++ b/rlinf/scheduler/collective/multi_channel_pg.py
@@ -13,11 +13,14 @@
 # limitations under the License.
 
 import logging
+import threading
+from contextlib import nullcontext
 from datetime import timedelta
 from typing import Optional
 
 import torch
 import torch.distributed as dist
+from packaging.version import parse as parse_version
 
 from ..hardware import AcceleratorType, AcceleratorUtil
 from .async_work import AsyncCollWork, AsyncWork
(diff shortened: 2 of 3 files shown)


andylin-hao:
Thanks for digging into this — the diagnosis is correct, and the upstream reference is the right one: [pytorch/pytorch#143598](https://github.com/pytorch/pytorch/pull/143598) ("c10d: no call_guard in init", merged 2024-12-20) fixes exactly this `pybind11_object_dealloc` abort on concurrent TCPStore construction/teardown, via pybind11#5473.

Before this lands, though, there's one blocking correctness issue I'd like to flag.

## The process-wide lock is held across a blocking rendezvous, which introduces a cross-process circular wait

`_create_process_group` blocks in `rendezvous()` / TCPStore until the peer rank joins (`rlinf/scheduler/collective/multi_channel_pg.py:499-563` on `main`). The groups being serialized are **not** built eagerly in a globally consistent order — they are created lazily, per peer-pair, on the `CollectiveWorkQueue` daemon threads (`collective_group.py:135`) from inside the send/recv/broadcast hot paths:

- `collective_group.py:352` (`_atomic_send`), `:464`, `:553`, `:675`
- per-pair sub-groups via `_get_or_create_ipc_sub_group(...)._init_process_group(...)` at `:824`, `:838`, `:941-1022`

With a single process-wide lock and no global ordering over which pair gets initialized first, this becomes reachable:

- worker A holds its lock, blocked in rendezvous waiting for B
- worker B holds its lock, blocked waiting for C
- worker C holds its lock, blocked waiting for A

Every partner thread that could satisfy one of those waits is itself queued behind its own process's lock. Prior to this change those inits proceeded concurrently, so no cycle was possible — the lock is what creates it.

The broadcast path makes it concrete: at `:821-824` the source rank walks every receiver **sequentially in one thread**, holding the lock through each pair's rendezvous, while each receiver only initializes its own pair.

This also does not fail fast. The rendezvous timeout is `RLINF_TIMEOUT`, which defaults to 180 *minutes* (`multi_channel_pg.py:134-141`), so the symptom is a multi-hour hang rather than a crash — a worse failure mode than the one being fixed.

### Suggested direction

Don't hold the lock across the network wait. Keying the lock per `init_method` / group name would let independent peer pairs initialize in parallel while still serializing churn against the same store, which is what pybind11#5473 actually concerns. If a genuinely process-wide lock turns out to be necessary, it should at least take a bounded acquire timeout that raises with a clear message instead of inheriting the 180-minute one.

### Why the test wouldn't have caught it

`test_concurrent_pg_init_safety.py` uses `world_size=1` with a private port per thread, so no thread ever waits on another — structurally the one configuration that cannot surface this. A ≥3-process cyclic first-touch scenario would.

Two smaller notes on the test while I'm here, both worth carrying into any future version of this work:

- `_stress_concurrent_create_process_group` returns `{"ok": True, ...}` where `"ok"` is a hardcoded literal, and the test asserts only `results["ok"] is True`. `errors` / `num_errors` are collected and returned but never checked, so every one of the 512 PG creations could raise and the test would still pass. `assert results["num_errors"] == 0, results["errors"]` would fix it.
- Nothing sets `RLINF_SERIALIZE_PG_LIFECYCLE`, so the child inherits the version-derived default. Since the repo pins `torch==2.11.0` (`pyproject.toml:151`), serialization is **off** in CI and the test exercises the unlocked path.

## Closing

We're standardizing on torch 2.11 (#1410), and pytorch#143598 is in well before that, so the underlying race is fixed upstream for every version we now support. With the version-derived default this PR would be inert on the default install anyway, which leaves the deadlock risk above as the only behavior change users could actually observe.

Closing on that basis rather than because the analysis was wrong — it was accurate and the writeup was genuinely useful. If we do end up needing to support torch < 2.7.0 again, this is a good starting point, and the per-`init_method` locking variant would be the thing to build on.

PR #1416, opened 2026-07-28
feat: adapt rlinf to sgl 0.5.12, cu13 and tf 5.6

sherlockcooper:
### Description

Companion to #1413 (which makes the sglang 0.5.12 + cu13 stack installable). This PR makes RLinf's sglang rollout, tokenizer, dataset and patcher code load and run on that stack. Each change is a dual-version fallback for one upstream API break; the 0.4.6 / cu12 / transformers 4.x path is unchanged.

- **sglang_worker.py** — sglang >=0.5.11 dropped the `enable_ep_moe` ServerArgs flag; EP is now enabled by `ep_size > 1`. Keep the `rollout.sglang.enable_ep_moe` config as the EP switch and translate it to `ep_size`.
- **fsdp_model_manager.py**, **vlm_reward_model.py** — transformers >=5.0 renamed `AutoModelForVision2Seq` to `AutoModelForImageTextToText` (old name removed); import with try/except for both versions.
- **workers/rollout/sglang/__init__.py** — raise the supported-sglang cap from 0.5.4 to 0.5.13 so 0.5.12.post1 is accepted.
- **tokenizer_manager.py** — `_Communicator` moved across sglang versions (0.4.x `tokenizer_manager` -> 0.5.x `tokenizer_communicator_mixin` -> 0.5.12 `communicator.FanOutCommunicator`); three-layer import fallback.
- **tokenizer_manager.py** — sglang 0.5.x renamed `Batch*Out` to `Batch*Output` (moved to `io_struct`) and relocated `logger`; import with fallback to the 0.4.x names.
- **tokenizer_manager.py**, **sgl_scheduler.py** — sglang 0.5.x made `TypeBasedDispatcher._mapping` an OrderedDict (was a list), so `.extend(...)` breaks; use dict assignment when it is a dict, else `.extend` for 0.4.x.
- **utils/patcher.py** — on the cu13 stack, nvidia-cutlass-dsl's `align` class raises in `__hash__` (references a missing `_value`), breaking the `cls in set` dedup; dedup traced classes by `id(cls)`.
- **data/datasets/reasoning.py**, **rstar2.py** — transformers >=5.0 (`TokenizersBackend`) removed `batch_encode_plus`; fall back to `tokenizer(...)` (`__call__`), which works on 4.x and 5.x.

### Motivation and Context

[...]

PR #1413, opened 2026-07-27
feat: adapt install to sglang 0.5.12 cu13

sherlockcooper:
### Description

Add a `sglang >=0.5.11` install path (torch 2.11+cu130) to `requirements/install.sh`, gated by a new `_sglang_needs_cu13` helper. The existing 0.4.x / cu12 path is unchanged.

All changes are gated on `_sglang_needs_cu13` (i.e. sglang >=0.5.11):

- **configure_nvidia**: route torch/torchvision/torchaudio through the cu130 PyTorch wheel index so `uv sync` can resolve cu13 wheels.
- **apply_sglang_override**: auto-derive `xgrammar` versions for sglang 0.5.5-0.5.12 (previously only 0.4.6-0.5.4).
- **install_flash_attn**: mirror flash-attn 2.8.3 wheels from the RLinf fork (cu12 + cu13 tagged wheels); set `FLASH_ATTENTION_FORCE_BUILD=TRUE` on the source fallback to skip flash-attn `setup.py`'s own GitHub wheel download (it has no timeout and hangs on networks that need a proxy).
- **install_agentic** (0.5.x branch): `--prerelease=allow`, pin `kernels<0.13`, switch mcore to `core_r0.17.0`, skip `megatron.txt` to avoid the TE 2.1.0 conflict, then source-build TE 2.17 (`core_cu13`) and install `rlinf-megatron-bridge==0.4.2` (a py3.10/3.11 fork of `megatron-bridge`, published on PyPI under a separate name).
- **uninstall_fa4_conditional**: drop flash-attn-4 on sm<9 so TE falls back to FA2 (sm80 fwd+bwd supported).
- **generate_nccl_env_script**: prepend the venv's cu13 nvidia libs to `LD_LIBRARY_PATH` (TE 2.17 `.so` has no RPATH); the path is resolved via `import nvidia`, so it is Python-version agnostic.

### Motivation and Context

[...]

PR #1437, opened 2026-08-05
feat: add DeepseekV3/Kimi-K2 MoE architecture support

sherlockcooper:
### Description

Add MoE model support for DeepseekV3 and Moonlight-16B-A3B architecture in the Megatron-actor + sglang-rollout hybrid engine. This is the model/MoE layer on top of the sglang 0.5.12 runtime adaptation (#1416) and the torch-2.11/cu13 install (#1413).

- **convertor/utils** — `DeepseekV3Convertor` support.
- **resharding/utils, reshard_config** — tp/tpe/ep/pp reshard fns for DeepseekV3 MLA + MoE; dispatch by `model_type`.
- **mcore_weight_reshard** — shared_experts support + EP/TP branch.
- **megatron_model_manager** — `sequence_parallel` override after finalize; mbridge MoE override whitelist; resume training with precision aware optimizer fixes.
- **logging** — configure third party megatron core logging level.
- **config** — register `DEEPSEEK_V3`.
- **e2e** — Moonlight test config + official example + CI job.

### Motivation and Context

RLinf's hybrid engine needs MoE support for DeepseekV3-architecture models (e.g. Moonlight-16B-A3B). This PR adds the weight convertor, EP/TP/PP reshard, and model-specific fixes (sequence_parallel, w2 scales, resume training with precision aware optimizer) to run them end-to-end.

No linked issue.

### How has this been tested?

- DeepseekV3 convertor verified on Kimi K2.5 architecture.
- Moonlight-16B-A3B: RL e2e (train + ckpt + clean exit) on sglang 0.5.12.post1 + torch 2.11+cu130 + TE 2.17 + transformers 5.6.
- The install + runtime sglang-0.5.12 adaptation is in main via #1413/#1416; full CI e2e for the MoE path needs GPUs/docker — happy to coordinate with maintainers.

### Types of changes

### Checklist

- [x] My code follows the code style of this project. (`ruff` + `ruff-format` pass on all changed files.)
- [ ] I have added tests to cover my changes. (Moonlight e2e config included; full CI e2e needs GPUs/docker — manual verification above.)
- [ ] All new and existing tests passed. (Manual e2e passed; CI runs via the `run-ci` label.)


rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
```diff
@@ -240,7 +254,6 @@ def sync_hf_weight(self, recv_req: SyncHFWeightInput):
                 )
             self.weight_norm_dict = None
 
-        self.flush_cache()
```
andylin-hao, line 243: Move the flush cache back here

rlinf/utils/logging.py
```diff
@@ -13,13 +13,55 @@
...
+    "megatron.core.distributed.distributed_data_parallel",
+    "megatron.core.optimizer",
+    "megatron.core.optimizer_param_scheduler",
+    "megatron.core.num_microbatches_calculator",
+)
+
+_THIRD_PARTY_LOGGING_CONFIGURED = False
+
+
+def configure_third_party_logging() -> None:
```
andylin-hao, line 38: There is a RLINF_LOG_LEVEL, RLINF_DBG_LEVEL duplicates it and is undocumented

.github/workflows/agent-e2e-tests.yml
```diff
@@ -747,3 +747,34 @@ jobs:
...
+        uses: actions/checkout@v5
+
+      - name: Create reason environment
+        run: |
+          unset UV_DEFAULT_INDEX
+          export UV_PATH=/workspace/dataset/.uv
+          export UV_LINK_MODE=symlink
+          export UV_CACHE_DIR=/workspace/dataset/.uv_cache
+          export UV_PYTHON_INSTALL_DIR=/workspace/dataset/.uv_python
+          bash requirements/install.sh agentic --sglang 0.5.12 --torch 2.11.0 --no-apex
```
andylin-hao: No need for --sglang 0.5.12 and --torch 2.11.0 and --no-apex now

rlinf/hybrid_engines/sglang/common/sgl_scheduler.py
```diff
@@ -226,6 +226,20 @@ def sync_hf_weight(self, recv_req: SyncHFWeightInput):
...
+        restore_model = self.tp_worker.worker.model_runner.model
+        for restore_layer in restore_model.model.layers:
+            restore_experts = getattr(restore_layer.mlp, "experts", None)
+            if restore_experts is not None:
+                for restore_scale_name in ("w2_input_scale", "w2_weight_scale"):
+                    restore_scale_param = getattr(
+                        restore_experts, restore_scale_name, None
+                    )
+                    if restore_scale_param is not None:
+                        restore_scale_param.fill_(1.0)
```
andylin-hao: This block runs inside `sync_hf_weight`, so it executes on **every weight sync for every SGLang rollout**, not just DeepSeek-V3. Three concerns:

1. **No version guard.** The comment says this is "only used in sglang 0.4.6", but nothing gates it — and the PR reports testing on 0.5.12.

2. **It can clobber real quantization state.** Any MoE model whose experts expose `w2_input_scale` / `w2_weight_scale` gets them forced to `1.0`. For a genuinely FP8-quantized MoE served by SGLang that silently discards the scales — no error, just wrong rollout outputs.

3. **`restore_model.model.layers` is unguarded.** Dense models are safe thanks to the `getattr(..., "experts", None)` check, but the attribute chain itself is not — an architecture without `.model.layers` raises `AttributeError` here, on the shared sync path.

Suggestion: gate on the sglang version the comment names and on the model types this PR adds, e.g.

```python
if sglang_version < parse("0.5.0") and model_type in ("deepseek_v3", "kimi_k2"):
    ...
```

If it is in fact still needed on >=0.5.x, could the comment say why it is safe to apply unconditionally?

rlinf/hybrid_engines/megatron/megatron_model_manager.py
```diff
@@ -340,6 +378,9 @@ def setup_mbridge_model_and_optimizer(self):
 
         provider.finalize()
         self.provider = provider
+        # Override sequence_parallel after finalize()
+        # required by moe_layer.py:547 validation when MoE+TP
+        provider.sequence_parallel = self.transformer_config.sequence_parallel
```
andylin-hao: For `deepseek_v3` / `kimi_k2` this line can never change anything.

`provider_field_names` is every dataclass field of the provider (L298), so `sequence_parallel` is in it, and it is not in `_mbridge_user_override_fields` (L307-317). The loop at L318-323 therefore takes the `else` branch for it and **raises** when config and provider disagree — which is exactly the case this override exists to handle. If they agree, this line is a no-op.

So either it is redundant, or the code raised before reaching it. The error message also tells the user to set `sequence_parallel: <provider_val>`, pointing them at a field this line would have overridden anyway.

Suggestion: add `"sequence_parallel"` to `_mbridge_user_override_fields` and drop this line, or move the override above the validation loop.

andylin-hao (commented):
Two correctness points on the shared (non-MoE) paths — details inline. The MoE-specific work looks well scoped: the EP/TP reshard gating, shared-expert exclusion, and the mbridge whitelist all leave dense and existing `qwen3_moe` paths unchanged.

andylin-hao (approved):

rlinf/hybrid_engines/megatron/megatron_model_manager.py
```diff
@@ -973,6 +1012,16 @@ def _iter_opts(opt):
                     cpu_data = self._get_pinned_buffer(buffer)
                     cpu_data.copy_(buffer.data, non_blocking=True)
                     buffer.storage().resize_(0)
+                # offload master_param too
```
Lin-xs: Consider merging these three if branches into one.

rlinf/utils/resharding/utils.py
```diff
@@ -27,6 +27,8 @@ def get_tp_reshard_fn(model_type: str):
         return tp_reshard_fn_qwen3_dense
     elif model_type == SupportedModel.QWEN3_MOE:
         return tp_reshard_fn_qwen3_moe
+    elif model_type == SupportedModel.DEEPSEEK_V3:
```
Lin-xs: I noticed that `KIMI_K2` was added to `SupportedMode`l, but it doesn't seem to be supported here. Is this intentional?

rlinf/utils/resharding/mcore_weight_reshard.py
```diff
@@ -266,14 +281,26 @@ def get_expert_num(param_name):
...
-                    tpe_group,
-                    self.config.reshard_tp_size,
-                    dst_tp_rank,
-                )
+                # reshard experts for the rollout layout.
+                #   ep_reshard_fn selects this rank's slice by global expert id.
+                if (
+                    self.config.rollout_ep_size > 1
+                    and self.config.ep_reshard_fn is not None
+                ):
```
Lin-xs, line 289: Consider adding an assertion for `tpe_size == 1` here.

rlinf/hybrid_engines/megatron/megatron_model_manager.py
sherlockcooper: Done — merged into a single loop; mirrors the onload path which already iterates these three keys.

rlinf/utils/resharding/utils.py
sherlockcooper: KIMI_K2 shares the DeepSeek-V3 architecture; routed it to the same reshard fns in all four dispatch functions instead of raising.

rlinf/utils/resharding/mcore_weight_reshard.py
sherlockcooper: Added

rlinf/utils/convertor/utils.py
```diff
@@ -621,11 +621,198 @@ def build_rules(self) -> list[ConvertorRule]:
...
+            converted = {**kv, **rest, **qa}
+        return converted
+
+
 _MG2HF_CONVERTOR_REGISTRY = {
     SupportedModel.QWEN2_5: Qwen25Convertor,
     SupportedModel.QWEN2_5_VL: Qwen25VLConvertor,
     SupportedModel.QWEN3: Qwen3DenseConvertor,
     SupportedModel.QWEN3_MOE: Qwen3MoEConvertor,
+    SupportedModel.DEEPSEEK_V3: DeepseekV3Convertor,
```
Lin-xs, line 815: Is KIMI_K2 supported here?

Lin-xs (approved):

guozhen1997 (approved):

PR #1447, opened 2026-08-06
docs: stop explaining what the install script does internally

andylin-hao:
## Description

Several recipe pages narrate the installer's implementation next to the command that runs it — which fork it clones, which directory under `.venv` the source lands in, that it "handles dependency isolation" or "all high-risk dependency conflicts".

A reader running `bash requirements/install.sh …` does not need any of that, and it rots: these same pages described LIBERO as a git clone into `.venv/libero` long after it became the `rlinf-libero` package (fixed separately in #1444).

Removed from the Evo-1, Lingbot-VLA and DreamZero SFT recipes, in both languages. What stays is anything the reader has to *act* on:

- Evo-1 — pointing `actor.model.evo1.repo_path` / `EVO1_REPO_PATH` at a manual checkout
- DreamZero SFT — setting `DREAMZERO_PATH`

## Type of Change

- [x] Documentation

## Testing

Both doc trees build with the same warning count as `main`.

## Checklist

- [x] Commit is signed off
- [x] Docs updated in both languages
- [x] No new build warnings

🤖 Generated with [Claude Code](https://claude.com/claude-code)


andylin-hao:
Folded into #1446 — same commit, no separate PR needed.

PR #1453, opened 2026-08-07
feat(docker): add franky venvs to the franka image

andylin-hao:
### Description

Adds `franky-*` venvs to the `embodied-franka-image` Docker stage.

`franka-franky` has been a supported env in `requirements/install.sh` (`SUPPORTED_ENVS`, plus `install_franka_franky_env` and `requirements/embodied/franky_install.sh`) but had no coverage in `docker/Dockerfile`, so the published Franka image shipped without it. The install-check skill flags this as a Req-7 gap:

```
env 'franka-franky' has no '--env franka-franky' install in docker/Dockerfile
```

The new block builds `franky-0.15.0` and `franky-0.19.0` alongside the existing `franka-$v` venvs. It intentionally does **not** reuse the `franka-$v` loop:

- `franka-franky` routes through `install_env_only`, which skips the ROS/catkin build entirely — `franky-control` ships libfranka inside a prebuilt wheel — so `FRANKA_ROS_VERSION` is not applicable.
- Wheels are published for only two libfranka versions, not the six the ROS-based venvs cover.

The default venv in `~/.bashrc` is unchanged (`franka-0.15.0`).

### Motivation and Context

Closes a Docker coverage gap for an env that already exists in the install script. Without this, anyone using the Franka image has to install franky by hand.

### How has this been tested?

- `bash .claude/skills/install-check/check.sh` — `franka-franky` no longer appears under the "Docker — models/envs missing from docker/Dockerfile" section. Remaining entries in that list are pre-existing and untouched here.
- `bash -n requirements/install.sh` — passes (unchanged, checked because the new line depends on its arg handling).
- Verified the dispatch path by reading `install.sh`: `--env franka-franky` with no `--model` sets `MODEL=""` → `install_env_only` → the `franka-franky` branch. There is no parse-time allowlist check that would reject it.
- Verified wheel availability against https://github.com/Brunch-Life/franky/releases — only `wheels-libfranka-0.15.0` and `wheels-libfranka-0.19.0` exist, which is why the loop covers exactly those two. Both publish cp39–cp314; the venv default is Python 3.11 (`PYTHON_VERSION=3.11.14`) → `cp311`, and `install_franka_franky_env` requests `manylinux_2_28`, which is satisfied by Ubuntu 20.04's glibc 2.31.

**Not tested:** the image was not actually built — no Docker daemon on the machine I had available. The `build-embodied-franka` CI job covers this stage and will exercise the new block; please treat that job as the gate before merging.

### Types of changes

### Checklist:

Docs were left alone deliberately: this adds no new env, config key, or user-facing name — `franka-franky` is already documented as an install target — it only makes the prebuilt image match what `install.sh` already supports. Existing CI (`build-embodied-franka`) covers the stage, so no new test was added.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -333,6 +333,14 @@ RUN for v in 0.10.0 0.13.3 0.14.1 0.15.0 0.18.0 0.19.0; do \
             LIBFRANKA_VERSION="$v" FRANKA_ROS_VERSION=0.10.0 bash requirements/install.sh ${INSTALL_MIRROR_OPTION} --platform nvidia embodied --venv "franka-$v" --env franka; \
         done
 
+# Franky environments. franky-control ships libfranka inside a prebuilt wheel,
+# so these need no ROS/catkin build and are kept in separate venvs from the
+# franka-$v ones above. Only the libfranka versions with published wheels are
+# built: https://github.com/Brunch-Life/franky/releases
+RUN for v in 0.15.0 0.19.0; do \
+            LIBFRANKA_VERSION="$v" bash requirements/install.sh ${INSTALL_MIRROR_OPTION} --platform nvidia embodied --venv "franky-$v" --env franka-franky; \
+        done
+
 # Set default env (recommended libfranka 0.15.0)
 RUN echo "source ${UV_PATH}/franka-0.15.0/bin/activate" >> ~/.bashrc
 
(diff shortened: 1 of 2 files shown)


andylin-hao:
Merged into #1444 (commit `a8856fb0`), which is where the franka image work now lives — it also carries a `.dockerignore` and finishes the `ghfast.top` → `gh-proxy.com` switch this PR started (the asset downloader, the FUSCO e2e step, and the docs still hardcoded the old host).

Closing in favour of #1444; nothing here is lost.

PR #1456, opened 2026-08-07
fix(async): safely overlap rollout weight sync

yicwang:
### Description
`actor.sync_weight_no_wait` previously affected only `AsyncEmbodiedRunner`.
`AsyncPPOEmbodiedRunner` is a sibling of that runner and always used the
blocking `EmbodiedRunner.update_rollout_weights()`, so the existing switch had
no effect for async PPO.

This change:

- adds an `AsyncWeightSyncMixin` shared by both async embodied runners;
- makes a background-sync handle complete only after rollout-side
  `WeightSyncer.apply(...)` has finished, rather than after the rollout worker
  merely acknowledges the request;
- keeps the initial actor-to-rollout sync blocking, drains an in-flight sync
  before a later blocking sync and normal teardown, and coalesces requests
  while one runner-level sync is in flight;
- permits no-wait synchronization only with the patch syncer. Bucket apply can
  yield between buckets, so allowing inference concurrently could expose a
  partially updated model;
- documents the behavior in English and Chinese and enables the no-wait path
  in the existing async-PPO ManiSkill e2e config.

The default remains `actor.sync_weight_no_wait=false`.

### Motivation and Context
Async PPO could not overlap actor-to-rollout weight transfer even though it
already exposed the same configuration as the other async embodied runner.
While sharing the implementation, two correctness gaps in the old async path
also needed to be closed: blocking transitions could race an outstanding
background apply, and normal teardown could stop workers before that apply had
finished.

Tracking rollout-side application is important for data correctness because
the applied model version is recorded on generated trajectories and used by
async PPO's staleness filtering.

### How has this been tested?
- `pytest -q tests/unit_tests/test_async_weight_sync.py`: **19 passed**. This
  covers both runner MROs, blocking/non-blocking behavior, runner and rollout
  coalescing, rollout-side apply completion, blocking and teardown drains,
  blocking startup, and the patch-only guard.
- selected patch/bucket round-trip cases from
  `tests/unit_tests/test_weight_syncer.py`: **5 passed**.
- `pre-commit run --files <changed files>`: Ruff lint, Ruff format, and
  committer signoff all passed.

GPU smoke test: one `rlinf-0` node with 4x NVIDIA H20 GPUs available; collocated
actor/env/rollout placement used GPU ranks 0-1. Pi0.5 + ManiSkill async PPO,
patch syncer, 4 environments, seed 0, 6 optimizer steps. The same config was
run with only `actor.sync_weight_no_wait` changed:

| | `false` | `true` |
| --- | ---: | ---: |
| Exit status | 0 | 0 |
| Actor versions | 1 through 6 | 1 through 6 |
| Maximum trajectory staleness observed | 2 | 2 |
| Discarded unused trajectories | 0 | 0 |
| Finite TensorBoard scalars | 274 / 274 | 285 / 285 |
| Mean runner-visible `update_rollout_weights` time | 2.621 s | 0.648 ms |

The no-wait run consumed trajectories tagged with rollout-applied versions
2-5 after initially consuming version 0, showing that applied versions—not
only actor-side counters—propagate into training data. This short run checks
the synchronization and staleness invariants plus numerical health; it is not
a convergence comparison.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/runners/async_embodied_runner.py
+++ b/rlinf/runners/async_embodied_runner.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, Union
 
 from omegaconf.dictconfig import DictConfig
 
+from rlinf.runners.async_weight_sync_mixin import AsyncWeightSyncMixin
 from rlinf.runners.embodied_runner import EmbodiedRunner
 from rlinf.scheduler import Channel
 from rlinf.scheduler import WorkerGroupFuncResult as Handle
@@ -38,7 +39,7 @@ if TYPE_CHECKING:
     )
 
 
-class AsyncEmbodiedRunner(EmbodiedRunner):
+class AsyncEmbodiedRunner(AsyncWeightSyncMixin, EmbodiedRunner):
     def __init__(
         self,
         cfg: DictConfig,
@@ -54,10 +55,7 @@ class AsyncEmbodiedRunner(EmbodiedRunner):
         self.env_metric_channel = Channel.create("EnvMetric")
         self.rollout_metric_channel = Channel.create("RolloutMetric")
 
-        self._pending_rollout_weight_sync = None
-        self._weight_sync_coalesced_total = 0
-        self._weight_sync_request_total = 0
-        self.sync_weight_no_wait = self.cfg.actor.get("sync_weight_no_wait", False)
+        self.init_weight_sync_state()
 
     def get_env_metrics(self) -> tuple[dict, list[dict], list[dict]]:
         results: list[dict] = []
@@ -103,39 +101,6 @@ class AsyncEmbodiedRunner(EmbodiedRunner):
         )
         return time_metrics, ranked_time_metrics_list
 
-    def _cleanup_pending_rollout_weight_sync(self, no_wait):
-        if self._pending_rollout_weight_sync is None:
-            return True
-
-        rollout_handle, actor_handle = self._pending_rollout_weight_sync
-        self.logger.info(
-            f"Rollout handle done: {rollout_handle.done()}, actor handle done: {actor_handle.done()}"
-        )
-        if no_wait and (not rollout_handle.done() or not actor_handle.done()):
-            return False
-
-        rollout_handle.wait()
-        actor_handle.wait()
-        self._pending_rollout_weight_sync = None
-        return True
-
-    def update_rollout_weights(self, no_wait=False):
-        if not no_wait:
-            return super().update_rollout_weights()
-
-        self._weight_sync_request_total += 1
-        if not self._cleanup_pending_rollout_weight_sync(no_wait):
-            self._weight_sync_coalesced_total += 1
-            self.logger.info(
-                f"Weight sync coalesced {self._weight_sync_coalesced_total} times.\n"
-                f"Request total {self._weight_sync_request_total} times."
-            )
-            return
-
-        rollout_handle: Handle = self.rollout.request_actor_sync_model()
-        actor_handle: Handle = self.actor.sync_model_to_rollout()
-        self._pending_rollout_weight_sync = (rollout_handle, actor_handle)
-
     def evaluate(self):
         env_handle: Handle = self.env.evaluate(
             input_channel=self.env_channel,
@@ -157,7 +122,9 @@ class AsyncEmbodiedRunner(EmbodiedRunner):
     def run(self):
         start_step = self.global_step
         start_time = time.time()
-        self.update_rollout_weights(no_wait=self.sync_weight_no_wait)
+        # Rollout must start from the actor's exact initial/resumed weights, so
+        # this first sync always blocks even when overlap is enabled.
+        self.update_rollout_weights()
 
         env_handle: Handle = self.env.interact(
             input_channel=self.env_channel,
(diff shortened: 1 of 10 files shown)


rlinf/runners/async_embodied_runner.py
```diff
@@ -308,6 +273,9 @@ def run(self):
             if profiled_step is not None:
                 self._close_profiling_window(profiled_step)
 
+        # Let any in-flight non-blocking sync land before the workers go away.
+        self.drain_pending_rollout_weight_sync()
```
guozhen1997, line 279: Drain waits on `request_actor_sync_model` + `sync_model_to_rollout` handles. `request_actor_sync_model` returns after setting a flag, and actor `sync()` returns after send — neither waits for `WeightSyncer.apply()`. So this does not actually wait until rollout weights are applied. Please complete the rollout handle only after apply finishes.

rlinf/runners/async_embodied_runner.py
guozhen1997: Commit says the initial sync stays blocking. PPO does that; this still passes `no_wait=self.sync_weight_no_wait`. Please call `self.update_rollout_weights()` here (blocking), and only use `no_wait` inside the loop.

rlinf/runners/async_ppo_embodied_runner.py
```diff
@@ -159,7 +156,7 @@ def run(self) -> None:
                 self.global_step += 1
                 self.actor.set_global_step(self.global_step).wait()
                 with self.timer("update_rollout_weights"):
-                    self.update_rollout_weights()
+                    self.update_rollout_weights(no_wait=self.sync_weight_no_wait)
```
guozhen1997, line 159: This is what first honors the flag for async PPO. The PR text says no-wait should be patch-only (bucket apply can yield a half-updated model). Please guard that here or in the rollout worker before recommending this path. Trajectory versions are updated only after apply.

rlinf/runners/async_embodied_runner.py
yicwang: request_actor_sync_model now keeps its coroutine alive until the rollout worker has applied the weights, so the handle is a real completion signal.

rlinf/runners/async_embodied_runner.py
yicwang: Fixed.

rlinf/runners/async_ppo_embodied_runner.py
yicwang: Added the guard, but in AsyncWeightSyncMixin.init_weight_sync_state rather than at the call site, so it fails at startup instead of at the first sync. Happy to move it into the rollout worker if you'd prefer it closer to the apply path.

rlinf/runners/async_weight_sync.py
```diff
@@ -0,0 +1,115 @@
+# Copyright 2025 The RLinf Authors.
```
anHappyDog: shoud change to 2026.

rlinf/runners/async_weight_sync.py
```diff
@@ -0,0 +1,115 @@
...
+        # the same config key.
+        self.sync_weight_no_wait = self.cfg.actor.get("sync_weight_no_wait", False)
+        if self.sync_weight_no_wait:
+            weight_syncer_type = self.cfg.get("weight_syncer", {}).get("type", None)
+            if weight_syncer_type != "patch":
+                raise ValueError(
+                    "actor.sync_weight_no_wait=true requires "
+                    "weight_syncer.type=patch so a rollout cannot observe a "
+                    "partially applied bucket sync."
+                )
```
anHappyDog: it might be better to be put into `validate_embodied_cfg`.

rlinf/runners/async_ppo_embodied_runner.py
```diff
@@ -159,7 +156,7 @@ def run(self) -> None:
...
                 self.rollout.set_global_step(self.global_step).wait()
```
anHappyDog: if `sync_weight_no_wait` is True, rollout might set newer global step before rollout-side weight application finishes. Maybe it's better to delete this line here.

rlinf/runners/async_weight_sync.py
anHappyDog: the file could be renamed to `async_weight_sync_mixin.py`？

yicwang: Fixed

rlinf/runners/async_ppo_embodied_runner.py
yicwang: Fixed.

rlinf/runners/async_weight_sync.py
yicwang: Added the call to the end of validate_embodied_cfg().

anHappyDog (approved):

guozhen1997 (approved):

guozhen1997:
Hi @yicwang, could you please rebase this PR onto the latest main branch?

yicwang:
> Hi @yicwang, could you please rebase this PR onto the latest main branch?

Done

guozhen1997:
Hi @yicwang, could you please fix the filter coverage test?
```
The following changed files are not covered by any CI filter:
rlinf/runners/async_weight_sync_mixin.py
```

andylin-hao (approved):

PR #1457, opened 2026-08-08
feat(franka): make Franky the default setup

lizuojun04:
### Description

Make Franky the default Franka backend, so one GPU host can run robot control, rollout, and training without ROS, and install robot dependencies from requirements files.

**Behavior change:** a Franka config without a `backend` key now runs Franky, `--env franka` installs Franky instead of ROS Noetic, and `gripper_type: franka` resolves to the Franky gripper. ROS deployments must set `backend: franka_ros` and install with `--env franka-ros`.

- `FrankaRobot.BACKEND` is `franky`; `--env franka` installs the prebuilt franky-control wheel, and ROS Noetic moves to `--env franka-ros`.
- Franky and DualFranka configs take `realtime_config`: `ignore` (default) runs without PREEMPT_RT and warns, `enforce` refuses. The ROS backend uses `FRANKA_REALTIME_CONFIG` at install time.
- The `embodied-franka` image uses the standard platform base (CUDA 12.8, Ubuntu 22.04) and contains the default `franky` venv and `openvla`, `openvla-oft`, `openpi`, and `gr00t` model venvs, all built with `--env franka`, so one container runs the actor, rollout, and robot control. It no longer builds ROS Noetic venvs, which need Ubuntu 20.04.
- `--env franka`, alone or with `--model`, also installs `RLinf-dexterous-hands[glove]`; `--env franka-dexhand` is removed.
- The installer retries torch index and `git clone` fetches from mirrors instead of silently picking an older CUDA wheel or failing on a transient `gh-proxy.com` error.
- `--model openvla|openvla-oft|openpi|gr00t --env franka` installs model venvs with Franka deps.
- Robot extras move to `requirements/embodied/envs/*.txt`; env-only robot venvs keep the `embodied` extra.
- The Franky installer fails early when no prebuilt wheel matches the host or `LIBFRANKA_VERSION`.
- `test_franka_controller.py` gains `--backend`, `--realtime-config`, and `getjoint`/`clear`/`home`/`open`/`close`; `test_franky_controller.py` is removed.
- `collect_data.sh` and `run_realworld_async.sh` forward Hydra overrides and honor `RLINF_LOG_DIR`.
- `RobotTask` and `RobotTaskEnv` are removed: every real-world task is a `gym.Env` registered in its robot package, per the New Real-World Tasks guide.
- **Real-World RL with Franka** covers a single GPU host with Franky, an optional real-time kernel, and a legacy ROS section on installing and running ROS; related pages follow.

### Motivation and Context

Running a Franka beside training on one GPU host should not require ROS Noetic or a separate controller machine. Robot and model dependencies should compose through the installer without resyncing away model packages.

### How has this been tested?

- `pytest tests/unit_tests/test_real_env.py`: 234 passed.
- `pytest tests/unit_tests/test_robotics.py`: 288 passed, 2 failed. `test_custom_driver_composes_with_each_franka_arm[franka_ros|franky]` fails identically on the earlier head 4bc05f969, because the Ray worker in this sandbox cannot import `test_robotics`.
- Mutation check: restoring `enforce` as the Franky default, or removing the `franka_ros` refusal of `realtime_config`, fails the new tests.
- Installer functions run against a stubbed `uv`: env-only order for `franka`, `franka-dexhand` and `franka-ros` with `SKIP_ROS`/`NO_ROOT`, model paths for `openvla` and `gr00t`, the wheel guard, and `FRANKA_REALTIME_CONFIG` applied to the real `franka_control_node.yaml` from `RLinf/franka_ros@0.10.0`.
- EN and ZH Sphinx builds exit 0 with the existing 28 autodoc warnings, none on changed pages; `check_rst_markup.py` and `pre-commit` pass.

Not yet validated: Docker builds, a full `--env franka` install with the Franka mock SAC e2e (local installs hit PyPI timeouts), and physical robot runs.

Migration:

- ROS deployments: add `backend: franka_ros` to every Franka hardware config and install with `--env franka-ros` natively on Ubuntu 20.04; the Docker image no longer includes ROS venvs.
- Hosts without a real-time kernel keep working with Franky; set `realtime_config: enforce` to require one.
- Replace `--env franka-dexhand` with `--env franka`, which now includes the dexterous-hand dependencies.
- `pyproject.toml` no longer defines the `franka`, `xsquare_turtle2`, `gim_arm`, `so101`, and `piper` extras; install with `bash requirements/install.sh embodied --env <robot>`.
- Use `python -m toolkits.realworld_check.test_franka_controller --backend franky` in place of `test_franky_controller.py`.
- Code that subclassed `RobotTask` or wrapped a robot in `RobotTaskEnv` becomes an env class registered in the robot's package.
- The published `rlinf/rlinf:agentic-rlinf0.4-franka` image predates this layout and needs a rebuild.

### Types of changes

### Checklist:


andylin-hao:
I don't think we can simply archive the multi-node setup. Users with old firmware will have trouble locating the docs since its "archived". A better organization is to modify the original franka doc:

1. In the real-time kernel section, let users (and also me) know which versions of franka firmware and libfranka no longer requires rt kernel, and how to upgrade. 
2. The setup section should be organized into two parts: the first part is for the new firmware and the second part is for the old firmware.

Also, the current franka docker image is CUDA-less, do we need a new image? Could you let me know which libfranka version is required for building the image?

lizuojun04:
> I don't think we can simply archive the multi-node setup. Users with old firmware will have trouble locating the docs since its "archived". A better organization is to modify the original franka doc:
> 
>     1. In the real-time kernel section, let users (and also me) know which versions of franka firmware and libfranka no longer requires rt kernel, and how to upgrade.
> 
>     2. The setup section should be organized into two parts: the first part is for the new firmware and the second part is for the old firmware.
> 
> 
> Also, the current franka docker image is CUDA-less, do we need a new image? Could you let me know which libfranka version is required for building the image?

I think we need a new docker image. The libfranka version is `0.19.0` which is not supported by the current docker image.
image

andylin-hao:
> > I don't think we can simply archive the multi-node setup. Users with old firmware will have trouble locating the docs since its "archived". A better organization is to modify the original franka doc:
> > ```
> > 1. In the real-time kernel section, let users (and also me) know which versions of franka firmware and libfranka no longer requires rt kernel, and how to upgrade.
> > 
> > 2. The setup section should be organized into two parts: the first part is for the new firmware and the second part is for the old firmware.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Also, the current franka docker image is CUDA-less, do we need a new image? Could you let me know which libfranka version is required for building the image?
> 
> I think we need a new docker image. The libfranka version is `0.19.0` which is not supported by the current docker image. image

OK, on it. The current install script does support 0.19.0, which is undocumented in the docs. I'll also add it. Do you know which franka firmware version is required for non-RT kernel?

lizuojun04:
> 1. In the real-time kernel section, let users (and also me) know which versions of franka firmware and libfranka no longer requires rt kernel, and how to upgrade.

I will update the document after you get the answer.

andylin-hao (dismissed):

docs/source-zh/rst_source/examples/embodied/franka_single_machine.rst
```diff
@@ -0,0 +1,624 @@
...
+
+.. figure:: https://raw.githubusercontent.com/RLinf/misc/main/pic/franka_arm_small.jpg
+   :align: center
+   :width: 80%
+
+   在一台带有 GPU 的主机上运行 Franka 真机强化学习流程。
+
+较新的 Franka 固件与 libfranka 版本不再依赖实时内核（Real-time Kernel），
+因此你可以将计算节点（训练 / rollout）与控制节点（Franka 控制） **安装在同一台有显卡的主机上**。
+本页指导你在单台主机上通过两个独立的 RLinf 环境完成真机 SAC / RLPD / PPO 训练。
```
Lin-xs: Could we note the minimum firmware and `libfranka` versions here?

Also: can users update the firmware themselves? If not, they may still need the legacy path. We should spell out when to use this approach vs. the legacy method.

docs/source-zh/rst_source/examples/embodied/franka_single_machine.rst
```diff
@@ -0,0 +1,624 @@
...
+RLinf 支持对多台 Franka 机器人进行统一管理，实现并行数据采集与训练。
+要启用多机器人设置，需要在配置文件的 `node_groups` 部分为每个机器人添加独立的配置。
+
+一个包含两台 Franka 机器人的配置示例位于
+``examples/embodiment/config/realworld_peginsertion_rlpd_cnn_async_2arms.yaml``，如下所示：
+
+.. code-block:: yaml
+
+  cluster:
+  num_nodes: 3 # 1 个训练 / rollout 节点 + 2 个机器人控制节点
```
Lin-xs: `num_nodes` / `component_placement` / `node_groups` are siblings of `cluster`, not children.

docs/source-zh/rst_source/examples/embodied/franka_single_machine.rst
```diff
@@ -0,0 +1,624 @@
...
+检查 Franka 固件版本
+----------------------------------------
+
+在机器人管理网页（一般为 ``http://<robot_ip>/desk``）中，点击 ``SETTINGS`` 选项卡，在 ``DashBoard`` 中查看 ``Control`` 后面的版本号，如下所示。
+请记录该固件版本号，后续设置 ``LIBFRANKA_VERSION`` 时会用到。
+
+.. raw:: html
+
+  <div style="flex: 1; text-align: center;">
+      <img src="https://raw.githubusercontent.com/RLinf/misc/main/pic/franka_firmware.png" style="width: 60%;"/>
```
Lin-xs: The version of firmware in this image is 5.7.2, which is misleading. Could you replace with a correct one?

ray_utils/realworld/cleanup.sh
```diff
@@ -0,0 +1,53 @@
...
+# cleanup.sh - Clean up leftover processes from RLinf real-world eval/training.
+# This includes: Ray clusters, Franka ROS controllers, roslaunch/roscore/rosmaster, FrankaController.
+set -u
+
+echo "==== 1. Stop the Ray cluster (head + control, single-machine) ===="
+# Prefer the ray in the current environment; fall back to a user-specified ray binary.
+if command -v ray >/dev/null 2>&1; then
+    ray stop 2>/dev/null || true
+else
+    (/path/to/your/venv/bin/ray stop 2>/dev/null || true)
```
Lin-xs: This shell script doesn't require users to set ray path at first, so you can throw an exception and exit. User will get an error like "/path/to/your/venv/bin/ray is not found", which is weird.

ray_utils/realworld/setup_compute_node.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+# setup_compute_node.sh - Startup script for the compute/rollout node (Ray head).
+# Usage:
+#   source setup_compute_node.sh          # Activate the compute environment
+#   source setup_compute_node.sh start    # Activate env and start the Ray head (rank 0)
+#   source setup_compute_node.sh stop     # Stop Ray on this node
+#
+# This node only serves the compute/rollout roles (actor / rollout / reward).
+# The control node (Franka) is started via setup_franka_node.sh and joins this
+# head as rank 1.
```
Lin-xs, line 10: Users need to set some environment variables before running this, but that isn’t documented anywhere, for example `RLINF_NODE_IP` and `FRANKA_ROBOT_IP`.

Please:
1. Document the required env vars at the top of the script/docs.
2. Explicitly list which ones must be set.
3. Raise an exception if they are unset.

The same issue applies to the other three scripts.

docs/source-zh/rst_source/examples/embodied/franka_single_machine.rst
lizuojun04: 1. The official FR3 compatibility matrix lists System Image >= `5.9.0` with
   `libfranka` >= 0.18.0. Therefore, the 5.9.x / 0.18+ pair is the minimum
   version combination I can substantiate for compatibility; libfranka `0.19.0`
   is also compatible with System Image `5.9.x` and works better than `0.18.0`.
2. I could not find an official firmware/libfranka version boundary after which
   a real-time kernel is no longer required. 

I am unable to test the version problem recently, so I didn't update this part of the documentation.

Users can update the firmware after they get a update file from the after-sales service manager of the franka. 
And I think that you can contact with the after-sales service manager as long as you are the owner of the franka.

docs/source-zh/rst_source/examples/embodied/franka_single_machine.rst
lizuojun04: I have updated it.

docs/source-zh/rst_source/examples/embodied/franka_single_machine.rst
lizuojun04: I have updated it.

ray_utils/realworld/cleanup.sh
lizuojun04: I have updated it.

ray_utils/realworld/setup_compute_node.sh
lizuojun04: I have updated it.

docs/source-en/rst_source/examples/embodied/franka_single_machine.rst
```diff
@@ -135,7 +135,7 @@ Please take a note of the firmware version for later use when setting ``LIBFRANK
 .. raw:: html
 
   <div style="flex: 1; text-align: center;">
-      <img src="https://raw.githubusercontent.com/RLinf/misc/main/pic/franka_firmware.png" style="width: 60%;"/>
+      <img src="https://raw.githubusercontent.com/RLinf/misc/main/pic/franka_firmware_single_machine.png" style="width: 60%;"/>
```
Lin-xs, line 139: This link is broken. Could you please double check it?

docs/source-zh/rst_source/examples/embodied/franka_single_machine.rst
Lin-xs: No need to pin an exact version cutoff. A note like this should be enough:

> Firmware 5.9.0 + libfranka 0.18.0 has been tested. If your firmware is older, or you run into other issues, please fall back to the legacy method (link to the RT kernel doc).

Lin-xs:
One more question: can users skip the RT-kernel requirement just by installing libfranka 0.18.0+ and firmware 5.9.0+ on a standard LTS Ubuntu (e.g. 20.04) without any modification to kernel / libfranka / others ? This matters a lot for reproducing this doc.

lizuojun04:
> One more question: can users skip the RT-kernel requirement just by installing libfranka 0.18.0+ and firmware 5.9.0+ on a standard LTS Ubuntu (e.g. 20.04) without any modification to kernel / libfranka / others ? This matters a lot for reproducing this doc.

I think users can skip the RT-kernel requirement. I have tested it on a standart LTS Ubuntu 20.04 with libfranka `0.19.0` and firmware `5.9.2` without the RT-kernel requirement. I have successfully run the evaluation and data collection pipline.

docs/source-en/rst_source/examples/embodied/franka_single_machine.rst
lizuojun04: I have updated it.

andylin-hao:
> > One more question: can users skip the RT-kernel requirement just by installing libfranka 0.18.0+ and firmware 5.9.0+ on a standard LTS Ubuntu (e.g. 20.04) without any modification to kernel / libfranka / others ? This matters a lot for reproducing this doc.
> 
> I think users can skip the RT-kernel requirement. I have tested it on a standart LTS Ubuntu 20.04 with libfranka `0.19.0` and firmware `5.9.2` without the RT-kernel requirement. I have successfully run the evaluation and data collection pipline.

libfranka cannot directly run on non-RT kernel. A config needs to be changed to ignore the RT kernel check in libfranka. For franka-ros this is:
```yaml
# franka_control/config/franka_control_node.yaml
realtime_config: ignore   # default: enforce
```
For franky this is:
```python
import franky

robot = franky.Robot(
    "172.16.0.2",
    relative_dynamics_factor=0.2,
    realtime_config=franky.RealtimeConfig.Ignore,   # default is .Enforce
)
```

This information is missing in the doc so we cannot reproduce the doc. The franky rt ignore also requires a code change in RLinf, not just config change. Also, during training the machine can still be heavily loaded and the OS scheduling could be untimely. We cannot simply remove the RT kernel note from the doc. At least a warning needs to be kept there recommending  rt kernel installation. The doc title "Single-Machine Franka" is also weird. Why isn't it the "Real-World RL" as before or simply "RL with Franka" since now this is the main franka doc entrypoint?

lizuojun04:
I add RT kernel installation guide in documentation. 
The process of installing the RT kernel is complicated and varies significantly depending on the operating system version. Therefore, I have not included the complete installation procedure in this document. 
Instead, I provide links to the official installation guides along with relevant installation notes.
[franka RT kernel installation guide](https://frankarobotics.github.io/docs/doc/libfranka/docs/real_time_kernel.html)
[ubuntu pro RT kernel installation guide](https://ubuntu.com/pro-client/docs/en/latest/howtoguides/enable_realtime_kernel)

lizuojun04:
Would you be able to test whether the training results on a single machine with the RT kernel installed are consistent with those from the NUC + 4090 machine? If the results are consistent, I’ll mark the original document as archived.

andylin-hao (approved):

guozhen1997 (approved):

Lin-xs (approved):

PR #1461, opened 2026-08-09
feat(env): add optional obs compression

### LINKED ISSUE #1392: [Feature]: add compression for env observation transfer
### Feature description

Introduce an optional lossless image compression mechanism for communication between Env and Rollout workers. The proposed pipeline applies XOR delta encoding between consecutive image observations, followed by LZ4 or Zstd compression. This can substantially reduce image payload sizes when adjacent frames share significant visual similarity. The feature should be configurable and disabled by default, with options to select the codec and compression level. It is intended to improve rollout throughput and communication latency in bandwidth-constrained or multi-node deployments while preserving the original image data exactly.

In a preliminary test using consecutive LIBERO observations generated by actual PI0 inference, each EnvWorker transferred a 3 MiB image batch:
XOR + LZ4 reduced the payload to 40.51% of its original size,a 59.49% size reduction and a 2.47× compression ratio. With two-way parallel processing and preallocated buffers, compression took approximately 2.65 ms, while decompression into pinned memory took 0.66 ms.
XOR + Zstd level 1 reduced the payload to 35.97%a 64.03% size reduction and a 2.78× compression ratio. Parallel compression took approximately 4.70 ms, while decompression into pinned memory took 1.93 ms.

### Additional context

_No response_

kakiuwang-ui:
Closes #1392

## Summary

Image observations are the dominant payload on the Env → Rollout channel in embodied tasks. In bandwidth-constrained or multi-node setups this transfer can throttle rollout throughput. This PR adds an **optional, disabled-by-default, lossless** codec that compresses the `uint8` image tensors before transfer and reconstructs them on the rollout side.

The codec is self-contained per message (no cross-message or cross-worker state), so it is robust to reordering and environment resets, and the reconstructed observations are bit-for-bit identical to the originals.

## What it does

- **XOR-delta (optional):** XORs consecutive frames along the batch axis. Parallel embodied envs (LIBERO, ManiSkill) render largely static backgrounds, so the delta is mostly zeros and compresses much better.
- **Entropy coding:** `lz4` (fastest) or `zstd` (better ratio), selectable via config.
- Only `uint8` image tensors are compressed; states, flags, and text fields pass through untouched.

## Changes

- `rlinf/utils/obs_compression.py`: the codec (`compress_obs` / `decompress_obs`), backends imported lazily.
- `rlinf/workers/env/env_worker.py`: compress in `_build_rollout_input_data` when enabled (covers sync/async/rtc env workers via inheritance).
- `rlinf/workers/rollout/hf/huggingface_worker.py`: decompress at the top of `_merge_obs_batches`, the single funnel for all recv paths. No-op when compression is off.
- `pyproject.toml`: optional `[compression]` extra (`lz4`, `zstandard`).
- Config example under `env.obs_compression` in `libero_10_grpo_openpi.yaml`.
- Docs: new performance guide (English + 中文) with usage and expected gains.
- Tests: `tests/unit_tests/test_obs_compression.py` — round-trip losslessness for both codecs (with/without XOR), single-frame batches, disabled passthrough, decompress-is-noop on uncompressed payloads, and non-image fields left untouched.

## Configuration

```yaml
env:
  obs_compression:
    enable: true      # off by default
    codec: lz4        # lz4 (faster) or zstd (better ratio)
    level: 1          # zstd compression level; ignored by lz4
    xor_delta: true   # XOR consecutive frames before compression
```

## Measured impact

On consecutive LIBERO observations (~3 MiB of images per Env Worker step):

| Configuration | Size vs. original | Compress | Decompress |
| --- | --- | --- | --- |
| XOR + LZ4 | ~40.5% | ~2.65 ms | ~0.66 ms |
| XOR + Zstd (level 1) | ~36.0% | ~4.70 ms | ~1.93 ms |

## Testing

```
pytest tests/unit_tests/test_obs_compression.py
```

9 passed; `ruff check --preview` and `ruff format` clean.


kakiuwang-ui:
Hi @andylin-hao @guozhen1997 @Lin-xs — this is my first PR here, so the CI workflows are waiting on a maintainer to approve them. Could one of you kick them off when you get a chance? The change is opt-in and off by default, and I'm happy to adjust anything after review. Thanks!

guozhen1997:
Hi @kakiuwang-ui, thanks for the PR — the motivation is clear, the codec is thoughtfully designed as opt-in / lossless / per-message, and the EN/ZH docs + unit tests are a good start. The default-off path looks safe.

The main concern is that enabling compression appears incompatible with the existing Env → Rollout channel routing.
Compressed markers break infer_batch_size / split_batch on send. Compression is applied in `EnvWorker._build_rollout_input_data` before `send_to`, which then calls `infer_batch_size` and `split_batch` (rlinf/scheduler/worker/routing.py).

kakiuwang-ui:
Thanks @guozhen1997, you're exactly right — that was a real bug. Compressing in `_build_rollout_input_data` (before `send_to`) meant `send_to` then ran `infer_batch_size` and `split_batch` on marker dicts, which crashes as soon as compression is enabled.

Fixed in the latest commit by moving compression into a `split_fn` (`_split_and_compress_obs`):

- `send_to` first splits the **uncompressed** payload with the normal `split_batch`, so `infer_batch_size` / `split_batch` only ever see plain tensors and routing is unchanged.
- Each per-rank shard is compressed *after* the split; the rollout side decompresses in `_merge_obs_batches` before use. This holds in both normal and decoupled mode, since both go through the same split path.

I also added an integration test (`test_routing_split_then_compress_roundtrip`) that runs the real `split_batch` / `merge_batches` end-to-end to guard against exactly this regression, and while tracing consumers I noticed the SGLang embodied rollout worker is a second receiver of the obs channel, so I wired decompression into its `_merge_obs_batches` too.

One note: `lz4`/`zstandard` are an opt-in `[compression]` extra, so the codec tests skip unless it's installed. Happy to add `uv pip install .[compression]` to the unit-test job if you'd like them exercised in CI — let me know your preference.

rlinf/workers/rollout/hf/huggingface_worker.py
thereAreDemonsNearby, line 901: If a batch consists entirely of images (i.e., there are no non-image features like states), the recv_from call at the Hugging Face worker may fail when inferring the batch size (_infer_env_batch_size in huggingface_worker.py), because compressed images are neither tensors nor lists. Please check and add unit tests for this.

tests/unit_tests/test_obs_compression.py
```diff
@@ -0,0 +1,177 @@
+# Copyright 2025 The RLinf Authors.
```
thereAreDemonsNearby: please change to 2026

kakiuwang-ui:
Thanks @thereAreDemonsNearby, both good catches — fixed in the latest commit.

1. **Images-only batch → batch-size inference.** You're right: `_infer_env_batch_size` runs on the receive path *before* decompression, so a batch whose only batched field is a compressed image (no `states` / `task_descriptions`) would fall through and raise. I added `infer_obs_batch_size` in `obs_compression.py` that reads the batch size straight from the marker's stored `shape`, and routed both the HF and SGLang embodied workers through it (this also removes the duplicated `_infer_env_batch_size`). New unit tests cover the images-only case (`test_infer_obs_batch_size_images_only`) plus the normal and compressed cases.

2. **Copyright year → 2026.** Done, for both new files.

The full split → compress → decompress → merge path is also now exercised end-to-end against the real `split_batch` / `merge_batches` in `test_routing_split_then_compress_roundtrip`.

thereAreDemonsNearby (approved):

guozhen1997 (approved):

kakiuwang-ui:
Thanks @thereAreDemonsNearby and @guozhen1997 for the reviews and approvals — much appreciated.

Quick status: I rebased onto the latest `main` (resolving the `rlt_td3` / smooth-intervene overlaps), so the branch is conflict-free again, and `commit-check`, `lint`, `pr-title-check`, and both docs builds are green. The heavier test jobs are still gated on the `run-ci` label.

Could a maintainer add the `run-ci` label so the e2e/unit suites run, and @andylin-hao / @Lin-xs take a look when you have a moment? Happy to address anything else.

guozhen1997:
Hi @kakiuwang-ui, could you fix the ci test errors?

kakiuwang-ui:
Thanks @guozhen1997 — fixed the real one.

The `unit-tests` failures were a genuine break I introduced: `test_overlap_env_bootstrap` constructs `EnvWorker` via `object.__new__` (bypassing `__init__`), so once the obs send path referenced `self._obs_split_fn`, it raised `AttributeError`. Fixed by giving `EnvWorker` a class-level `_obs_split_fn = None` (meaning "use the scheduler's default split"), so partially-constructed instances and the compression-disabled path are both safe; `__init__` still installs the compressing `split_fn` when `env.obs_compression` is enabled. All `test_obs_compression` and `test_overlap_env_bootstrap` tests pass locally.

The remaining red checks look like infra, not this PR:
- `embodied-gr00t-libero-ascend`: `SSLError: UNEXPECTED_EOF` while downloading `RLinf/LIBERO-assets` from huggingface.co (network).
- `unit-tests-rocm`: 24h timeout (runner hung).
- `reason-qwen-grpo-test-rollout-logprobs`: reasoning path, unrelated to the embodied obs change.

Could you re-run CI once the fix is picked up? Happy to look further if anything else turns red.

kakiuwang-ui:
The only red check, `embodied-openpi-calvin-test`, fails before it reaches any code touched by this PR. The traceback is a dependency mismatch in that runner's venv:

```
train_embodied_agent.py -> rlinf.workers.reward -> peft -> transformers -> huggingface_hub
ImportError: cannot import name 'is_offline_mode' from 'huggingface_hub'
```

`is_offline_mode` was removed from newer `huggingface_hub`, so the installed `transformers` can't import it. This is an import-time env issue unrelated to observation compression — none of the changed files (obs_compression, env_worker, rollout workers) appear in the trace, and every other openpi e2e job in the same run (maniskill, behavior, robocasa, metaworld, ...) passed. Looks like a flaky venv on that one runner.

Could you re-run `embodied-openpi-calvin-test`? All unit-tests (cpu/cuda/ascend/musa/rocm) and the rest of the e2e suite are green after the `_obs_split_fn` fix.

pyproject.toml
```diff
@@ -95,6 +95,11 @@ embodied = [
     "gymnasium",
     "gym",
 ]
+# Optional codecs for `env.obs_compression` (Env -> Rollout observation transfer).
+compression = [
```
andylin-hao: Could you move these two up to the main dependency block rather than an optional dependency block.

andylin-hao (approved):

pyproject.toml
kakiuwang-ui: Done in 93d0cc5c — moved `lz4` and `zstandard` into the core `dependencies` block and dropped the `compression` extra. Also removed the now-redundant explicit codec installs from the unit-test CI jobs and the `pip install rlinf[compression]` step from the docs, since they ship with the package now.

PR #1467, opened 2026-08-10
feat(robocasa): add RPent commands to subprocess worker

qurakchin:
## Summary
- The robocasa subprocess `_worker` loop only handled the standard venv commands, so every agentic facade command from RPent's `RoboCasaEnvFacade` hit `NotImplementedError` and killed the worker.
- Port the agent-side helpers to module scope and wire the missing worker commands (camera calibration, raw render, grasp contact, action reassembly, success-criteria introspection, task progress, seeding).
- `set_seed` reseeds `random`/`np`/`env.seed`/`env.rng` inside the subprocess so the `RLDX_RESET_SEED` paired-comparison flow reproduces the exact scene the fullshot eval scored.

## Test plan
- [ ] `Robocasa365Env` facade methods round-trip through the subprocess pipe without `NotImplementedError`.
- [ ] `set_seed` reproduces a known scene when given the same `RLDX_RESET_SEED`.
- [ ] Run a full RoboCasa365 eval episode end-to-end via RPent's env_client.


andylin-hao (commented):
_Superseded — replaced by the inline comments on the relevant lines._

rlinf/envs/robocasa365/robocasa365_env.py
```diff
@@ -749,3 +749,124 @@ def close(self) -> None:
...
+        keys (``robot0_eef_pos``, ``language``, …) directly.
+        """
+        raw_obs, info_list = self.env.reset(id=[env_idx])
+        if self.current_raw_obs is None:
+            self.current_raw_obs = [None] * self.num_envs
+        if self.current_info_list is None:
+            self.current_info_list = [None] * self.num_envs
+        self.current_raw_obs[env_idx] = raw_obs[env_idx]
+        self.current_info_list[env_idx] = info_list[env_idx]
+        return raw_obs[env_idx], info_list[env_idx]
```
andylin-hao, line 783: `env_idx` is used as an index into the *returned* arrays, but `BaseVectorEnv.reset(id=[k])` returns arrays of length `len(id)` indexed from zero, not by env id:

```python
ret_list = [self.workers[i].recv() for i in id]   # venv/venv.py:815
...
obs = np.stack(obs_list)
```

So `raw_reset(env_idx=1)` resets env 1 and then does `raw_obs[1]` on a length-1 array → `IndexError`. Should be `raw_obs[0]` / `info_list[0]` (the position within the requested subset), while `self.current_raw_obs[env_idx]` on line 781 correctly stays keyed by env id.

Latent today since RPent builds this with `num_envs=1` and only ever passes `env_idx=0` — but the parameter currently promises something the code doesn't deliver. Either fix the indexing or drop the parameter and assert `num_envs == 1` for this whole facade block.

rlinf/envs/robocasa365/robocasa365_env.py
```diff
@@ -749,3 +749,124 @@ def close(self) -> None:
...
+
+    def raw_step(self, action, env_idx: int = 0):
+        """Step a single env with a flat ``[action_dim]`` action.
+
+        Returns the robosuite 4-tuple ``(obs, reward, done, info)`` (single
+        env, env dim stripped).  ``info`` already carries ``success`` and
+        ``ep_meta`` (injected by the robocasa worker loop).
+        """
+        actions = np.asarray(action, dtype=np.float64)[None, :]
+        raw_obs, rewards, dones, info_list = self.env.step(actions)
```
andylin-hao, line 793: Mirror of the `raw_reset` issue, in the other direction: no `id=` is passed, so `BaseVectorEnv.step` falls back to all envs and asserts `len(action) == len(id) == num_envs`. With `num_envs > 1` a single-env action fails that assert before `env_idx` is ever consulted.

Compare `raw_reset` above, which does pass `id=[env_idx]` — the two are inconsistent. Suggest `self.env.step(actions, id=[env_idx])` and indexing the result at `[0]`.

rlinf/envs/venv/venv.py
```diff
@@ -432,6 +432,63 @@ def render_camera(
...
+        Handled by the robocasa worker loop."""
+        self.parent_remote.send(["get_task_progress", {}])
+        return self.parent_remote.recv()
+
+    def set_seed(self, seed) -> Any:
+        """Full re-seed (random + np + env.seed + env.rng) for paired
+        comparisons. Handled by the robocasa worker loop."""
+        self.parent_remote.send(["set_seed", seed])
+        return self.parent_remote.recv()
+
```
andylin-hao, line 491: These eight methods are RoboCasa-specific but land on the generic `SubprocEnvWorker`, while only the RoboCasa `_worker` loop in `rlinf/envs/robocasa/venv.py` handles the corresponding commands. Calling any of them against a maniskill or metaworld worker sends an unknown cmd, and that loop's fallback is:

```python
else:
    p.close()
    raise NotImplementedError(f"Unknown command: {cmd}")
```

The worker closes its end of the pipe and dies; the parent is left blocking in `parent_remote.recv()`. Nothing calls these for other envs today, so it's latent — but every env's worker API now advertises RoboCasa semantics it can't honour.

A `RobocasaSubprocEnvWorker` subclass would keep the surface where the handlers live. Failing that, at minimum a "RoboCasa worker only" line in each docstring.

rlinf/envs/robocasa/venv.py
```diff
@@ -35,6 +37,213 @@
...
+                return _local
+
+            return _local
+        return None
+
+    import sys as _sys
+
+    old = _sys.gettrace()
+    try:
+        _sys.settrace(_tracer)
```
andylin-hao, line 214: Global `sys.settrace` — the previous tracer is restored in the `finally`, which is the important part, but two things worth noting:

1. It will fight any debugger or `coverage` run attached to the worker for the duration of the traced call.
2. It runs `_check_success()` an *extra* time per invocation, and RPent calls `get_task_progress` once per recorded step (via `current_state_dict`). Worth confirming `_check_success` is genuinely side-effect-free across all 50 RoboCasa365 tasks — several of them mutate counters (`washed_time`, `success_time`) in nearby code paths.

A comment naming the trade-off would help the next reader.

rlinf/envs/robocasa/venv.py
```diff
@@ -35,6 +37,213 @@
...
+
+
+def _set_seed(env, seed):
+    """Re-seed the subprocess env (random/np/env.seed/env.rng) — used by the
+    RLDX_RESET_SEED paired-comparison flow."""
+    import random
+
+    sd = int(seed)
+    random.seed(sd)
+    np.random.seed(sd)
```
andylin-hao, line 239: This reseeds the process-global `random` and `numpy` RNGs, not just the env. Correct for the single-env agentic server this targets, but it makes the command unsafe the moment a worker hosts more than one env — worth stating that constraint in the docstring, since the command is reachable from the generic `SubprocEnvWorker.set_seed`.

rlinf/envs/robocasa/venv.py
```diff
@@ -140,6 +349,41 @@ def get_ep_meta(env, env_return):
...
+                        _render_raw(
+                            env,
+                            cam=data["camera_name"],
+                            h=data["height"],
+                            w=data["width"],
+                            depth=data["depth"],
+                        )
+                    )
+                else:
+                    p.send(env.render(**data) if hasattr(env, "render") else None)
```
andylin-hao, line 370: This fallback can't work: `data` here is `{"camera_name", "height", "width", "depth"}`, and `gym.Env.render()` accepts none of those — it would raise `TypeError` rather than degrade gracefully. Either drop the branch (every RoboCasa env has `.sim`) or map the kwargs to something `render()` actually takes.

andylin-hao (commented):
A few line-level notes below.

Two things that look like bugs but aren't, so nobody re-checks them: `check_success` and `get_camera_meta` get no new parent-side wrapper here because they already exist on `SubprocEnvWorker` (`venv/venv.py:399,598`); and `raw_step`'s 4-tuple unpack is correct — it matches how the existing `Robocasa365Env.step` already unpacks `self.env.step()` at line 620.

Separately: nine new commands cross a process boundary with no tests, and every box in the test plan is unchecked. One test round-tripping each command against a stub worker would have caught the `env_idx` indexing issue below without needing MuJoCo.

PR #1469, opened 2026-08-11
fix(runner): skip incomplete checkpoints during auto resume

yicwang:
### Description

With `runner.resume_dir: auto`, `ReasoningRunner.init_workers` picked the highest
`global_step_<N>` directory under `checkpoints/`, without checking whether that
checkpoint had finished writing. A run killed midway through `_save_checkpoint`
leaves a directory that looks newest but is incomplete, and the next launch resumes
from it.

`_save_checkpoint` writes the actor first, then the optional critic, then the
dataloader state. This PR publishes the dataloader state through a temporary file
and an atomic rename, so `data/data.pt` exists only once the whole checkpoint has
been written, and auto resume walks the checkpoints newest-first and takes the first
one whose actor, optional critic, and dataloader state are all present. Skipped
candidates are logged at warning level.

### Motivation and Context

Resuming from a half-written checkpoint fails in ways that are easy to misread: a
missing `critic/` directory surfaces as a load error, a truncated `data/data.pt`
fails to deserialize, and a `data/data.pt` that was never written at all only warns
before restarting the data stream from the beginning — so the resumed run quietly
retrains on samples it had already consumed.

### How has this been tested?

New unit tests, run on CPU with Python 3.12:

```bash
pytest tests/unit_tests/test_auto_resume_checkpoint.py
# 6 passed
```

They cover: auto resume skipping an incomplete newest checkpoint, taking the newest
complete one, and starting from scratch when none is complete; the critic being
required only when a critic is configured; the dataloader state being written to a
temporary path and renamed into place; and an interrupted save publishing neither
`data.pt` nor a leftover `data.pt.tmp`.

`ruff check` and `ruff format --check` are clean.

Worth flagging for review: `data/data.pt` is used as the completion marker, so a
checkpoint directory that has an actor but no dataloader state is now skipped by
auto resume, whereas before it would have been selected and resumed with the data
stream reset. Explicit `runner.resume_dir: <path>` is unaffected — the check applies
only to `auto`.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/runners/reasoning_runner.py
+++ b/rlinf/runners/reasoning_runner.py
@@ -307,17 +307,21 @@ class ReasoningRunner:
                     and os.path.isdir(os.path.join(checkpoints_dir, d))
                 ]
 
-            if checkpoint_steps:
-                max_step = max(checkpoint_steps)
-                self.cfg.runner.resume_dir = os.path.join(
-                    checkpoints_dir, f"global_step_{max_step}"
+            resume_dir = None
+            for step in sorted(checkpoint_steps, reverse=True):
+                candidate = os.path.join(checkpoints_dir, f"global_step_{step}")
+                if self._is_complete_checkpoint(candidate):
+                    resume_dir = candidate
+                    break
+                logging.warning(
+                    f"Skipping incomplete checkpoint {candidate} during auto resume."
                 )
-                logging.info(
-                    f"Auto resume from checkpoint: {self.cfg.runner.resume_dir}"
-                )
+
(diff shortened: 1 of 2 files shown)


guozhen1997 (approved):

andylin-hao (approved):

guozhen1997:
Hi @yicwang, could you please rebase this PR onto the latest main branch?

yicwang:
> Hi @yicwang, could you please rebase this PR onto the latest main branch?

Done

PR #1475, opened 2026-08-11
fix(algorithms): normalize PPO actor metrics by mask width

### LINKED ISSUE #1471: [Bug]: actor/approx_kl and actor/clip_fraction are inflated by action_dim x num_action_chunks (metrics only)
### Affected component

`rlinf/algorithms/losses.py` — the `actor/approx_kl` and `actor/clip_fraction`
metrics reported by `compute_ppo_actor_loss` (and the same construction in
`compute_decoupled_ppo_actor_loss`).

Metrics only: the loss itself is unaffected.

### Bug description

In `compute_ppo_actor_loss`, `actor/approx_kl` and `actor/clip_fraction` divide a sum taken over a **broadcast** tensor by `loss_mask.count_nonzero()`, which counts the **un-broadcast** mask. When `loss_mask` and `ratio` have different trailing shapes, both metrics are inflated by `numel(broadcast) / count_nonzero(loss_mask)`.

I measured **84x** inflation on an embodied PPO/GRPO run (`reward_type: chunk_level` + `logprob_type: token_level`).

The same function already handles this shape mismatch for its sibling metrics 20 lines below, which is what makes this look unintended:

https://github.com/RLinf/RLinf/blob/main/rlinf/algorithms/losses.py#L294-L296

```python
# Only broadcast when ratio has action_dim dimension and loss_mask's last dim is 1
# This handles token_level mode: ratio [bsz, num_chunks, action_dim], loss_mask [bsz, num_chunks, 1]
if len(ratio.shape) > 2 and loss_mask.shape[-1] == 1 and ratio.shape[-1] > 1:
    loss_mask_for_metrics = loss_mask.expand_as(ratio)
```

`ratio`, `ratio_abs`, `clipped_ratio` and `dual_cliped_ratio` all use the expanded mask. `approx_kl` and `clip_fraction` are computed **before** it, on lines 277-278, and use the un-expanded `loss_mask_count` from line 242.

`loss_mask` is built in `compute_loss_mask` as `[n_chunk_step, bsz, num_action_chunks]`, then collapsed on the last dim when `reward_type == "chunk_level"`:

```python
if reward_type == "chunk_level":
    loss_mask = loss_mask.any(dim=-1, keepdim=True)
```

while `logprob_type: token_level` keeps `logprobs` / `ratio` at full `action_dim` width. So the two shapes disagree by construction for that config combination — which is exactly the one named in the comment on line 295.

This is not an exotic combination. Three shipped example configs use it:

| config | `reward_type` | `logprob_type` |
|---|---|---|
| `examples/embodiment/config/behavior_ppo_openpi.yaml` | `chunk_level` | `token_level` |
| `examples/embodiment/config/behavior_ppo_openpi_pi05.yaml` | `chunk_level` | `token_level` |
| `examples/embodiment/config/behavior_ppo_openvlaoft.yaml` | `chunk_level` | `token_level` |

(The `behavior_ppo_*` ones go through `compute_decoupled_ppo_actor_loss`, which has the same construction — see below.)

### Configuration YAML file

The triggering combination is `reward_type: chunk_level` + `logprob_type: token_level`:

```yaml
algorithm:
  adv_type: grpo
  loss_type: actor
  logprob_type: token_level
  entropy_type: token_level
  reward_type: chunk_level
  group_size: 2
  normalize_advantages: True
```

Three shipped configs use the same combination — see the table above.

### Log file

Wrapping `compute_ppo_actor_loss` and comparing the reported metric against a properly masked per-element mean, on two independent ranks:

```
logprobs (1, 12, 7)   loss_mask (1, 1, 1)
numerator elements = 84   denominator = 1   -> 84.000x

rank 0:  true mean(-log_ratio) = 0.001637   reported actor/approx_kl = 0.137518
rank 1:  true mean(-log_ratio) = 0.000731   reported actor/approx_kl = 0.061426
```

`0.001637 x 84 = 0.137508` and `0.000731 x 84 = 0.061404` — the reported value is exactly `84 x` the per-element mean, to the precision printed.

Two visible symptoms in the same run:

1. **`actor/approx_kl` goes negative** (observed `-0.291`). It is the k1 estimator `-mean(log_ratio)`, which is signed, so a small true value amplified 84x scatters around zero with magnitude ~1. In this run `actor/grad_norm = 0.0000` (all advantages were zero), so the weights never moved and the true log-ratio was pure recompute noise: per-element quantiles `[min -0.235, p50 0.000, max 0.361]`, no tail (`0/84` elements below `-1`). The reported `approx_kl` still swung between `-0.291` and `+1.293` across steps.

2. **`clip_fraction` is no longer a fraction.** Line 277 is `(clip_mask * loss_mask).sum() / float(loss_mask_count)`, where `clip_mask` has `ratio`'s shape. With the same 84x factor its range becomes `[0, 84]`. This is the part that seems hard to read as intentional: whatever one thinks the right normalization for a KL estimate is, a quantity named `clip_fraction` should be bounded by 1.

I only measured `approx_kl` (advantages were zero in my run, so `clip_mask` was all-false and `clip_fraction` stayed at 0). The `clip_fraction` claim is from the code path, not from a measurement.

### Environment

- RLinf @ `9dc2e07` (2026-08-10)
- FSDP actor, `adv_type: grpo`, `loss_type: actor`, `logprob_type: token_level`, `entropy_type: token_level`, `reward_type: chunk_level`
- LIBERO-style embodied setup, action chunk 12, action dim 7

### Other reproduction info

`approx_kl` is the standard knob users watch to decide whether a policy is moving too fast, and the usual rules of thumb (~0.01-0.5) are per-element. Inflated by `action_dim x num_action_chunks`, a perfectly healthy run reads as a KL blow-up. I spent a while chasing a "25x too large KL" in my own policy before finding that the true per-element value was ~0.002.

`compute_decoupled_ppo_actor_loss` (lines 128-141) has the same construction for `clip_fraction`, `dual_clip_fraction`, `proximal_approx_kl` and `behav_approx_kl` — all dividing broadcast sums by `loss_mask_count`. Not measured, reported by inspection only.

Move the mask expansion above the metric computation and use it for all of them, e.g.

```python
loss_mask_for_metrics = loss_mask
if len(ratio.shape) > 2 and loss_mask.shape[-1] == 1 and ratio.shape[-1] > 1:
    loss_mask_for_metrics = loss_mask.expand_as(ratio)
metrics_mask_count = loss_mask_for_metrics.count_nonzero() or 1

clip_fraction = (clip_mask & loss_mask_for_metrics).sum() / float(metrics_mask_count)
approx_kl = -torch.where(loss_mask_for_metrics, log_ratio.detach(), 0.0).sum() / float(metrics_mask_count)
```

Note `loss_mask_count` is also used for the loss itself, so it should keep its current meaning there; only the metrics need the expanded count.

If the current behaviour for `approx_kl` is deliberate (a per-sequence joint KL summed over action dims is a defensible quantity), it would help a lot to say so in the metric name or the docs, since it sits in the same dict as `ratio_abs`, which is a per-element mean. `clip_fraction` would still need fixing either way.

kakiuwang-ui:
### Description

Normalize the PPO actor logging metrics by a mask expanded to the metric tensor's width in `rlinf/algorithms/losses.py`, so they are no longer inflated by `action_dim`.

- `compute_ppo_actor_loss`: `actor/approx_kl` and `actor/clip_fraction` are now reduced with `masked_mean` over `loss_mask` broadcast to the metric width, matching the sibling ratio metrics.
- `compute_decoupled_ppo_actor_loss`: `actor/clip_fraction`, `actor/dual_clip_fraction`, `actor/proximal_approx_kl`, `actor/behav_approx_kl`, `actor/behav_clip_fraction`, `actor/proximal_ratio`, and `actor/clipped_proximal_ratio` are all put on the same expanded-mask basis.
- No loss/gradient computation is changed — this is a metrics-only fix.

### Motivation and Context

These metrics summed over the broadcast, full-width tensor `[bsz, num_chunks, action_dim]` but divided by the count of the collapsed `loss_mask` `[bsz, num_chunks, 1]` (last dim collapsed when `reward_type: chunk_level`). When `logprob_type: token_level` keeps the metric tensors at full `action_dim` width, the denominator undercounts and the metrics are inflated by `action_dim` — measured at ~84x on an embodied PPO/GRPO run. The combination ships in three example configs (`behavior_ppo_openpi.yaml`, `behavior_ppo_openpi_pi05.yaml`, `behavior_ppo_openvlaoft.yaml`), which route through the decoupled loss. `compute_ppo_actor_loss` already applied this broadcast fix to its ratio metrics but computed `approx_kl`/`clip_fraction` before it; the decoupled variant had no such fix at all.

Fixes #1471

### How has this been tested?

Added `tests/unit_tests/test_ppo_actor_loss_metrics.py` (4 tests, all passing) covering both loss functions: it pins the exact per-element metric value, asserts the previous (buggy) normalization was exactly `action_dim`x larger, and asserts the metrics are invariant to whether the caller passes a collapsed or pre-expanded mask.

```
$ python -m pytest tests/unit_tests/test_ppo_actor_loss_metrics.py -q
4 passed in 1.32s
$ ruff check rlinf/algorithms/losses.py tests/unit_tests/test_ppo_actor_loss_metrics.py
All checks passed!
```

Reported and diagnosed in #1471 (with the line references to the existing partial fix in `compute_ppo_actor_loss`).

### Types of changes

### Checklist:


kakiuwang-ui:
@guozhen1997 @andylin-hao could you take a look when you have a moment? This is a metrics-only fix for #1471.

A couple of points worth your eyes:

- **Scope**: besides the two metrics named in the issue (`actor/approx_kl`, `actor/clip_fraction` in `compute_ppo_actor_loss`), I also put the decoupled-PPO siblings on the same expanded-mask basis (`proximal_approx_kl`, `dual_clip_fraction`, `proximal_ratio`, `clipped_proximal_ratio`, `behav_*`), since they share the identical broadcast/narrow-count mismatch. Happy to trim the scope back to just the two named metrics if you'd prefer a minimal diff.
- **Subtlety to double-check**: in `compute_decoupled_ppo_actor_loss`, `behav_clip_fraction` is a ratio of two counts. I expand `behav_mask` the same way as `loss_mask` so both numerator and denominator scale together — please confirm that matches the intended semantics.

The loss/gradient path is untouched. Added `tests/unit_tests/test_ppo_actor_loss_metrics.py` covering both functions.

Thanks also to the reporter for the precise diagnosis.

tests/unit_tests/test_ppo_actor_loss_metrics.py
guozhen1997, line 1: Please delete this file.

kakiuwang-ui: Done — removed the test file in a8cbf4a9. The PR is now just the metrics-only change in rlinf/algorithms/losses.py.

andylin-hao (approved):

guozhen1997 (approved):

PR #1476, opened 2026-08-12
feat: support tensor compression in collective group

anHappyDog:
### Description

Adds optional lossless compression for CPU tensors on the generic `Worker.send`/`Worker.recv` path, configured job-wide under a new `cluster.collective` block.

- New `cluster.collective.tensor_compression`: `codec` selects `lz4` or `zstd`, with `enabled`, `min_bytes` (16384), `excluded_dtypes` (`[float32]`) and the codec's own parameters — `acceleration`, or `level` and `max_inflight`.
- New `cluster.collective.tensor_buffer_pool.max_bytes` (default 2 GiB) bounding the reusable CPU byte buffers a codec borrows, per Worker.
- Both blocks are typed `ClusterConfig` fields validated on the driver, so an unknown or misspelled key fails at startup like any other cluster key, and the resolved configuration reaches every Worker with the rest of `ClusterConfig`.
- Compression applies only to CPU tensors carried by `send`/`recv` as a tensor, tensor list or tuple, tensor dict, or dataclass tensor fields. Accelerator tensors, arbitrary pickled objects, `broadcast`, and `send_tensor`/`recv_tensor` are deliberately unchanged.
- Every step may decline: no codec free, no buffer within budget, an unsupported size, or output no smaller than the input each leave that tensor raw, so a transfer is never worse than today.
- When `cluster.net_emulation` is enabled it charges the compressed CPU byte count, so emulated bandwidth reflects the saving.
- Adds `liblz4` and `libzstd` to `requirements/sys_deps.sh` (apt, dnf, yum, pacman) and to the Docker base stage; each Worker probes the configured library at startup so a missing dependency fails early.
- Documents scope, codec trade-offs, config keys and fallback rules in EN and ZH `concepts/collective.rst`.

Omitting `cluster.collective`, or setting `enabled: false`, keeps the existing wire path exactly.

### Motivation and Context

On links where transfer time dominates codec time, shrinking CPU tensor payloads reduces wall-clock transfer cost. This makes that trade available as an opt-in, job-wide setting rather than something each call site has to arrange, and keeps it off by default so no existing run changes behaviour. No linked issue.

### How has this been tested?

`pytest tests/unit_tests/test_tensor_compression.py test_comm.py test_channel.py test_broadcast_sync.py test_cluster_config.py test_net_emulation.py` — 369 passed. That covers 52 new unit tests for the codecs, buffer-pool eviction and budget exhaustion, codec/buffer release ordering, config validation and Worker resource setup, plus a sync and async end-to-end compressed tensor-list transfer in `test_comm.py`, whose module fixture now enables compression for all 123 cases in that file.

`pre-commit run --files <changed files>` — clean.

Round-trip smoke check through the real `libero_spatial_ppo_openpi` cluster config: it parses to `LZ4CompressionConfig`, and a 524288-byte int64 tensor compresses to 262292 bytes and restores byte-for-byte. Byte-exact round trips also verified across float32, float16, bfloat16, int64, int32, uint8, bool and complex64 for both codecs, and under eight concurrent threads for Zstd.

`tests/e2e_tests/embodied/libero_spatial_ppo_openpi.yaml` now enables compression, so the existing e2e job exercises it in CI.

Not measured: end-to-end training throughput on a bandwidth-constrained link. The numbers above are correctness and ratio checks on synthetic payloads, not a benchmark, so the defaults (`lz4`, `min_bytes: 16384`, `excluded_dtypes: [float32]`) are chosen on reasoning rather than measurement. Since the feature is off unless `cluster.collective` is present, this does not affect existing runs, but a reviewer should treat the defaults as a starting point.

### Types of changes

### Checklist:


rlinf/utils/tensor_codec.py
```diff
@@ -0,0 +1,264 @@
...
+    def _check_result(self, result: int, operation: str) -> None:
+        if self._library.ZSTD_isError(result):
+            error = self._library.ZSTD_getErrorName(result).decode()
+            raise RuntimeError(f"Zstd {operation} failed: {error}.")
+
+
+def create_tensor_codec(name: str, *, level: int = 1) -> TensorCodec:
+    """Construct a direct tensor codec by transport name."""
+    if name == "lz4":
+        return LZ4TensorCodec(acceleration=level)
```
guozhen1997: lz4 `acceleration` is not the same meaning as zstd `level`

docs/source-en/rst_source/concepts/collective.rst
```diff
@@ -48,6 +48,77 @@ With the process groups in place, ``CollectiveGroup`` can perform communications
...
+tensor when the encoded payload is not smaller. When a slot is released, it is
+prioritized over never-used slots, which greedily reuses its workspace buffers.
+For asynchronous sends, the slot remains leased until every payload send
+completes. Received compressed payloads borrow a decoder slot only while they
+are restored; decoder slots have no workspace buffers.
+
+Each ``CollectiveGroup`` owns one ``TensorCodecPool`` and therefore uses one
+codec and level for the lifetime of the group. The wire metadata identifies the
+compressed payload and validates that its codec matches the job-wide
+configuration propagated to the receiving Worker. Compression currently applies
```
guozhen1997: Compression applies to CPU tensors in generic send/recv for a single tensor, a list/dict of tensors, or a dataclass with tensor fields. Pickled Python objects (OBJECT) and send_tensor/recv_tensor / broadcast / GPU-NCCL paths are not compressed.

rlinf/scheduler/worker/worker.py
```diff
@@ -557,6 +559,45 @@ def create_group(
...
+        if not isinstance(config, dict):
+            raise ValueError(
+                "Collective tensor compression configuration must be a mapping."
+            )
+
+        from ..collective import TensorCompressionOptions
+
+        return TensorCompressionOptions.from_dict(config)
+
+    def _resolve_collective_options(self, options: Optional["CollectiveGroupOptions"]):
```
guozhen1997: Enabling job-wide tensor compression currently changes NCCL process-group
initialization, even though compression is a per-transfer CPU-tensor feature.

Worker._resolve_collective_options() injects tensor_compression into
CollectiveGroupOptions on every send/recv. That object is then passed to
CollectiveGroup._init_process_group() → MultiChannelProcessGroup.init() →
AcceleratorUtil.get_accel_pg_options(). On NVIDIA/AMD/MUSA, get_accel_pg_options
does:

    if options is None or options.is_empty_options():
        return None

is_empty_options() compares the whole dataclass to CollectiveGroupOptions(),
whose tensor_compression default is None. After injection the comparison is
always False, so the first GPU communication builds ProcessGroupNCCL.Options
with Megatron-style defaults (cga_cluster_size=4, max_ctas=32, min_ctas=1)
instead of leaving pg_options=None (PyTorch/NCCL defaults). If send and recv
workers do not get the same compression env, the two sides can also initialize
NCCL with different option objects.

CollectiveGroupOptions is documented as taking effect only when two workers
first communicate. Compression is per object transfer and should not ride that
path.

Please keep process-group options independent of compression. Minimal fix —
ignore tensor_compression in the emptiness check:

    def is_empty_options(self) -> bool:
        """Return True if no accelerator process-group options are set."""
        return (
            self.accel_cluster_size is None
            and self.accel_max_ctas is None
            and self.accel_min_ctas is None
            and not self.is_high_priority_stream
            and not self.use_ring_broadcast
        )

Stronger fix, also strip compression before PG init so a later field cannot
regress this:

    def _init_process_group(self, options=None):
        pg_options = (
            replace(options, tensor_compression=None) if options is not None else None
        )
        ...
        self._mc_group.init(..., options=pg_options)

Please add a unit test that CollectiveGroupOptions(tensor_compression=...) is
still empty for PG purposes, or that get_accel_pg_options returns None when
only tensor_compression is set.

rlinf/scheduler/collective/collective_group.py
```diff
@@ -2172,70 +2331,101 @@ def _send_tensor_list(
...
+                                async_op=async_op,
+                            )
+                            if sent_work is not None:
+                                async_works.append(sent_work)
+        except BaseException:
+            if compression_lease is not None:
+                compression_lease.release()
+            raise
+
+        if not async_op:
```
guozhen1997: The AsyncWorkGroup + “hold the compression lease until async payload sends
complete” machinery is unreachable from Worker.send / CollectiveGroup.send.

CollectiveGroup.send(async_op=True) only enqueues _atomic_send on the work
queue. _atomic_send always calls _send_tensor_list / _send_tensor_dict /
_send_tensor_dataclass without async_op, so the inner sends stay synchronous
(async_op=False). That means:

- the lease is always released on the `if not async_op:` path
- send_work.then(compression_lease.release) never runs
- AsyncWorkGroup is never returned to Worker.send callers

That happens to be correct for today’s call graph (the compressed workspace
stays alive for the duration of the blocking Gloo send). The extra async
composition is still misleading: it looks as if inner async send + delayed
lease release is wired up, but it is not.

It is also fragile if someone later passes async_op through:

- AsyncWorkGroup.then() calls first_work.then(...). AsyncCollWork.then() on
  main asserts len(self._works) == 1.
- _send_tensor_dataclass does
      works = [w for w in (tensor_work, skeleton_work) if w is not None]
      return AsyncWorkGroup(works) if len(works) > 1 else works[0]
  which IndexErrors when both works are None.

Please pick one and delete the other half:

1) Simplify (matches actual behavior): drop AsyncWorkGroup and the
   lease-until-async-complete branch. Acquire/release the lease only around
   the synchronous _send_tensor_list body. Worker.send(async_op=True) already
   serializes that work on CollectiveWorkQueue.

2) Actually wire it: pass the outer async_op into _send_tensor_list, keep the
   lease until every payload work completes, and make the dataclass return
   robust:

      works = [w for w in (tensor_work, skeleton_work) if w is not None]
      if not works:
          return None
      return AsyncWorkGroup(works) if len(works) > 1 else works[0]

If you keep AsyncWorkGroup.then(), do not assume first_work.then() accepts a
multi-work AsyncCollWork; wait for all works explicitly (the current
wait_for_remaining helper is fine only when first_work.then() succeeds).

rlinf/utils/tensor_codec.py
anHappyDog: Done. The documentation now clarifies that level maps to LZ4 acceleration for lz4, and to the compression level for zstd.

docs/source-en/rst_source/concepts/collective.rst
anHappyDog: Compression applies only to generic tensor-object send/recv paths. Pickled Python objects are typically small, so compression overhead is usually not worthwhile and they remain uncompressed. GPU tensors use NCCL/IPC transport and remain uncompressed. send_tensor/recv_tensor and broadcast also remain uncompressed because their tensor sizes are known in advance; supporting variable-sized compressed payloads would require an additional GLOO metadata exchange. Generic send/recv is the dominant path, so we keep compression scoped there.

rlinf/scheduler/worker/worker.py
anHappyDog: Done. Tensor compression options are removed from collective group options.

rlinf/scheduler/collective/collective_group.py
anHappyDog: Done, fixed.

rlinf/scheduler/collective/collective_group.py
```diff
@@ -1365,6 +1394,130 @@ def _partition_tensors(
...
+                pool_options.codec,
+                pool_options.level,
+                pool_options.max_inflight,
+            ) != (options.codec, options.level, options.max_inflight):
+                raise ValueError(
+                    "A CollectiveGroup cannot use multiple tensor codec configurations."
+                )
+            return self._tensor_codec_pool
+
+    def _prepare_tensor_compression(
```
thereAreDemonsNearby: Naming: `_prepare_tensor_compression` implies this method only prepares or validates compression inputs/parameters, but it actually performs the compression directly. Consider renaming it to match its behavior.

rlinf/scheduler/collective/collective_group.py
```diff
@@ -1365,6 +1394,130 @@ def _partition_tensors(
...
+            or not any(
+                tensor.is_cpu
+                and tensor.numel() * tensor.element_size()
+                >= compression_options.min_bytes
+                for tensor in tensors
+            )
+        ):
+            return cpu_tensors, None, None
+
+        lease = self._get_tensor_codec_pool(
```
thereAreDemonsNearby: Consider renaming "lease" to match 'decoder' in _recv_cpu_tensor_payloads.

anHappyDog: Done, renamed to `compressor`.

rlinf/scheduler/collective/collective_group.py
anHappyDog: Done, renamed to `_compress_cpu_tensor`.

guozhen1997:
Could you provide the performance gains achieved after enabling tensor compression?

anHappyDog:
> Could you provide the performance gains achieved after enabling tensor compression?


with `libero_spatial_ppo_openpi` tested, actor and rollout take a 4-card machine,  env take a 4-card machine, 100Mbps:

image

More data will be added.

docs/source-en/rst_source/concepts/collective.rst
```diff
@@ -48,6 +48,66 @@ With the process groups in place, ``CollectiveGroup`` can perform communications
...
+         enabled: true
+         codec: lz4
+         level: 1
+         min_bytes: 65536
+         max_inflight: 2
+
+``tensor_compression`` is disabled when this block is absent or when
+``enabled: false``. Its fields are:
+
+- ``codec``: ``lz4`` or ``zstd``.
```
FxxxxU: Add a description of 'lz4' and 'zstd'; I really don't understand what the parameters mean here.

docs/source-zh/rst_source/concepts/collective.rst
```diff
@@ -53,6 +53,54 @@
...
+       tensor_compression:
+         enabled: true
+         codec: lz4
+         level: 1
+         min_bytes: 65536
+         max_inflight: 2
+
+未配置 ``tensor_compression`` 段，或设置 ``enabled: false`` 时，压缩关闭。各字段含义如下：
+
+- ``codec``：``lz4`` 或 ``zstd``。
```
FxxxxU: Same as above.

docs/source-en/rst_source/concepts/collective.rst
anHappyDog: done. short introduction added.

docs/source-zh/rst_source/concepts/collective.rst
anHappyDog: done. short introduction added.

FxxxxU (approved):

thereAreDemonsNearby (approved):

rlinf/scheduler/collective/collective_group.py
```diff
@@ -1365,6 +1391,151 @@ def _partition_tensors(
...
+        preserve the original CPU tensors. Codecs are returned as soon as CPU
+        compression completes; only the returned buffers span the wire transfer.
+        """
+        compression_options = self._tensor_compression
+        if compression_options is None or not compression_options.enabled:
+            return cpu_tensors, None, []
+
+        codec_pool = self._worker._get_tensor_codec_pool()
+        assert codec_pool is not None
+        if not any(compression_options.should_compress(tensor) for tensor in tensors):
```
guozhen1997: `_compress_cpu_tensors` constructs the codec pool before `should_compress`. With enabled: true, the first generic tensor send loads liblz4/libzstd even when every tensor is too small or float32. Missing system libs then fail all tensor sends, not only compressible ones.
Suggested fix: Run the should_compress check first; only then call _get_tensor_codec_pool(). Add a unit test: enabled + ineligible tensors must not create the pool / load the library.

anHappyDog: done.

guozhen1997 (approved):

rlinf/scheduler/collective/collective_group.py
```diff
@@ -109,9 +114,14 @@ class CollectiveGroupOptions:
     rollout ranks)."""
 
     def is_empty_options(self) -> bool:
-        """Check if the options are empty."""
-        empty_options = CollectiveGroupOptions()
-        return self == empty_options
+        """Return whether accelerator process-group options are unset."""
```
andylin-hao: Why?

rlinf/scheduler/collective/collective_group.py
```diff
@@ -376,41 +389,50 @@ def _atomic_send(
 
         if object_type == CollectiveGroup.TENSOR:
             # Out-of-place tensor send/recv is done via tensor list send/recv with a list of one tensor
-            return self._send_tensor_list(
```
andylin-hao, line 379: Why?

rlinf/scheduler/collective/collective_group.py
```diff
@@ -2136,106 +2307,107 @@ def _send_tensor_list(
         tensors: list[torch.Tensor],
         comm_id: int,
         tensor_data: TensorData,
-        async_op: bool = False,
```
andylin-hao, line 2139: Why?

rlinf/scheduler/collective/collective_group.py
```diff
@@ -2136,106 +2307,107 @@ def _send_tensor_list(
...
+                            )
+        finally:
+            for buffer in payload_buffers:
+                buffer.release()
 
     def _recv_tensor_list(
         self,
         comm_id: int,
         work: Optional[AsyncFuncWork] = None,
+        options: Optional[CollectiveGroupOptions] = None,
```
andylin-hao: Why add this throughout the code but never use it?

anHappyDog: deleted.

rlinf/scheduler/collective/collective_group.py
anHappyDog: restored.

rlinf/scheduler/collective/collective_group.py
anHappyDog: restored.

rlinf/scheduler/collective/collective_group.py
anHappyDog: restored.

andylin-hao (approved):

PR #1484, opened 2026-08-17
feat(libero): skip intermediate renders

GanDing:
### Description
Adds an opt-in `env.train.skip_intermediate_renders` flag for the LIBERO environment that
disables offscreen camera rendering on the substeps of an action chunk whose observations
are never consumed.

`LiberoEnv.chunk_step()` executes `num_action_chunks` physics substeps per call (10 for
`pi0.5`), and each substep renders both cameras (`agentview_image`,
`robot0_eye_in_hand_image`) at 256x256. Downstream, only the *last* substep's observation is
fed back to the policy — `EnvWorker.env_interact_step()` takes `obs_list[-1]`. The
intermediate renders are pure overhead unless an external reward model or video recording
consumes them. When the training hardware doesn't have gpu for the rendering, this feature help
saving training time.

Changes:

- `rlinf/envs/libero/venv.py`
  - `_set_camera_rendering()` toggles the robosuite camera observables via
    `Observable.set_enabled()`, which suppresses the `sim.render()` call inside the sensor
    without reconfiguring or resetting the environment.
  - New `set_camera_rendering` command in the subprocess worker loop, plumbed through
    `ReconfigureSubprocEnvWorker` and `ReconfigureSubprocEnv`.
  - `LIBERO_CAMERA_OBS_NAMES` is now shared between `venv.py` and `libero_env.py`.
- `rlinf/envs/libero/libero_env.py`
  - `chunk_step()` disables rendering for substeps `0..chunk_size-2` and re-enables it for
    the final substep, then restores rendering unconditionally before auto-reset.
  - When a camera observable is disabled, robosuite drops the key from the observation dict
    entirely, so a per-env camera cache (`_cached_camera_obs`) holds the last real frame and
    is patched back in for skipped substeps. The cache is primed on `reset()` and refreshed
    on every rendered substep; a cold cache forces a render on substep 0.
  - `step()` gains an internal `_skip_obs_wrap` argument so the observation tensor pipeline
    can be bypassed when the result is discarded.
- `examples/embodiment/config/libero_10_ppo_openpi_pi05.yaml` — enables the flag for
  training. Eval is deliberately left untouched so evaluation videos stay intact.

The flag defaults to `False`, so behavior is unchanged for every config that does not opt in.

### Motivation and Context
Offscreen rendering is one of the larger per-substep costs in the LIBERO env worker, and with
a chunk size of 10 we currently pay it 10x per `chunk_step()` while using exactly one of the
resulting frames. Reducing that to one render per chunk is the cheapest available reduction in
`time/env/env_interact_step`, which is on the critical path for embodied RL throughput.

### How has this been tested?
- `tests/unit_tests/test_skip_render.py` — new unit test covering the `should_render` schedule
  in `chunk_step()`: asserts intermediate substeps are disabled, that rendering is re-enabled
  for the final substep, and that no toggling happens when the flag is off.
- Manual training run on the remote server with
  `examples/embodiment/config/libero_10_ppo_openpi_pi05.yaml` (LIBERO-10, `pi0.5`,
  `total_num_envs: 64`, `num_action_chunks: 10`, 256x256 cameras), comparing
  `time/env/env_interact_step` with the flag on and off.

Measured speedup so far is materially below the expected ~2x. Numbers and the profiling
breakdown to follow in a comment.

Test env: total_num_envs=64, rollout_epoch=2

Before the change:

image

After the change:

image

The growing trend on SR proves the change doesn't break the training:

image

### Types of changes

### Checklist:


rlinf/envs/libero/libero_env.py
```diff
@@ -931,16 +956,30 @@ def chunk_step(self, chunk_actions):
         raw_chunk_terminations = []
         raw_chunk_truncations = []
         for i in range(chunk_size):
+            should_render = True
+            if self.skip_intermediate_renders:
+                cache_is_cold = any(c is None for c in self._cached_camera_obs)
+                should_render = i == chunk_size - 1 or (cache_is_cold and i == 0)
+                self.env.set_camera_rendering(should_render)
```
guozhen1997: With the flag on, chunk_step calls it once per substep plus once at the end, even when the desired state did not change.
Suggest toggling only on transitions (disable once before the skipped range, enable once for the last substep). Fan out with send-all / recv-all. Or better: fold enabled into the existing step command so skipped substeps need no extra IPC. Wrap the loop in try/finally so rendering is always restored.

rlinf/envs/libero/libero_env.py
```diff
@@ -931,16 +956,30 @@ def chunk_step(self, chunk_actions):
...
             extracted_obs, step_reward, terminations, truncations, infos = self.step(
-                actions, auto_reset=False
+                actions, auto_reset=False, _skip_obs_wrap=not should_render
             )
+            if self.skip_intermediate_renders:
+                if should_render:
+                    self._refresh_camera_cache(self.current_raw_obs)
+                else:
+                    self.current_raw_obs = self._apply_cached_camera_obs(self.current_raw_obs)
+                    extracted_obs = self._wrap_obs(self.current_raw_obs)
```
guozhen1997: Do not wrap or cache-patch intermediate substeps; append None. Keep the camera cache only for the next rendered step if needed. If a wrapper must see a full obs_list, document that and keep skip off when video_cfg.save_video or CollectEpisode is on.

examples/embodiment/config/libero_10_ppo_openpi_pi05.yaml
```diff
@@ -73,6 +73,7 @@ env:
     max_episode_steps: 480 # max episode steps for truncation
     max_steps_per_rollout_epoch: 480
     reward_coef: ${algorithm.reward_coef}
+    skip_intermediate_renders: True
```
guozhen1997, line 76: The recipe sets env.train.skip_intermediate_renders: True but does not turn off train video. RecordVideo wraps chunk_step and records every non-None obs. Because skipped obs are filled with the last real frame, train MP4s will repeat one image for 9/10 substeps. Either return None for skipped obs (videos then have one frame per chunk, as Behavior documents), or set env.train.video_cfg.save_video: False in this recipe, or force-disable skip when save_video is on.

andylin-hao (commented):
A few notes on the obs contract and on where the expected speedup is going. Full review in the PR thread; inline comments cover the high / medium-high items.

rlinf/envs/libero/libero_env.py
andylin-hao: `BehaviorEnv` already implements skipped-substep observations, and the contract there is to append `None` (`rlinf/envs/behavior/behavior_env.py:206,660`). `RecordVideo` was written against it and filters `None` entries out (`rlinf/envs/wrappers/record_video.py:381`). Fabricating a stale frame instead is more work, and it is silently wrong rather than absent.

It also means `_wrap_obs` still runs on every substep, so `_skip_obs_wrap` currently saves nothing.

**Fix:** `obs_list.append(None)` when `should_render` is False, and drop `_cached_camera_obs`, `_get_camera_cache`, `_refresh_camera_cache`, `_apply_cached_camera_obs`. Every consumer already reads `obs_list[-1]` (`env_worker.py:495`, `:580`, `:634`).

rlinf/envs/libero/libero_env.py
```diff
@@ -920,6 +923,28 @@ def step(self, actions=None, auto_reset=True):
...
         )
 
+    def _get_camera_cache(self, obs):
+        return {name: np.array(obs[name], copy=True) for name in LIBERO_CAMERA_OBS_NAMES}
+
+    def _refresh_camera_cache(self, obs_list, env_idx=None):
+        if env_idx is None:
+            env_idx = np.arange(self.num_envs)
+        for local_idx, global_idx in enumerate(env_idx):
+            self._cached_camera_obs[global_idx] = self._get_camera_cache(obs_list[local_idx])
```
andylin-hao: Two full-frame copies per env per skipped substep: `copy=True` here, and again in `_apply_cached_camera_obs`. At 64 envs x 9 skipped substeps x 2 copies x 2x256x256x3 that is roughly 450 MB of memcpy per `chunk_step` — likely a good part of the missing speedup.

Neither copy is needed: the cached arrays are never mutated (`get_libero_image` only takes a reversed view, and `_wrap_obs` already clones).

**Fix:** dropping the cache entirely (see the `chunk_step` comment) removes both.

rlinf/envs/libero/venv.py
```diff
@@ -256,3 +276,11 @@ def reconfigure_env_fns(self, env_fns, id=None):
...
         for j, i in enumerate(id):
             self.workers[i].reconfigure_env_fn(env_fns[j])
+
+    def set_camera_rendering(self, enabled: bool, id=None):
+        self._assert_is_not_closed()
+        id = self._wrap_id(id)
+        if self.is_async:
+            self._assert_id(id)
+        for i in id:
+            self.workers[i].set_camera_rendering(enabled)
```
andylin-hao: This does a blocking `send()` + `recv()` round-trip per worker, serialized — unlike `BaseVectorEnv.step`, which sends to all workers first and only then receives (`rlinf/envs/venv/venv.py:836-845`). It sits on the per-substep hot path.

`chunk_step` also calls it `chunk_size + 1` times per chunk (9x False, 1x True, plus the trailing True) when 2 toggles are enough.

**Fix:** split into a send loop followed by a recv loop, and only send when the value actually changes.

examples/embodiment/config/libero_10_ppo_openpi_pi05.yaml
andylin-hao, line 76: This corrupts training video for this config. `env/libero_10.yaml` sets train `video_cfg.save_video: True` and `info_on_video: True`, and nothing here overrides it, so `EnvWorker` wraps the train env in `RecordVideo` (`env_worker.py:411`). With fabricated frames, 9 of every 10 recorded frames are duplicates of the previous one while the reward/termination overlay keeps advancing.

**Fix:** returning `None` for skipped substeps makes `RecordVideo` drop them automatically (that is what the behavior env does). Otherwise set `env.train.video_cfg.save_video: False` here and note it in the PR.

guozhen1997 (approved):

andylin-hao (approved):

PR #1492, opened 2026-08-21
perf(nft): speed up video nft RL

thereAreDemonsNearby:
### Description

This PR cuts GPU↔host transfer and GPU-memory overhead during off-policy NFT training by routing model offloads through pinned memory, deferring the training-model load, and caching the reference model; it also selects the correct GPU build of PaddlePaddle for the OCR reward.

- Add `rlinf/utils/pinned_offload.py`: a context manager that patches `Tensor.to("cpu")` to allocate a pinned buffer, maximising PCIe bandwidth for GPU→CPU weight offload. The CPU→GPU direction is left untouched (a pinned source already runs at full bandwidth).
- NFT FSDP actor (`fsdp_nft_policy_worker.py`): offload/load the un-sharded `rollout_model_state_dict` alongside FSDP params so it no longer strands on GPU during `recv_rollout_trajectories`; defer loading the training model until after `_precompute_nft_training_inputs` when `nft_tau < 1` (precompute offloads the model anyway, so an early load is a wasted GPU→CPU→GPU round trip); cache the reference model across iterations in `_recompute_v_old` instead of rebuilding it each step (only `load_state_dict` + `to(device)` per call).
- HF rollout worker (`huggingface_worker.py`): wrap `offload_model`'s `.to("cpu")` calls in `pinned_offload_context`.
- Install (`requirements/install.sh`, `requirements/embodied/models/diffusion.txt`): install the `paddlepaddle-gpu` build matching the system CUDA version (CUDA 12 / 11 wheel index, CPU fallback) instead of pinning the CPU-only `paddlepaddle==2.6.2` wheel, so the GPU OCR reward can run on-device.

### Motivation and Context

During off-policy NFT training of Wan2.2-TI2V-5B, the HF rollout worker's `offload_model` (a pageable `model.to("cpu")` of the 5B model) takes ~10 s per call, leaving the GPU idle for that window every step. The un-sharded `rollout_model_state_dict` also stayed resident on GPU while the FSDP model was offloaded, and the reference model was rebuilt from scratch every step — both wasting GPU memory and PCIe bandwidth. The CPU-only PaddlePaddle wheel additionally forced the OCR reward onto the host. This PR routes those offloads through pinned memory to maximise PCIe bandwidth and shrink the GPU-idle gap, defers/caches the redundant model work, and selects the correct GPU paddle build.

### How has this been tested?

- `ruff check` and `ruff format` on the three changed Python files: clean.
- Exercised the off-policy NFT actor path end-to-end on Wan2.2-TI2V-5B (4×A800-80GB, 50 epochs per run). Training is stable and the pinned-offload / deferred-load / cached-ref-model paths execute without error. Steady-state per-step times (steps 30–50): image-OCR ~75–99 s/step, video-OCR ~280–474 s/step; the `pp2` overlap and GPU-OCR variants are faster, as expected. One `pp2`+CPU-OCR video config OOMs on 460 GB RAM at ~epoch 17 — a config-level memory over-provision, not a code regression (the same config with GPU-OCR completes all 50 epochs).

Machine: 4xA800

Config name format: `wan22_ti2v_5b_nft_<task>_<ocr_device>_pp<pipeline_stages>`, where `ocr` = image OCR (`num_frames=1`) and `video_ocr` = video OCR (`num_frames=5`, `frame_interval=4`); `cpuocr` / `gpuocr` = PaddleOCR reward on CPU (`use_gpu: False`) / GPU (`use_gpu: True`); `pp1` / `pp2` = rollout `pipeline_stage_num` 1 (no overlap) / 2 (two-stage overlap). For `pp2`, to keep the per-inference batch size unchanged, `total_num_envs` is doubled and `rollout_epoch` is halved (image: 96→192 / 6→3; video: 192→384 / 6→3), so each training step still consumes the same number of samples via two staggered rollout stages.

Key config values:

| config | total_num_envs | rollout_epoch | pipeline_stage_num | reward use_gpu |
|---|---|---|---|---|
| ocr_cpuocr_pp1 | 96 | 6 | 1 | False |
| ocr_cpuocr_pp2 | 192 | 3 | 2 | False |
| ocr_gpuocr_pp1 | 96 | 6 | 1 | True |
| ocr_gpuocr_pp2 | 192 | 3 | 2 | True |
| video_ocr_cpuocr_pp1 | 192 | 6 | 1 | False |
| video_ocr_cpuocr_pp2 | 384 | 3 | 2 | False |
| video_ocr_gpuocr_pp1 | 192 | 6 | 1 | True |
| video_ocr_gpuocr_pp2 | 384 | 3 | 2 | True |

Steady-state per-step times (steps 30–50, 21 samples each):

| config | avg (s) | min (s) | max (s) |
|---|---|---|---|
| ocr_cpuocr_pp1 | 99.10 | 96.94 | 101.53 |
| ocr_cpuocr_pp2 | 83.24 | 81.46 | 85.72 |
| ocr_gpuocr_pp1 | 77.27 | 75.66 | 79.51 |
| ocr_gpuocr_pp2 | 74.80 | 73.34 | 76.85 |
| video_ocr_cpuocr_pp1 | 473.78 | 472.10 | 475.84 |
| video_ocr_gpuocr_pp1 | 300.38 | 299.35 | 301.85 |
| video_ocr_gpuocr_pp2 | 279.78 | 279.11 | 280.27 |

`video_ocr_cpuocr_pp2` (384 envs + CPU OCR reward) is omitted — it OOMs before reaching step 30.

### Types of changes

### Checklist:


rlinf/utils/pinned_offload.py
```diff
@@ -0,0 +1,66 @@
...
+
+"""Pinned-memory offload context manager for accelerator-agnostic model offload."""
+
+from contextlib import contextmanager
+
+import torch
+
+
+@contextmanager
+def pinned_offload_context():
```
anHappyDog, line 23: `pinned_offload_context` monkey-patches `torch.Tensor.to` process-wide and misinterprets the dtype and tensor overloads as devices, causing valid calls to raise `TypeError`. Overlapping contexts can also restore the wrong function and leave the patch installed globally.  It might be bettor to replace it with a scoped helper or make overload handling and restoration concurrency-safe.

requirements/install.sh
```diff
@@ -2214,13 +2210,59 @@ install_dreamzero_model() {
     esac
 }
 
+install_paddle_gpu() {
+    # PaddlePaddle GPU build. The PyPI ``paddlepaddle`` package is CPU-only;
+    # the GPU build lives on the PaddlePaddle wheel index as
+    # ``paddlepaddle-gpu``.  Select the CUDA-12 index when the system CUDA
+    # major version is >= 12, otherwise fall back to the CUDA-11 index, and
+    # finally to the CPU package if no CUDA is detected.
+    local paddle_ver="2.6.2"
```
anHappyDog, line 2219: `paddlepaddle-gpu==2.6.2` does not match the CUDA 12 build published as `2.6.2.post120`. Since `-f` only adds candidates and keeps PyPI enabled, pip may install the plain 2.6.2 wheel successfully and never try the fallback. Please map each supported CUDA version to the correct Paddle version/index and verify the installed CUDA build.

rlinf/workers/actor/fsdp_nft_policy_worker.py
```diff
@@ -267,21 +328,28 @@ def _recompute_v_old(
...
+                ref_config = getattr(ref_model, "config", None)
+                # Compiling the reference model adds repeated compilation and cache overhead.
+                if hasattr(ref_config, "compile_transformer_forward"):
+                    ref_config.compile_transformer_forward = False
+                ref_model.eval()
+                ref_model.requires_grad_(False)
+                self._cached_ref_model = ref_model
+            else:
+                ref_model = self._cached_ref_model
+
```
anHappyDog, line 350: `_cached_ref_model` stores the complete object returned by `model_provider_func()` for every NFT model. For pipeline-based models such as Wan and SD3, this also retains auxiliary components such as the VAE and text/image encoders, although `_recompute_v_old` only uses the transformer path with precomputed inputs. Could we cache only the components required for NFT forward, possibly through a model-specific reference-model provider?

thereAreDemonsNearby: In the previous implementation using model_provider_func, every time a new model was created, unused components of pipeline-based models such as Wan/SD3 would also be loaded, so this optimization is strictly better than the old implementation. On the other hand, for the sake of code simplicity, we should not introduce special-case type checks for models here either.

requirements/install.sh
thereAreDemonsNearby: You're right — my previous approach would install the CUDA 11.8 version of PaddlePaddle. I'll change it to the post120 version.

anHappyDog:
The overload handling is fixed, but overlapping contexts are still unsafe. With A enter → B enter → A exit → B exit, B restores A’s patched function, leaving torch.Tensor.to patched globally. Please make restoration concurrency-safe or avoid the global monkey patch.

rlinf/workers/actor/fsdp_nft_policy_worker.py
```diff
@@ -267,21 +328,28 @@ def _recompute_v_old(
...
             ref_model.to(self.device)
+            ref_model.load_state_dict(self.get_rollout_state_dict(), strict=False)
```
anHappyDog: rollout_model_state_dict is already on CPU here. ref_model.to(self.device) first transfers the cached weights to GPU, then load_state_dict() transfers the full rollout weights again and overwrites them. Please load the state dict before moving the reference model to GPU to avoid the redundant model-sized H2D transfer.

thereAreDemonsNearby:
> The overload handling is fixed, but overlapping contexts are still unsafe. With A enter → B enter → A exit → B exit, B restores A’s patched function, leaving torch.Tensor.to patched globally. Please make restoration concurrency-safe or avoid the global monkey patch.

On the one hand, in our scenario (rollout model offload or ref model offload), there does not seem to be any multi-threaded concurrency; on the other hand, I don't know how to make a context manager concurrency-safe. If necessary, please provide a tutorial.

rlinf/workers/actor/fsdp_nft_policy_worker.py
thereAreDemonsNearby: Thank you for pointing this out. Done.

anHappyDog:
> > The overload handling is fixed, but overlapping contexts are still unsafe. With A enter → B enter → A exit → B exit, B restores A’s patched function, leaving torch.Tensor.to patched globally. Please make restoration concurrency-safe or avoid the global monkey patch.
> 
> On the one hand, in our scenario (rollout model offload or ref model offload), there does not seem to be any multi-threaded concurrency; on the other hand, I don't know how to make a context manager concurrency-safe. If necessary, please provide a tutorial.

it's better to add some comments on them, noting that this 

> > The overload handling is fixed, but overlapping contexts are still unsafe. With A enter → B enter → A exit → B exit, B restores A’s patched function, leaving torch.Tensor.to patched globally. Please make restoration concurrency-safe or avoid the global monkey patch.
> 
> On the one hand, in our scenario (rollout model offload or ref model offload), there does not seem to be any multi-threaded concurrency; on the other hand, I don't know how to make a context manager concurrency-safe. If necessary, please provide a tutorial.

It's better to add comments on this function, noting that this function is not thread-safe.

requirements/install.sh
```diff
@@ -2281,6 +2281,50 @@ install_dreamzero_model() {
...
+        if pip install "paddlepaddle-gpu==${paddle_ver}" \
+               -f "${paddle_gpu_index_cuda11}" 2>&1; then
+            return 0
+        fi
+        echo "[install.sh] GPU paddlepaddle-gpu install failed; falling back to CPU." >&2
+    fi
+
+    echo "[install.sh] No CUDA detected; installing CPU paddlepaddle==${paddle_ver}."
+    uv pip install "paddlepaddle==${paddle_ver}"
 install_cosmos3_deps() {
```
chenchaoxu7575, line 2328: `install_paddle_gpu()` is missing its closing `}` before this function. This makes `requirements/install.sh` fail `bash -n` with `unexpected end of file`.

rlinf/workers/rollout/hf/huggingface_worker.py
```diff
@@ -34,6 +34,7 @@
 from rlinf.models import get_model
 from rlinf.models.embodiment.base_policy import BasePolicy
 from rlinf.scheduler import Channel, Cluster, Worker, split_channel_message
+from rlinf.utils.pinned_offload import pinned_offload_context
```
chenchaoxu7575, line 37: This import order fails Ruff `I001`. `pinned_offload` should be placed after `obs_compression`.

rlinf/workers/actor/fsdp_nft_policy_worker.py
```diff
@@ -44,9 +45,53 @@ def init_worker(self) -> None:
...
+        offloaded (recv phase), on GPU when the model is loaded (training
+        phase).  Pinned memory is used for the GPU→CPU copy to maximise PCIe
+        bandwidth.
+        """
+        super().offload_param_and_grad(offload_grad)
+        rollout_state = getattr(self, "rollout_model_state_dict", None)
+        if rollout_state:
+            with pinned_offload_context():
+                for key, value in rollout_state.items():
+                    if torch.is_tensor(value) and value.is_cuda:
```
chenchaoxu7575, line 67: This restricts the offload path to CUDA, while `pinned_offload_context()` is described as accelerator-agnostic. Could we use a device-type check instead? The same issue also appears in `load_param_and_grad()`.

rlinf/utils/pinned_offload.py
```diff
@@ -0,0 +1,76 @@
...
+        if dtype is None:
+            dtype = kwargs.get("dtype")
+
+        if (
+            device is not None
+            and dtype is None
+            and torch.device(device).type == "cpu"
+            and self.device.type != "cpu"
+        ):
+            buf = torch.empty_like(self, device="cpu", pin_memory=True)
```
chenchaoxu7575, line 66: These buffers become the tensors’ CPU storage. With the real Wan2.2-5B pipeline, this requires about 22.5 GiB of pinned memory per worker, which is effectively retained throughout training. Could we document this cost? A bounded streaming buffer could be considered as a follow-up.

chenchaoxu7575 (changes requested):
Thanks for the optimization. I tested the exact PR implementation with the official Wan2.2-TI2V-5B Diffusers checkpoint on an NVIDIA L20 over PCIe 4.0 x16, whose theoretical one-way bandwidth is approximately 31.5 GB/s after PCIe encoding overhead.

For full-pipeline GPU-to-CPU offload:

- Cold pageable: 10.18s
- Cold pinned: 10.75s
- Steady-state pageable: 8.72–8.79s
- Steady-state pinned: 0.946s

The steady-state D2H speedup is about **9.2×**, reaching approximately **25.6 GB/s**, or **81% of the theoretical one-way PCIe bandwidth**. GPU memory was successfully released from about **24.2 GB to zero**.

The trade-off is about **22.5 GiB of host pinned memory per Wan worker, which is effectively retained throughout training**.

Please document the pinned-memory lifetime and per-worker host-memory cost in the PR description. A bounded streaming implementation can be left as a follow-up.

Issue #1238, opened 2026-06-03
[Feature]: Add support for pi0-fast


ryanhoangt:
### Feature description

Hi team,

Thanks for the great work.
I'm seeing the repo currently supports $\pi_0$ and $\pi_{0.5}$, but not $\pi_0-FAST$. Does the team have any plan to support this VLA in the near future or welcome a contribution of this VLA?

Thanks. 

### Before submitting a new issue...

- [x] Have you checked relevant issues, FAQs (https://rlinf.readthedocs.io/en/latest/rst_source/faq.html), or asked the chatbot at the top right corner of the [documentation page](https://rlinf.readthedocs.io).

zoeyuchao:
Pi_0 Fast is not on our to-do list. We welcome the community to contribute.

chenxijun1029:
Hi @zoeyuchao , I’m interested in contributing pi0-FAST support for RLinf.

Would you be open to an optional `pi0_fast` / `openpi_fast` backend based on Hugging Face LeRobot’s `PI0FastPolicy`?

My proposed initial scope would be:
- add a dedicated optional install path with a pinned LeRobot dependency;
- support LIBERO first;
- integrate PPO/GRPO rollout with FAST action tokens and token-logprob replay;
- avoid changing the existing `openpi` backend behavior.

If this direction sounds acceptable, I can prepare a short design and start with a minimal PR.

QuanluZhang:
hi @chenxijun1029 , thanks for your interest in contributing pi0-FAST in RLinf. The initial scope looks good. We would suggest showing a good convergence and test success rate in LIBERO before pr merge.

chenxijun1029:
> hi [@chenxijun1029](https://github.com/chenxijun1029) , thanks for your interest in contributing pi0-FAST in RLinf. The initial scope looks good. We would suggest showing a good convergence and test success rate in LIBERO before pr merge.

Ok，then I'll get started.

chenxijun1029:
The PI0-FAST integration is now available in #1493, including baseline parity and GRPO convergence/evaluation results. Feedback on the integration direction would be appreciated.

PR #1493, opened 2026-08-21
feat(embodied): support PI0-FAST GRPO

chenxijun1029:
### Description

This PR adds a reproducible PI0-FAST backend for LIBERO evaluation and token-level GRPO fine-tuning through LeRobot's `PI0FastPolicy`.

The main changes are:

- register `pi0_fast` as an embodied policy backend and convert RLinf observations into the LeRobot input schema;
- keep PI0-FAST's native autoregressive action sequence, decode FAST action tokens, and execute a safe zero action for malformed outputs without resampling;
- cache sampled action tokens and replay them with a single teacher-forcing forward pass during actor updates;
- compute token-level log-probabilities, entropy, PPO ratios, clipping, and GRPO losses with explicit token masks;
- add all-linear LoRA support and an optional FP32-master AdamW implementation for FSDP1 `NO_SHARD` and `FULL_SHARD` LoRA training;
- install PI0-FAST in an isolated Python 3.12 / PyTorch 2.11 runtime with pinned LeRobot, policy, text-tokenizer, and action-tokenizer revisions;
- add LIBERO-Long baseline and two-node GRPO reference configurations, English and Chinese documentation, unit tests, and an embodied E2E workflow;
- preserve neutral PPO diagnostics for reward-filtered empty microbatches (`ratio=1`, zero loss and zero gradient).

The backend is intentionally scoped to PI0-FAST on LIBERO. It does not change the existing OpenPI policy backends or shared LIBERO reset behavior.

### Motivation and Context

Addresses #1238.

RLinf currently has no PyTorch PI0-FAST path that can both execute the pretrained policy and replay sampled action tokens for an on-policy actor update. A direct integration also cannot reuse the default embodied runtime: the validated PI0-FAST implementation requires newer LeRobot, PyTorch, and Transformers versions.

This implementation therefore provides:

1. a pinned and independently reproducible runtime;
2. baseline parity before reinforcement learning;
3. rollout/actor probability alignment for the same sampled token sequence;
4. a token-level objective that avoids multiplying a long action sequence into one unstable PPO ratio;
5. LoRA-based updates that retain the pretrained policy more reliably than the full-parameter experiments used during development.

### How has this been tested?

The integration passed the targeted PI0-FAST unit suite, a 500-episode baseline evaluation, a two-node GRPO training run, and a 500-episode checkpoint evaluation.

#### Unit and lint checks

Validated in the PI0-FAST runtime (`Python 3.12.12`, `PyTorch 2.11.0`, CUDA 12.8, `pytest 9.1.1`):

```bash
pre-commit run --all-files

ruff check \
  rlinf/algorithms/losses.py \
  tests/unit_tests/test_algorithm_losses.py

python -m pytest -q \
  tests/unit_tests/test_algorithm_losses.py \
  tests/unit_tests/test_fp32_master_adamw.py \
  tests/unit_tests/test_fp32_master_adamw_fsdp.py \
  tests/unit_tests/test_huggingface_worker_sampling.py \
  tests/unit_tests/test_pi0_fast_config.py \
  tests/unit_tests/test_pi0_fast_data_pipeline.py \
  tests/unit_tests/test_pi0_fast_install.py \
  tests/unit_tests/test_pi0_fast_lerobot_api.py \
  tests/unit_tests/test_pi0_fast_registration.py \
  tests/unit_tests/test_pi0_fast_replay.py

CUDA_VISIBLE_DEVICES=0,1 \
RUN_FSDP_DISTRIBUTED_TESTS=1 \
torchrun --standalone --nproc_per_node=2 -m pytest -q \
  tests/unit_tests/test_fp32_master_adamw_fsdp.py
```

Result: all pre-commit hooks passed; **80 tests passed and 2 were skipped** in the regular suite. Both skipped FSDP1 `FULL_SHARD` cases then passed under the two-GPU `torchrun` command above for `use_orig_params=False` and `use_orig_params=True`.

The tests cover, among other cases:

- pinned dependency and config composition;
- rollout sampling parameter propagation;
- complete-sequence generation and malformed-sequence fallback;
- teacher-forcing alignment (`[BOS, a1, a2]` predicts `[a1, a2, a3]`) with one forward pass;
- token masks, finite log-probabilities, dual clipping, and neutral empty-mask metrics;
- FP32-master AdamW state handling under FSDP1 `NO_SHARD` and `FULL_SHARD`.

#### Baseline evaluation

The pinned `lerobot/pi0fast-libero` checkpoint was evaluated on 500 ordered, fixed LIBERO-Long episodes with greedy decoding:

| Checkpoint | Episodes | `success_once` | `success_at_end` |
|---|---:|---:|---:|
| Pretrained baseline | 500 | 85.8% | 75.8% |
| GRPO step 361 | 500 | **95.8%** | **82.2%** |

This is a +10.0 percentage-point improvement in `success_once` under the same 500-episode protocol.

#### GRPO training validation

The development run used two nodes / 16 GPUs with:

- 1,024 trajectories per update;
- group size 8;
- actor microbatch size 16 and global batch size 13,312;
- sampling temperature 0.3 and greedy evaluation;
- all-linear LoRA, rank 16;
- 256 fixed evaluation episodes every 10 updates.

Selected periodic evaluation results:

| Training step | Episodes | `success_once` | `success_at_end` |
|---:|---:|---:|---:|
| 10 | 256 | 87.50% | 75.39% |
| 100 | 256 | 95.31% | 82.81% |
| 200 | 256 | 94.53% | 82.42% |
| 300 | 256 | **99.22%** | 85.16% |
| 360 | 256 | **98.44%** | 82.81% |

The final five periodic `success_once` evaluations (steps 320-360) average **96.80%**. This is one development seed, so the peak is evidence that the integration learns, not a multi-seed convergence guarantee.

```mermaid
xychart-beta
    title "PI0-FAST GRPO: LIBERO-Long eval success_once"
    x-axis "Training step" [10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120, 130, 140, 150, 160, 170, 180, 190, 200, 210, 220, 230, 240, 250, 260, 270, 280, 290, 300, 310, 320, 330, 340, 350, 360]
    y-axis "Success rate (%)" 85 --> 100
    line [87.50, 87.89, 86.72, 91.41, 91.02, 91.41, 90.63, 90.63, 90.23, 95.31, 92.19, 95.31, 92.58, 94.53, 95.70, 93.75, 95.31, 92.58, 96.09, 94.53, 94.14, 94.53, 94.14, 96.88, 94.92, 94.53, 95.31, 96.88, 95.70, 99.22, 95.70, 97.27, 94.92, 95.70, 97.66, 98.44]
```

The curve uses 256-episode periodic evaluations. The 500-episode rows above are separate full evaluations and are not pooled with the curve.

#### Pinned artifacts

| Artifact | Revision |
|---|---|
| LeRobot source | `8a74e0ac6d01706d67fddfed682a09d694d9c8c0` |
| `lerobot/pi0fast-libero` | `840f4b503f4c09110421c33c810a85b6684fd658` |
| `google/paligemma-3b-pt-224` | `35e4f46485b4d07967e7e9935bc3786aad50687c` |
| `jadechoghari/tokenizer-lib-mean` | `79ae83e3cbd8786dcb84b628569f8d076ca8151e` |

The PaliGemma tokenizer is gated. Users must accept its Hugging Face terms and authenticate before installation or evaluation.

#### Review notes

- Invalid action sequences are not resampled. They execute a safe zero action and receive the normal environment feedback, preserving on-policy sampling.
- Rollout and actor replay use the same temperature and sampled token history. The first replay before an optimizer update is expected to have a PPO ratio close to one.
- Reward filtering can produce an empty actor microbatch. Such a batch contributes zero loss and zero gradient; its ratio diagnostics use the neutral identity value one.
- FP32-master optimizer support is opt-in and restricted to LoRA parameters. Existing optimizer paths remain unchanged by default.
- Flash Attention remains opt-in because the validated PyTorch 2.11 runtime may not have a compatible prebuilt wheel on every platform.

### Types of changes

### Checklist:


chenxijun1029:
Hi @guozhen1997 @andylin-hao, could one of you help review or route this PI0-FAST integration?

It addresses #1238 and includes baseline parity (85.8%), GRPO training validation, and a 500-episode checkpoint eval (95.8%). The main review areas are token-level rollout/replay semantics, the optional FP32-master FSDP optimizer, and the isolated runtime.

I’m happy to split or simplify the PR if that would make review easier. Thanks!

docs/source-zh/rst_source/examples/embodied/pi0_fast.rst
```diff
@@ -0,0 +1,145 @@
...
+策略语义
+--------------------
+
+PI0-FAST 原生生成完整动作字符串，RLinf 不预先注入 ``Action:`` 前缀。policy mask
+包含模型生成的前缀、动作正文和第一个完整的 ``|`` 结束标记，不包含 padding 和结束
+标记之后的 token。同一条轨迹的所有 token 共享轨迹级 GRPO advantage；loss 先对每条
+序列的有效 token 求平均，再对序列求平均。
+
+非法序列不会重采样，而是执行安全零动作，由环境正常返回失败反馈，并继续参与策略
+目标。这样既保持 on-policy 采样。
```
Lin-xs: ZH “这样既保持 on-policy 采样。” is an unfinished 既…又…; EN is “This keeps sampling on-policy.”
Suggested fix: “...，这样可以保持 on-policy 采样。”

requirements/install.sh
```diff
@@ -302,6 +314,32 @@ parse_args() {
     fi
 }
 
+configure_pi0_fast_runtime() {
```
Lin-xs: Please pass these versions via the script’s native flags (e.g. `--python` for `PYTHON_VERSION`) instead of env vars.

Also: do these constraints have to be followed strictly, or are they just untested? If the latter, we only need to document which combination has been tested.

requirements/install.sh
```diff
@@ -976,6 +1014,34 @@ restore_pyproject() {
     fi
 }
 
+cleanup_install() {
+    restore_pyproject
+    unset_mirror
+}
+
+apply_pi0_fast_dependency_overrides() {
```
Lin-xs: Same issue here. Could you pin the `transformers` version in `pi0_fast.txt`?

requirements/install.sh
```diff
@@ -1910,6 +1972,55 @@ EOF
...
+        maniskill_libero|libero)
+            create_and_sync_venv
+            install_common_embodied_deps
+            install_${ENV_NAME}_env
+            uv pip install "cmake<4"
+            uv pip install -r "$SCRIPT_DIR/embodied/models/pi0_fast.txt"
+            uv pip install \
+                "transformers==${TRANSFORMERS_VERSION:-$PI0_FAST_TRANSFORMERS_VERSION}" \
+                "tokenizers>=0.22,<0.23" \
+                "huggingface-hub>=1.0,<2.0"
```
Lin-xs: Same issue here. Could you pin the version in `pi0_fast.txt`?

rlinf/utils/utils.py
```diff
@@ -341,12 +341,21 @@ def seq_mean_token_sum(values: torch.Tensor, mask: torch.Tensor, dim: int = -1):
...
-    )  # token-mean
-    loss = torch.mean(seq_losses)  # seq-mean
-    return loss
+def seq_mean_token_mean(
+    values: torch.Tensor,
+    mask: torch.Tensor,
+    loss_mask_ratio: torch.Tensor | None = None,
+    dim: int = -1,
+):
+    del loss_mask_ratio
```
Lin-xs: This `loss_mask_ratio` is passed in and then discarded. If it isn’t used, please don’t take it as an argument — just delete it.

rlinf/models/__init__.py
```diff
@@ -350,7 +361,7 @@ def get_model(cfg: DictConfig):
     ):
         model = model.to(Worker.torch_device_type)
 
-    if cfg.is_lora:
+    if cfg.is_lora and SupportedModel(model_type) != SupportedModel.PI0_FAST:
```
Lin-xs: Please delete.

rlinf/models/embodiment/pi0_fast/fast_replay.py
```diff
@@ -0,0 +1,710 @@
...
+# limitations under the License.
+
+from __future__ import annotations
+
+from typing import Any
+
+import torch
+
+
+def _lerobot_constant(name: str, fallback: str) -> str:
```
Lin-xs: Is this necessary when we have pin the version of lerobot?

rlinf/models/embodiment/pi0_fast/__init__.py
```diff
@@ -0,0 +1,446 @@
...
+) -> str:
+    """Download a pinned HF artifact and return its local snapshot directory."""
+    local_files_only = _opt_cfg(model_cfg, "local_files_only", bool)
+    if revision is None and not local_files_only:
+        return str(name_or_path)
+    if not _looks_like_hf_repo_id(str(name_or_path)):
+        return str(name_or_path)
+
+    try:
+        from huggingface_hub import snapshot_download
```
Lin-xs: Please keep all installation in `install.sh` and the docs. Don’t download any repos or assets under `rlinf/`.

requirements/install.sh
```diff
@@ -113,11 +120,14 @@ Targets:
 
 Options (for target=embodied):
     --model <name>         Embodied model to install: ${SUPPORTED_MODELS[*]}.
+                           pi0_fast defaults to Python 3.12.12, torch 2.11.0,
```
Lin-xs: Delete.

requirements/install.sh
```diff
@@ -113,11 +120,14 @@ Targets:
...
+                           and an isolated .venv-pi0-fast runtime.
     --env <name>           Single environment to install: ${SUPPORTED_ENVS[*]}.
 
 Common options:
     -h, --help             Show this help message and exit.
-    --venv <dir>           Virtual environment directory name (default: .venv).
+    --venv <dir>           Virtual environment directory name (default: .venv;
+                           pi0_fast: .venv-pi0-fast).
```
Lin-xs: Please use default name.

.github/workflows/embodied-e2e-tests.yml
```diff
@@ -1541,3 +1541,53 @@ jobs:
...
+
+      - name: PI0-Fast FSDP FULL_SHARD optimizer test
+        timeout-minutes: 10
+        run: |
+          unset PYTHONPATH
+          export REPO_PATH=$(pwd)
+          export RUN_FSDP_DISTRIBUTED_TESTS=1
+          source .venv-pi0-fast/bin/activate
+          torchrun --standalone --nproc-per-node=2 -m pytest -q \
+            tests/unit_tests/test_fp32_master_adamw_fsdp.py
```
Lin-xs: This yaml is meant for e2e tests only. Please drop all unit tests here.

docs/source-en/rst_source/examples/embodied/pi0_fast.rst
```diff
@@ -0,0 +1,153 @@
...
+     - ``840f4b503f4c09110421c33c810a85b6684fd658``
+   * - Text tokenizer
+     - ``google/paligemma-3b-pt-224``
+     - ``35e4f46485b4d07967e7e9935bc3786aad50687c``
+   * - Action tokenizer
+     - ``jadechoghari/tokenizer-lib-mean``
+     - ``79ae83e3cbd8786dcb84b628569f8d076ca8151e``
+
+Baseline evaluation
+-------------------
```
Lin-xs: Please move the evaluation section after the training section.

docs/source-en/rst_source/examples/embodied/pi0_fast.rst
```diff
@@ -0,0 +1,153 @@
...
+     - 75.8%
+
+GRPO fine-tuning
+----------------
+
+Launch the two-node, 16-GPU reference configuration with:
+
+.. code:: bash
+
+   bash examples/embodiment/run_embodiment.sh libero_10_grpo_pi0_fast
```
Lin-xs, line 102: The docs should spell out:

- how to download the model (including a pinned version, if one is required)
- which fields in the config must be updated

Please don’t leave this implicit in the code.

chenxijun1029:
Hi @Lin-xs, I pushed commit `aa329e275` to address the review comments.

The main updates are:

- Removed the PI0-FAST-specific runtime defaults from `install.sh`. The validated runtime is now passed through the existing `--python`, `--torch`, and `--no-flash-attn` flags, and the virtual environment uses the standard `.venv` name.
- Moved the PI0-FAST dependency constraints, including `transformers`, `tokenizers`, and `huggingface-hub`, into `requirements/embodied/models/pi0_fast.txt`. The documentation now describes Python 3.12.12 / PyTorch 2.11.0 as the validated combination instead of installer-enforced defaults.
- Removed the unused `loss_mask_ratio` argument and restored the shared `seq_mean_token_mean` and PPO loss aggregation code to the upstream behavior. PI0-FAST uses `token-mean` in its own configuration, so no PI0-FAST-specific change to the generic loss API is needed.
- Removed unit tests from the embodied E2E workflow; the workflow now only installs the runtime, downloads the pinned artifacts, and runs the PI0-FAST LIBERO E2E case.
- Updated both English and Chinese documentation: fixed the Chinese wording, moved baseline evaluation after training, and added explicit download commands, pinned revisions, and the configuration fields users need to update.


I also re-ran the clean-venv installation checks, the targeted unit suite, the two-GPU FSDP1 FULL_SHARD tests, and `pre-commit run --all-files`; they all passed.

Could you please take another look?

rlinf/hybrid_engines/fsdp/strategy/fsdp.py
```diff
@@ -151,6 +151,12 @@ def wrap_model(self, model: nn.Module, device_mesh: DeviceMesh) -> FSDP:
             param_dtype=param_dtype,
             reduce_dtype=reduce_dtype,
             buffer_dtype=buffer_dtype,
+            cast_forward_inputs=mixed_precision_config.get(
+                "cast_forward_inputs", False
```
guozhen1997: `cast_forward_inputs` default is True.

rlinf/algorithms/losses.py
```diff
@@ -295,20 +297,86 @@ def compute_ppo_actor_loss(
...
-            clipped_ratio_for_metrics, loss_mask_for_metrics
-        ),
+        "actor/clipped_ratio": _masked_mean_or(clipped_ratio_for_metrics, 1.0),
         "actor/dual_cliped_ratio": masked_mean(
             dual_cliped_ratio_for_metrics, loss_mask_for_metrics
         ),
         "actor/approx_kl": approx_kl.detach(),
         "actor/clip_fraction": clip_fraction.detach(),
     }
+    if log_logprob_diagnostics:
```
guozhen1997: The `_masked_stat` boolean-indexes the same mask seven times, each a GPU compaction, and treats “no valid tokens” and “all valid tokens are NaN/Inf” the same by returning 0, so logprob_delta_mean=0 can look healthy while finite_fraction=0. The string-dispatch helper also duplicates _masked_mean_or above, and the extra isfinite checks on new/old logprobs are redundant once raw_log_ratio is finite.
Sugget fix:

**rlinf/utils/utils.py**
```
def masked_reduce(
    values: torch.Tensor,
    mask: Optional[torch.Tensor],
    reducer: Literal["mean", "std", "min", "max"],
    empty_value: Optional[torch.Tensor] = None,
) -> torch.Tensor:
    """Reduce masked values to a scalar.

    Empty masks return ``empty_value`` (zeros with ``values`` dtype/device if
    omitted). Callers that need NaN for "mask is non-empty but all-invalid"
    should substitute after the fact.
    """
    if empty_value is None:
        empty_value = values.new_zeros(())
    if mask is None:
        mask = torch.ones_like(values, dtype=torch.bool)
    else:
        mask = torch.broadcast_to(mask.bool(), values.shape)

    if reducer == "mean":
        value = masked_mean(values, mask)
    elif reducer == "std":
        mean = masked_mean(values, mask)
        value = masked_mean((values - mean) ** 2, mask).sqrt()
    elif reducer == "min":
        value = torch.where(
            mask, values, values.new_full((), float("inf"))
        ).amin()
    elif reducer == "max":
        value = torch.where(
            mask, values, values.new_full((), float("-inf"))
        ).amax()
    else:
        raise ValueError(f"Unsupported reducer: {reducer}")

    return torch.where(mask.any(), value, empty_value)
```

**rlinf/algorithms/losses.py**

```
if log_logprob_diagnostics:
    with torch.no_grad():
        valid = loss_mask_for_metrics
        stat_mask = valid & torch.isfinite(raw_log_ratio)
        valid_count = valid.sum()
        finite_count = stat_mask.sum()
        use_nan = (finite_count == 0) & (valid_count > 0)
        zero = raw_log_ratio.new_zeros(())
        nan = raw_log_ratio.new_full((), float("nan"))

        def _masked_reduce_or_nan(
            values: torch.Tensor, reducer: str
        ) -> torch.Tensor:
            return torch.where(
                use_nan,
                nan,
                masked_reduce(values, stat_mask, reducer, empty_value=zero),
            )

        metrics_data.update(
            {
                "actor/logprob_delta_mean": _masked_reduce_or_nan(
                    raw_log_ratio, "mean"
                ),
                "actor/logprob_delta_std": _masked_reduce_or_nan(
                    raw_log_ratio, "std"
                ),
                "actor/logprob_delta_min": _masked_reduce_or_nan(
                    raw_log_ratio, "min"
                ),
                "actor/logprob_delta_max": _masked_reduce_or_nan(
                    raw_log_ratio, "max"
                ),
                "actor/logprob_delta_abs": _masked_reduce_or_nan(
                    raw_log_ratio.abs(), "mean"
                ),
                "actor/logprob_finite_fraction": torch.where(
                    valid_count > 0,
                    finite_count.float() / valid_count.clamp_min(1),
                    raw_log_ratio.new_ones(()),
                ),
                "actor/logprob_new_mean": _masked_reduce_or_nan(logprobs, "mean"),
                "actor/logprob_old_mean": _masked_reduce_or_nan(
                    old_logprobs, "mean"
                ),
            }
        )
```

guozhen1997 (approved):

andylin-hao (approved):

Issue #1375, opened 2026-07-13
[Troubleshoot]: Accuracy issue of OpenVLA-OFT-SFT-libero-object-traj1 on LIBERO-Object


DearFishi:
### Problem description

I evaluated [Openvla-oft-SFT-libero-object-traj1 ](https://huggingface.co/Haozhan72/Openvla-oft-SFT-libero-object-traj1)on libero-object, and the accuracy is only 25.78%.
[The WAN training documentation states](https://rlinf.readthedocs.io/en/latest/rst_source/examples/embodied/wan.html) that OpenVLA-OFT (LoRA-base) achieves 36.7% accuracy on Object.

### Configuration YAML file

YAML config file:

You can also paste the full config here.
```
defaults:
  - env/libero_object@env.eval
  - model/openvla_oft@rollout.model
  - override hydra/job_logging: stdout

hydra:
  run:
    dir: .
  output_subdir: null
  searchpath:
    - file://${oc.env:EMBODIED_PATH}/config/

cluster:
  num_nodes: 1
  component_placement:
    env,rollout: all

runner:
  task_type: embodied_eval
  logger:
    log_path: "../results"
    project_name: rlinf
    experiment_name: "libero_object_openvlaoft_eval"
    logger_backends: ["tensorboard"]

  max_epochs: 1000
  max_steps: -1

  only_eval: True
  val_check_interval: -1
  save_interval: -1

  resume_dir: null
  ckpt_path: null

env:
  group_name: "EnvGroup"

  eval:
    rollout_epoch: 4
    total_num_envs: 64
    auto_reset: True
    ignore_terminations: True
    max_episode_steps: 512
    max_steps_per_rollout_epoch: 512
    group_size: 1
    use_fixed_reset_state_ids: True
    reset_gripper_open: False
    is_eval: True
    video_cfg:
      save_video: True
      video_base_dir: ${runner.logger.log_path}/video/eval

rollout:
  sampling_params:
    do_sample: True
    temperature_train: 1.6
    temperature_eval: 1.6
    top_k: -1
    top_p: 1.0
    repetition_penalty: 1.0

  group_name: "RolloutGroup"
  generation_backend: "huggingface"
  enable_offload: False
  pipeline_stage_num: 1

  model:
    model_path: "/workspace/Openvla-oft-SFT-libero-object-traj1/" 
    model_type: "openvla_oft"
    unnorm_key: libero_object_no_noops
    max_prompt_length: 128
```

### Log file

[...]

PR #1496, opened 2026-08-22
fix(libero): pin mujoco to the era-faithful 3.3.0 for embodied LIBERO

### LINKED ISSUE #1460: [Bug]: Tighten the mujoco pin for embodied LIBERO (the upper-bound-only constraint admits >=3.4.0, which breaks libero_spatial task 5 and makes the published GRPO numbers unreproducible)
### Affected component

The mujoco version constraint of the embodied LIBERO environment, in both places it lives:
- v0.x `requirements/install.sh`: `install_libero_env` runs `uv pip install "mujoco<=3.9.0"` (cap added in #1302; a fresh install resolves to 3.9.0)
- main: the `rlinf-libero` package metadata requires `mujoco>=3.0.0,<3.9.0` (a fresh install resolves to 3.8.1)
Both admit mujoco>=3.4.0, which silently breaks the libero_spatial task-5 initial state (details in the description).

## Summary

The embodied LIBERO env constrains mujoco only from above: v0.3's `install_libero_env` runs `uv pip install "mujoco<=3.9.0"` (cap added in #1302; a fresh install resolves to 3.9.0), and current main installs the `rlinf-libero` package, whose metadata requires `mujoco>=3.0.0,<3.9.0` (resolves to 3.8.1). Both land inside the broken range. From **mujoco 3.4.0** (box-box collision bugfix) the stored init states of **libero_spatial task 5** no longer settle into the task premise: inside `reset()`, before the first observation, the bowl ends up tilted on the ramekin's rim instead of seated on it. Result: the published `RLinf-OpenVLAOFT-GRPO-LIBERO-spatial` numbers can't be reproduced on a fresh environment through no fault of the checkpoint.

## Evidence (same GRPO checkpoint, greedy, 500 episodes, only mujoco swapped)

| mujoco | suite | task 5 |
|---|---|---|
| 3.2.7 | 92.8 | 98% |
| 3.3.0 | 94.2 (≈ model-card 94.76) | 98% |
| 3.8.1 | 85.0 | **12%** |
| 3.9.0 | 86.0 | **19%** |

3.9.0 is what a fresh v0.3 install resolves to; 3.8.1 is what current main resolves to via `rlinf-libero`. A same-physics A/B (keep 3.8.1, swap only task-5 init file for a re-settled version) restores task 5 to 49/50; the whole effect is the broken initial state, not in-episode physics. Note the checkpoints were trained/evaluated in the 3.3.x era, so **3.3.x is the era-faithful choice**.

Related sighting in this repo: in #1375 a maintainer already advises a user to downgrade to mujoco 3.3.2, citing the separate visual regression (Lifelong-Robot-Learning/LIBERO#88). This report adds the physics channel, with the exact 3.4.0 boundary and end-to-end causal isolation. Both channels point the same way: 3.3.x is best enforced by the install constraint itself rather than by per-user advice in comments.

## Suggested change

Pin `mujoco>=3.3,<3.4` for the LIBERO embodied environment, in both places the constraint lives (the `rlinf-libero` package metadata on main, and the direct `pip install` in v0.x `install.sh`). We run v0.3 with 3.3.0. This would simply bring LIBERO in line with the other mujoco-based envs, which `install.sh` already hard-pins (robocasa365 → `mujoco==3.3.1`, roboverse → `mujoco==3.3.7`), and it only touches the LIBERO env's own install path. One caveat worth stating: if any newer checkpoints were trained under mujoco>=3.4, the era-faithful version differs per model family — which is one more argument for the long-term fix (version-independent regenerated init states, see the write-up below). Two more version cliffs worth documenting while at it: mujoco 3.10.0 removes `MjData.qM` (hard-breaks robosuite 1.4.1), and 3.11 is fully incompatible.

Root-cause write-up with a no-policy reproduction probe: https://github.com/huggingface/lerobot/issues/4390.

### Configuration YAML file

N/A: dependency-resolution issue, no YAML involved; evidence runs use the stock openvla-oft LIBERO eval config

### Log file

N/A: no runtime error (the breakage is silent). Machine evidence from the settle probe
(same 50 official init states, only the mujoco wheel swapped):

{"mujoco": "3.3.0", "suite": "libero_spatial", "task_id": 5, "settle_steps": 250, "n_inits": 50, "n_slid_off": 50, "xy_cm_mean": 1.45, "xy_cm_max": 1.45, "tilt_deg_mean": 16.0, "tilt_deg_max": 16.0}
{"mujoco": "3.8.1", "suite": "libero_spatial", "task_id": 5, "settle_steps": 250, "n_inits": 50, "n_slid_off": 50, "xy_cm_mean": 3.72, "xy_cm_max": 3.72, "tilt_deg_mean": 10.5, "tilt_deg_max": 10.5}

(1.45 cm = bowl seated on the ramekin with the pose's natural lean; 3.72 cm = bowl tilted on the rim, half overhanging.
Values are bit-identical across all 50 init states per version.
n_slid_off just applies the probe's default 1 cm xy threshold, which both rows exceed; the discriminating signal is the drift magnitude and the final pose.)

### Environment

python -V: Python 3.11.14
uv pip list (relevant subset):
  mujoco 3.3.0 (pinned manually; swapped to 3.8.1 / 3.9.0 for the broken-side evidence runs)
  robosuite 1.4.1
  torch 2.6.0+cu124
  openvla-oft 0.0.1
nvidia-smi: 3x NVIDIA A800 80GB PCIe, driver 595.71.05, CUDA 13.2
RLinf version: v0.3 (tag; installed via requirements/install.sh embodied --model openvla-oft --env libero)
Docker image tag: N/A (bare-metal uv venv)

### Other reproduction info

_No response_

AmirF194:
### Description

Pin `mujoco==3.3.0` (`--no-deps`, mirroring `install_robocasa365_env`/`install_roboverse_env`) right after each `rlinf-libero` install in `install_libero_env`, `install_liberopro_env`, and `install_liberoplus_env`. `install_maniskill_libero_env` calls `install_libero_env` and inherits the fix.

### Motivation and Context

`rlinf-libero`'s package metadata caps mujoco only from above (`mujoco<3.9.0,>=3.0.0`), so a fresh `uv pip install rlinf-libero` resolves to 3.8.1 today. mujoco 3.4.0 changed box-box collision handling, which desyncs `libero_spatial` task 5's saved init state: the bowl no longer settles onto the ramekin during `reset()`, and task-5 success collapses from 98% at the 3.3.x models were trained under to 12% at 3.8.1 (see #1460's evidence table). This makes the published GRPO LIBERO-spatial numbers unreproducible on a fresh install, through no fault of the checkpoint.

A prior v0.x install path had this exact pin (`mujoco<=3.9.0`, added in #1302), until #1410 dropped it while migrating `install_libero_env` off the git-clone-based LIBERO install onto the `rlinf-libero` PyPI package, with no replacement.

Fixes #1460

### How has this been tested?

- Confirmed on current HEAD, via `uv pip install rlinf-libero` in a clean container, that mujoco resolves to 3.8.1 unpinned; with `uv pip install --no-deps mujoco==3.3.0` added, it resolves to 3.3.0 as intended.
- Read #1410's diff directly to confirm the prior pin was dropped as refactor fallout, not a deliberate revert.
- Ran `.cursor/skills/install-check/check.sh`: same 33 pre-existing candidates before and after, no new hits from this change.
- Not independently re-run: the physics regression itself (no LIBERO/mujoco simulation available in this environment). That part relies on #1460's own settle-probe evidence (raw JSON output at two mujoco versions, same 50 init states) and the maintainer's related advice in #1375 to downgrade to mujoco 3.3.2 for a separate visual regression.

### Types of changes

### Checklist:

### DIFF
--- a/requirements/install.sh
+++ b/requirements/install.sh
@@ -2373,6 +2373,9 @@ retry_cmd() {
 
 install_libero_env() {
     uv pip install rlinf-libero
+    # rlinf-libero caps mujoco only from above (<3.9.0); >=3.4.0 desyncs
+    # LIBERO's saved init states, so pin the era-faithful version explicitly.
+    uv pip install --no-deps mujoco==3.3.0
     materialize_package_files rlinf-libero
     retry_cmd libero-download-assets --skip-existing
     reset_libero_config
@@ -2450,6 +2453,7 @@ install_d4rl_env() {
 
 install_liberopro_env() {
     uv pip install rlinf-libero rlinf-liberopro
+    uv pip install --no-deps mujoco==3.3.0
     materialize_package_files rlinf-libero
     materialize_package_files rlinf-liberopro
     retry_cmd libero-download-assets --skip-existing
@@ -2459,6 +2463,7 @@ install_liberopro_env() {
 
 install_liberoplus_env() {
     uv pip install rlinf-libero "rlinf-liberoplus>=0.1.3"
+    uv pip install --no-deps mujoco==3.3.0
     materialize_package_files rlinf-libero
     materialize_package_files rlinf-liberoplus
     retry_cmd libero-download-assets --skip-existing


andylin-hao (approved):

zoeyuchao (approved):

andylin-hao:
Thanks for investigating this. The compatibility fix is now shipped in the RLinf-maintained LIBERO fork packages, so an installer-level override in RLinf is no longer needed:

- [RLinf/LIBERO@abdd25c](https://github.com/RLinf/LIBERO/commit/abdd25c), released as [`rlinf-libero==0.1.3`](https://pypi.org/project/rlinf-libero/0.1.3/)
- [RLinf/LIBERO-PRO@23f0f88](https://github.com/RLinf/LIBERO-PRO/commit/23f0f88), released as [`rlinf-liberopro==0.1.3`](https://pypi.org/project/rlinf-liberopro/0.1.3/)
- [RLinf/LIBERO-plus@1c52dfc](https://github.com/RLinf/LIBERO-plus/commit/1c52dfc), released as [`rlinf-liberoplus==0.1.5`](https://pypi.org/project/rlinf-liberoplus/0.1.5/)

The root cause is the interaction between LIBERO's saved initial state and MuJoCo's physics version. The bowl is serialized with its center about 11 cm above the ramekin, and `reset()` relies on MuJoCo to drop and settle it before the policy acts. MuJoCo 3.4 corrected box-box contact-distance handling, which changes that settling trajectory and can leave the bowl laterally displaced. This changes the policy's initial observation and sharply reduces task success.

We initially tried replacing the affected serialized init states with already-settled states, but that modification caused E2E CI failures. Those binary state changes have now been reverted in all three forks. The final fix preserves the original benchmark states and constrains each package to `mujoco>=3.0.0,<3.4.0`, keeping execution on the compatible pre-3.4 physics behavior.

The new wheel and source releases have been published, and their PyPI metadata confirms the `mujoco<3.4.0` constraint. Closing this PR remains appropriate because the compatibility requirement is now enforced by the fork packages themselves for RLinf and direct package users.

PR #1498, opened 2026-08-23
feat: support StreamingVLA SFT

GhahahahaG:
### Description

Add isolated StreamingVLA Streaming Flow Policy (SFP) supervised fine-tuning support to RLinf.

This PR introduces:

- A private `rlinf.models.embodiment.streamingvla` implementation.
- StreamingVLA-specific SFP objectives, data transforms, normalization, and strict checkpoint loading.
- Episode-level `action_states` support.
- LIBERO-to-LeRobot conversion and StreamingVLA normalization-statistics tools.
- An independent SFT configuration and installation option.
- English and Chinese training documentation.

The existing OpenPI and OpenPI-RLinf model implementations are not modified. Inference, rollout, streaming action generation, LIBERO evaluation, and reinforcement learning are out of scope.

### Motivation and Context

StreamingVLA uses modified OpenPI training semantics that are not equivalent to RLinf's existing OpenPI implementations. The implementation therefore lives in a private StreamingVLA namespace so existing models retain their current behavior.

#### Numerical training validation

The PR implementation was compared with the reference `pi05_libero_sfp` PyTorch training loop using `pi05_libero_pytorch` initialization, real `action_states`, SFP, action horizon 10, action dimension 32, seed 42, global/micro batch 16/4, two A800 GPUs, and 100 optimizer steps (Reference DDP versus RLinf FSDP2).

Results:

- Initial loss relative error: 0.1416%
- Median / maximum 100-step loss relative error: 0.3824% / 1.4836%
- Maximum 10-step moving-average error: 0.4549%
- `action_out_proj` update-norm relative error: 0.2560%
- All losses, gradients, and parameter updates were finite
- The step-100 RLinf checkpoint was saved successfully

The follow-up review changes only restore unrelated OpenPI installer text, update documentation commands, reuse the existing norm-stat loader helpers, and correct a comment; they do not change the training forward/backward path.

#### Regression and documentation validation

- Existing model/LeRobot/metric/worker safe regressions: 29 passed
- Ruff lint and format checks passed
- Full pre-commit and DCO checks passed
- Python compile and StreamingVLA norm-tool import/copy checks passed
- English and Chinese 209-page Sphinx builds passed with warnings treated as errors (apart from the pre-existing environment-only `autodoc.import_object` warning suppression)
- Existing `openpi/`, `openpi_rlinf/`, common FSDP, metrics, and `tests/unit_tests/` paths have zero diff from upstream

PR-specific test files were removed from the final diff at maintainer request. GPU/checkpoint numerical validation remains documented above and uses external assets rather than downloading large files in CI.

### Additional information

A previous full training run completed 100,000 optimizer steps without NaN/Inf and produced complete checkpoints every 5,000 steps:

https://wandb.ai/ghhhhhh-tsinghua-university/rlinf/runs/zrxhf0c0

### Types of changes

### Checklist:

- [ ] I have added tests to cover my changes. (PR-specific tests were removed at maintainer request.)
- [x] All new and relevant existing tests passed.


rlinf/models/embodiment/streamingvla/openpi_compat/preprocessing.py
```diff
@@ -0,0 +1,205 @@
...
+                if max_h > 0 and max_w > 0:
+                    # Use tensor operations instead of .item() for torch.compile compatibility
+                    start_h = torch.randint(0, max_h + 1, (1,), device=image.device)
+                    start_w = torch.randint(0, max_w + 1, (1,), device=image.device)
+                    image = image[
+                        :,
+                        start_h : start_h + crop_height,
+                        start_w : start_w + crop_width,
+                        :,
+                    ]
```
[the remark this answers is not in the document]
GhahahahaG: Thanks for flagging this. I verified the exact expression with real PyTorch: a one-element integer tensor is accepted through Tensor.__index__ as a slice boundary, and the crop completed with the expected shape. The upstream StreamingVLA training reference uses the same shape-(1) torch.randint slicing. I am preserving that reference-equivalent behavior because it does not raise TypeError.

rlinf/models/embodiment/streamingvla/openpi_compat/preprocessing.py
```diff
@@ -0,0 +1,205 @@
...
+                    align_corners=False,
+                ).permute(0, 2, 3, 1)  # [b, c, h, w] -> [b, h, w, c]
+
+                # Random rotation (small angles)
+                # Use tensor operations instead of .item() for torch.compile compatibility
+                # Random angle between -5 and 5 degrees.
+                angle = torch.rand(1, device=image.device) * 10 - 5
+                if torch.abs(angle) > 0.1:  # Only rotate if angle is significant
+                    # Convert to radians
+                    angle_rad = angle * torch.pi / 180.0
```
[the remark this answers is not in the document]
GhahahahaG: Thanks. A shape-(1) tensor contains exactly one element, so Python bool conversion is defined; the ambiguous-bool RuntimeError applies when the tensor has more than one element. I ran the exact condition with real PyTorch and it completed successfully. The upstream StreamingVLA training reference also uses this exact shape-(1) condition, so I am keeping the numerically equivalent reference implementation.

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -26,11 +26,72 @@
...
+            SupportedModel(cfg.actor.model.model_type) == SupportedModel.STREAMINGVLA
+        )
+        self._streamingvla_step_inputs: Any | None = None
+        if self._is_streamingvla:
+            from rlinf.models.embodiment.streamingvla.training import (
+                seed_streamingvla_training,
+            )
+
+            rank = int(os.environ.get("LOCAL_RANK", os.environ.get("RANK", "0")))
+            seed_streamingvla_training(cfg.actor.seed, rank)
```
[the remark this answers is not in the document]
GhahahahaG: Fixed in 384cfa8. StreamingVLA RNG seeding now prefers the global RANK and falls back to LOCAL_RANK only when RANK is unavailable. I also added tests for both the multi-node-style RANK/LOCAL_RANK case and the fallback case.

toolkits/lerobot/calculate_streamingvla_norm_stats.py
```diff
@@ -0,0 +1,202 @@
...
+class RemoveStrings(transforms.DataTransformFn):
+    """Remove non-numeric fields that are irrelevant to norm statistics."""
+
+    def __call__(self, data: dict[str, Any]) -> dict[str, Any]:
+        """Return only fields that can be converted to numeric arrays."""
+        return {
+            key: value
+            for key, value in data.items()
+            if not np.issubdtype(np.asarray(value).dtype, np.str_)
+        }
```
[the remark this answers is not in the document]
GhahahahaG: Fixed the contract description in 384cfa8. The implementation intentionally matches RLinf existing toolkits/lerobot/calculate_norm_stats.py behavior: it removes string fields rather than all object-dtype fields. I updated the docstring accordingly and added a regression test confirming that strings are removed while compatible numeric/object fields are preserved.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
```diff
@@ -0,0 +1,133 @@
...
+
+本文档介绍 RLinf 中面向 LIBERO 示教数据的 **仅训练** StreamingVLA
+集成。模型通过 ``model_type: streamingvla`` 注册，在 Pi0.5 主干上实现
+Streaming Flow Policy（SFP）训练目标，并支持通过 FSDP2 进行全参数
+SFT。本轮集成不包含动作生成、rollout、仿真评测或强化学习。
+
+StreamingVLA 的代码隔离在
+``rlinf/models/embodiment/streamingvla`` 中，不修改 RLinf 现有的
+``openpi`` / ``openpi_rlinf`` 模型、全局 Transformers 包、通用归一化或
+优化器实现。
```
Lin-xs: Please polish this document and drop wording that reads like a PR description (e.g. “It does not implement action generation, rollout, simulator evaluation, or reinforcement learning.”).

Please also refine the Chinese version: replace stiff/literal translations (e.g. “demonstrations” → “示教数据”).

docs/source-en/rst_source/examples/sft_index.rst
```diff
@@ -21,6 +21,17 @@ After running SFT here, continue to :doc:`vla_wam_index` (model-centric RL) or :
...
+       <a href="embodied/sft_streamingvla.html" style="display: block;"><img src="https://raw.githubusercontent.com/RLinf/misc/main/pic/pi0_icon.jpg"
+            style="width: 100%; height: 200px; object-fit: cover; border-radius: 8px; box-shadow: 0 2px 6px rgba(0,0,0,0.15);" /></a>
+       <p style="margin-top: 8px; font-size: 14px; line-height: 1.4;">
+         <a href="embodied/sft_streamingvla.html" style="text-decoration: underline; color: blue;">
+           <b>StreamingVLA Supervised Fine-Tuning</b>
+         </a><br>
+         Run isolated SFP training for StreamingVLA on LIBERO demonstrations
+       </p>
+     </div>
+
```
Lin-xs, line 34: Please move this section after `OpenPI Mixed-Precision SFT` so the order matches the index page. Also, please use a matching image here instead of the pi0 one.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
```diff
@@ -0,0 +1,133 @@
...
+--------
+
+在仓库根目录中创建 StreamingVLA + LIBERO 环境：
+
+.. code:: bash
+
+   bash requirements/install.sh embodied --model streamingvla --env libero
+   source .venv/bin/activate
+
+与 ``--model openpi`` 的安装路径不同，StreamingVLA 不会将替换文件复制到全局
```
Lin-xs: Delete this paragraph.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
```diff
@@ -0,0 +1,133 @@
...
+Transformers 目录；兼容的 Gemma、SigLIP、图像预处理和 SFP 逻辑都从
+StreamingVLA 私有命名空间导入。
+
+
+准备数据和权重
+--------------
+
+先将公开的 LIBERO RLDS 数据转换为 LeRobot 格式；转换器会直接写入每个时刻
+动作执行前、从零起点累计得到的 ``action_states``：
```
Lin-xs: Please provide a public LIBERO RLDS suite to run this example.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
```diff
@@ -0,0 +1,133 @@
...
+
+StreamingVLA 的局部 transform 将 LIBERO 的 7 维 state、action state 和动作
+delta 填充到 32 维模型空间，并使用以零为中心的线性分位数归一化：
+
+.. math::
+
+   x / (\max(|q_{0.01}|, |q_{0.99}|) + 10^{-6}).
+
+将 ``norm_stats.json`` 放在
+``<assets_dir>/<asset_id>/norm_stats.json``。统计量必须包含数据集使用的
```
Lin-xs: Where is `<assets_dir>/<asset_id>`?

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
```diff
@@ -0,0 +1,133 @@
...
+``state``、``actions`` 和 ``action_states`` 项。
+可通过 StreamingVLA 专用工具生成：
+
+.. code:: bash
+
+   python toolkits/lerobot/calculate_streamingvla_norm_stats.py \
+       --repo-id /path/to/libero_lerobot_dataset \
+       --assets-dir /path/to/dataset-parent \
+       --asset-id libero_lerobot_dataset
```
Lin-xs: Please explain what `repo_id`, `assets_dir`, and `asset_id` are. An example repo / asset dir / id would help.

tests/unit_tests/test_streamingvla_config.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2026 The RLinf Authors.
```
Lin-xs: Please delete all unit tests.

requirements/install.sh
```diff
@@ -1976,6 +1980,18 @@ EOF
     uv pip uninstall pynvml || true
 }
 
+install_streamingvla_model() {
+    case "$ENV_NAME" in
+        libero)
+            install_openpi_model
```
Lin-xs: Please copy the corresponding scripts from `install_openpi_model` into this path. Otherwise it’s odd to still branch on `"$MODEL" = "openpi"` inside `install_openpi_model`.

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -24,13 +24,78 @@
...
+            self._configure_streamingvla_grad_norm_group()
+            self._streamingvla_step_inputs = StreamingVLAStepInputBuffer(
+                seed=cfg.actor.seed,
+                rank=self._rank,
+                local_batch_size=self.global_batch_size // self._world_size,
+                action_dim=int(cfg.actor.model.streamingvla.model_action_dim),
+                device=self.device,
+            )
+
+    def _configure_streamingvla_grad_norm_group(
```
Lin-xs: Delete this method if it is unnecessary.

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -24,13 +24,78 @@
...
+        if model_type == SupportedModel.STREAMINGVLA:
+            from rlinf.models.embodiment.streamingvla.data import (
+                build_streamingvla_dataloader,
+            )
+
+            if eval_dataset:
+                raise NotImplementedError(
+                    "StreamingVLA validation is not implemented in the SFT-only integration."
+                )
+            return build_streamingvla_dataloader(self.cfg, self._world_size)
```
Lin-xs: Please keep this builder’s arguments consistent with the other dataloader builders, and move the `eval_dataset` check into this method.

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -24,13 +24,78 @@
...
+        self._is_streamingvla = (
+            SupportedModel(cfg.actor.model.model_type) == SupportedModel.STREAMINGVLA
+        )
+        self._streamingvla_step_inputs: Any | None = None
+        if self._is_streamingvla:
+            from rlinf.models.embodiment.streamingvla.training import (
+                seed_streamingvla_training,
+            )
+
+            seed_streamingvla_training(cfg.actor.seed, _get_streamingvla_seed_rank())
```
Lin-xs: Replace `_get_streamingvla_seed_rank` with `self._rank`

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -24,13 +24,78 @@
...
+            process_group = torch.distributed.group.WORLD
+        self._strategy._dp_group = process_group
+
+    def set_global_step(self, global_step: int) -> None:
+        """Set the step and prepare partition-invariant StreamingVLA RNG inputs."""
+        super().set_global_step(global_step)
+        if not self._is_streamingvla:
+            return
+
+        if self._streamingvla_step_inputs is None:
```
Lin-xs: Delete all these checks because `self._streamingvla_step_inputs` has been initialized at first.

rlinf/models/embodiment/streamingvla/svla_pytorch.py
```diff
@@ -0,0 +1,252 @@
...
+from .openpi_compat.preprocessing import preprocess_observation_pytorch
+from .sfp import (
+    compute_sfp_flow_targets,
+    create_sinusoidal_pos_embedding,
+    make_attention_masks_and_position_ids,
+    sample_beta,
+)
+
+
+class SVLAPytorch(nn.Module):
```
Lin-xs: Please keep the name consistent with other classes / files (i.e., `StreamingVLAPytorch`, `streamingvla_pytorch.py`).

rlinf/models/embodiment/streamingvla/__init__.py
```diff
@@ -0,0 +1,169 @@
...
+        str(streaming_cfg.config_name),
+        model_path=str(cfg.model_path),
+        data_kwargs=streaming_cfg.get("data", None),
+    )
+    model_config = train_config.model
+    precision = str(cfg.get("precision", "bf16")).lower()
+    dtype = (
+        "float32" if precision in {"fp32", "float32", "32", "32-true"} else "bfloat16"
+    )
+    object.__setattr__(model_config, "dtype", dtype)
```
Lin-xs: Please use `model_config.dtype = dtype`.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
```diff
@@ -0,0 +1,133 @@
...
+启动训练
+--------
+
+在仓库根目录使用 RLinf 现有 VLA SFT 启动脚本：
+
+.. code:: bash
+
+   bash examples/sft/run_vla_sft.sh libero_sft_streamingvla
+
+默认 logger 是 TensorBoard。如需显式上传 Weights & Biases，先使用常规 W&B
```
Lin-xs: Delete.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
```diff
@@ -0,0 +1,133 @@
...
+``actor.model.model_path`` 应指向含有 ``model.safetensors`` 的 Pi0.5
+PyTorch 基础权重目录。OpenPI 旧格式基础权重和 RLinf StreamingVLA checkpoint
+都由私有严格加载器处理；任何 missing key、unexpected key 或 shape mismatch
+都会立即终止初始化。
+
+
+配置训练
+--------
+
+提交的配置文件只包含可移植占位符：
```
Lin-xs: Too specific. Delete.

docs/source-en/rst_source/examples/embodied/sft_streamingvla.rst
```diff
@@ -0,0 +1,143 @@
+StreamingVLA Supervised Fine-Tuning
```
Lin-xs, line 1: Please refine the English and Chinese documentation:

- Provide detailed, reproducible steps rather than walking through code details.
- Drop statements that sound like a PR description.
- Re-read the Chinese doc and replace leftover machine-translation wording with natural Chinese.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
GhahahahaG: Addressed in c5b57ab6. I rewrote the English and Chinese pages as a reproducible setup and training workflow, removed PR-description wording, and revised the Chinese terminology to natural phrases such as 示教数据 and 动作状态.

docs/source-en/rst_source/examples/sft_index.rst
GhahahahaG: Addressed in c5b57ab6. The StreamingVLA card now follows OpenPI Mixed-Precision SFT in both languages and uses the StreamingVLA project image instead of the Pi0 image.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
GhahahahaG: Removed in c5b57ab6 as requested.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
GhahahahaG: Addressed in c5b57ab6. The guide now downloads the public openvla/modified_libero_rlds dataset referenced by the StreamingVLA project and includes the conversion command.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
GhahahahaG: Addressed in c5b57ab6. The example now uses assets_dir=/data/streamingvla/assets and asset_id=libero_streamingvla, so the generated file is explicitly /data/streamingvla/assets/libero_streamingvla/norm_stats.json.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
GhahahahaG: Addressed in c5b57ab6. The guide explains repo_id as the LeRobot dataset identifier, assets_dir as the normalization asset root, and asset_id as its child directory. It also provides concrete values for all three.

tests/unit_tests/test_streamingvla_config.py
GhahahahaG: Addressed in 0115e3bb. All six StreamingVLA test files introduced by this PR were removed and tests/unit_tests/pytest.ini was restored to the upstream version. No existing RLinf tests were removed.

requirements/install.sh
GhahahahaG: Addressed in 0115e3bb. install_streamingvla_model now contains its own LIBERO/OpenPI dependencies, version pins, and asset installation steps. The StreamingVLA path no longer calls install_openpi_model, and the OpenPI Transformers overlay is restored to the OpenPI-only path.

rlinf/workers/sft/fsdp_vla_sft_worker.py
GhahahahaG: Addressed in 0115e3bb. The helper method was removed. The minimal StreamingVLA-only fallback remains inline: when FSDP2 has no data-parallel group, it uses the fsdp mesh group so gradient norms cover all parameter shards. Removing that behavior changes clipping and updates. The post-change two-GPU 100-step run passed every numerical gate, and its loss, grad norm, LR, and parameter-update summaries are identical to the pre-review clean-PR run.

rlinf/workers/sft/fsdp_vla_sft_worker.py
GhahahahaG: Addressed in 0115e3bb. The StreamingVLA builder now follows the common cfg, world_size, rank, data_paths, eval_dataset signature and owns the eval_dataset rejection check.

rlinf/workers/sft/fsdp_vla_sft_worker.py
GhahahahaG: Addressed in 0115e3bb. The helper was removed and StreamingVLA seeding now directly uses self._rank, which is initialized by Worker.__new__.

rlinf/workers/sft/fsdp_vla_sft_worker.py
GhahahahaG: Addressed in 0115e3bb. The redundant None checks were removed; the StreamingVLA step-input buffer is used directly after its guaranteed initialization.

rlinf/models/embodiment/streamingvla/svla_pytorch.py
GhahahahaG: Addressed in 0115e3bb. The file and class are now streamingvla_pytorch.py and StreamingVLAPytorch, and all internal imports were updated.

rlinf/models/embodiment/streamingvla/__init__.py
GhahahahaG: Pi0Config is declared with dataclasses.dataclass(frozen=True) in the pinned OpenPI package, so direct model_config.dtype assignment raises FrozenInstanceError. In 0115e3bb I use dataclasses.replace to set dtype, action_horizon, and action_dim in one immutable update. The use_sfp compatibility field is added separately because this pinned Pi0Config predates that field. Please confirm this frozen-dataclass-compatible approach.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
GhahahahaG: Removed in c5b57ab6 as requested.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
GhahahahaG: Removed in c5b57ab6 as requested.

docs/source-en/rst_source/examples/embodied/sft_streamingvla.rst
GhahahahaG: Addressed in c5b57ab6. Both pages now present a reproducible sequence covering installation, public dataset download, LeRobot conversion, norm statistics, Pi0.5 checkpoint conversion, configuration, and launch. Code-internal and PR-description wording was removed, and the Chinese page was rewritten with natural terminology.

GhahahahaG:
@Lin-xs Thank you for the detailed review. The requested changes have been addressed in `0115e3b` and `c5b57ab`, and the branch has been rebased onto the current `upstream/main` used by this update.

Validation summary:

- Full pre-commit passed: Ruff, Ruff format, and DCO signoff.
- Existing safe VLA/OpenPI regression suite: 30 passed, 4 deselected (the deselected cases create shared GPU worker groups).
- English and Chinese Read the Docs checks both passed.
- Two-GPU FSDP2 smoke passed from `pi05_libero_pytorch` with a complete step-1 checkpoint.
- The 100-step Reference/RLinf comparison passed all gates: initial loss relative error 0.142%, median/max per-step loss error 0.382%/1.484%, maximum 10-step moving-average error 0.455%, and `action_out_proj` update-norm error 0.256%. No NaN, Inf, OOM, missing/unexpected keys, or shape mismatches occurred.
- The existing OpenPI/OpenPI-RLinf model directories, common FSDP/optimizer/scheduler/normalization behavior, and `tests/unit_tests` remain unchanged relative to `upstream/main`.

I replied to all review comments and resolved the implemented threads. One Pi0Config thread remains open intentionally: the pinned OpenPI `Pi0Config` is a frozen dataclass, so direct `model_config.dtype = dtype` raises `FrozenInstanceError`; the implementation uses `dataclasses.replace` for `dtype`, horizon, and action dimension. Please take another look when convenient.

rlinf/models/embodiment/streamingvla/__init__.py
GhahahahaG: Thanks for following up on this. I re-checked the exact OpenPI package pinned by RLinf (`rlinf-openpi==0.1.1`). Its `openpi/models/pi0_config.py` declares the configuration as:

```python
@dataclasses.dataclass(frozen=True)
class Pi0Config(_model.BaseModelConfig):
    dtype: str = "bfloat16"
```

Because the dataclass is frozen, `model_config.dtype = dtype` raises `dataclasses.FrozenInstanceError: cannot assign to field 'dtype'` before model construction. This is why the current implementation uses `dataclasses.replace(...)` to create one new immutable config with the requested `dtype`, `action_horizon`, and `action_dim`. It preserves the same value change without mutating the pinned OpenPI object.

The pinned `Pi0Config` also predates the StreamingVLA-only `use_sfp` field. We therefore attach that missing compatibility field only to the newly created private StreamingVLA config; no existing OpenPI/OpenPI-RLinf config or global class is modified.

Could you please confirm that the frozen-dataclass-compatible `dataclasses.replace` approach is acceptable here? If you prefer another immutable update pattern, I am happy to adjust it, but direct attribute assignment is not executable with the dependency version installed by this PR.

requirements/install.sh
```diff
@@ -1957,13 +1957,13 @@ install_openpi_model() {
...
 
-    # Replace transformers models with OpenPI's modified versions
+    # The legacy OpenPI model needs its source-tree Transformers overlay.
     local py_major_minor
     py_major_minor=$(python - <<'EOF'
 import sys
 print(f"{sys.version_info.major}.{sys.version_info.minor}")
 EOF
-)
+    )
```
Lin-xs: These two edits look unnecessary, please revert them.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
```diff
@@ -0,0 +1,129 @@
...
+
+StreamingVLA 从官方 ``pi05_libero`` 权重开始微调。RLinf 需要一个包含
+``model.safetensors`` 的 PyTorch checkpoint 目录。请在 `OpenPI
+<https://github.com/Physical-Intelligence/openpi>`_ 仓库中下载并转换官方权重：
+
+.. code:: bash
+
+   python -c "from openpi.shared import download; download.maybe_download('gs://openpi-assets/checkpoints/pi05_libero')"
+
+   uv run examples/convert_jax_model_to_pytorch.py \
```
Lin-xs: `uv run` -> `python`.

toolkits/lerobot/calculate_streamingvla_norm_stats.py
```diff
@@ -0,0 +1,202 @@
+# Copyright 2026 The RLinf Authors.
```
Lin-xs, line 1: Could you import the shared loader helpers from the existing script (`calculate_norm_stats.py`) or just reuse it?

rlinf/workers/sft/fsdp_vla_sft_worker.py
```diff
@@ -26,11 +26,57 @@
...
         super().__init__(cfg)
+        if self._is_streamingvla:
+            from rlinf.models.embodiment.streamingvla.training import (
+                StreamingVLAStepInputBuffer,
+            )
+
+            # FSDP2 stores only local gradient shards. Its norm helper skips the
+            # cross-rank reduction when no data-parallel group is configured.
+            if self._strategy._dp_group is None:
+                self._strategy._dp_group = self._device_mesh["fsdp"].get_group()
```
Lin-xs, line 48: Why should we set `_strategy._dp_group` here?

Could you please help to check it? @guozhen1997

requirements/install.sh
GhahahahaG: Reverted in a08f2735. The unrelated comment wording and heredoc-closing indentation in install_openpi_model now match upstream again; the independent StreamingVLA installer remains separate. bash -n and full pre-commit both pass.

docs/source-zh/rst_source/examples/embodied/sft_streamingvla.rst
GhahahahaG: Updated in a08f2735: the checkpoint conversion command now uses python in both the English and Chinese guides. Both 209-page Sphinx builds pass with warnings treated as errors (with only the pre-existing environment-only autodoc warning suppressed).

toolkits/lerobot/calculate_streamingvla_norm_stats.py
GhahahahaG: Updated in a08f2735. calculate_streamingvla_norm_stats.py now imports create_torch_dataloader and create_rlds_dataloader from the existing calculate_norm_stats.py and removes the duplicate loader/RemoveStrings implementation. It retains only the StreamingVLA config/output behavior, the complete-batch guard, and copying the full actions statistics object to action_states. Ruff, import assertions, CLI import, and pre-commit pass.

rlinf/workers/sft/fsdp_vla_sft_worker.py
GhahahahaG: This is needed for correct FSDP2 gradient clipping, not only metric logging. RLinf currently builds a 1-D device mesh named fsdp, while FSDPManager initializes _dp_group only when a ddp mesh dimension exists. Therefore _dp_group is otherwise None, and get_grad_norm skips the cross-rank all-reduce of the local parameter-shard norms. In a two-GPU measurement, omitting this fallback produced grad_norm=50.6862 versus Reference DDP=71.9728 (approximately 1/sqrt(2)); using self._device_mesh["fsdp"].get_group() produced 72.0690 (0.134% relative error), and the 100-step loss/update gates then passed. The assignment is guarded by model_type == streamingvla and only fills a missing group, so existing models are unchanged. I have left this thread unresolved for maintainer guidance: should we retain this scoped fallback here, or address the general 1-D FSDP2 group initialization in a separate PR?

GhahahahaG:
Follow-up review changes are now available in `a08f2735` (rebased onto current `main`):

- reverted the two unrelated edits inside `install_openpi_model`;
- changed the checkpoint conversion command from `uv run` to `python` in both guides;
- reused the existing `calculate_norm_stats.py` loader helpers and removed the duplicate implementation;
- corrected the stale immutable-config comment;
- updated the PR description to reflect the maintainer-requested removal of PR-specific tests and the completed Sphinx validation.

Validation after the rebase/change:

- full pre-commit (Ruff, format, DCO): passed;
- existing safe regressions: 29 passed;
- norm-tool shared-loader/import/action-stat-copy assertions: passed;
- English and Chinese 209-page local Sphinx builds: passed;
- both Read the Docs checks on `a08f2735`: passed;
- existing OpenPI/OpenPI-RLinf, common FSDP/metrics, and `tests/unit_tests` protected paths: zero diff.

I replied with runtime evidence in the remaining `_dp_group` thread and left it unresolved for maintainer guidance. The frozen `Pi0Config` thread also remains unresolved because direct assignment raises `FrozenInstanceError` in the pinned `rlinf-openpi==0.1.1`; the current `dataclasses.replace` implementation is executable and isolated.

@Lin-xs @guozhen1997 Could you please review these two remaining design points and the updated PR when convenient?

Lin-xs (approved):

requirements/install.sh
```diff
@@ -1977,6 +1977,27 @@ EOF
     uv pip uninstall pynvml || true
 }
 
+install_streamingvla_model() {
```
andylin-hao, line 1984: I can see that the only difference of the streamingvla and openpi is whether to replace the transformers with openpi's patch. Why can't the streamingvla work with the patch?

GhahahahaG: Thanks for pointing this out. The original reason for isolating the StreamingVLA path was its different training/data semantics, especially its StreamingVLA-local action/state normalization and `action_states` handling. We wanted those changes to remain private instead of modifying OpenPI global transforms.

However, that normalization difference is independent of the Transformers overlay and does not make the OpenPI patch incompatible. I verified the environment used for our successful two-GPU smoke and 100-step Reference comparison: all five installed Gemma/PaliGemma/SigLIP files exactly match `openpi/models_pytorch/transformers_replace`, so StreamingVLA already trains correctly with the patch.

Therefore you are right that the installation logic can be reused. We will keep the modified normalization/data logic in the private StreamingVLA namespace, retain the `libero`-only entry check, and have `install_streamingvla_model` delegate to `install_openpi_model` instead of duplicating the dependency/assets installation while omitting the overlay.

GhahahahaG: A clarification to my previous reply after a direct source-to-source comparison: there are two separate normalization changes in the StreamingVLA reference.

1. `openpi/transforms.py` changes dataset quantile normalization to `x / (max(abs(q01), abs(q99)) + 1e-6)`. This change is required for SFP training. The PR implements it locally as `StreamingVLANormalize` / `StreamingVLAUnnormalize`, so it does not modify OpenPI global transforms.

2. The reference also changes `transformers_replace/models/gemma/modeling_gemma.py` inside adaptive RMSNorm. Compared with the pinned `rlinf-openpi==0.1.1` patch, that delta reshapes/broadcasts `adarms_cond` for per-token conditions. The source comment identifies training-time RTC as the reason. That path is not used by this PR: SFP supplies one shared condition per sample (`[B, width]`) for its single action token, and `StreamingVLAPytorch` uses `rlinf.models.embodiment.openpi_rlinf.pi0_model.gemma`, not `transformers.models.gemma`. The existing RLinf RMSNorm handles this shared-condition shape.

So the conclusion for the current training-only SFP scope is still that the standard OpenPI Transformers overlay can be installed without changing StreamingVLA numerical behavior. Our successful 100-step target run used an environment containing that standard overlay, while the required dataset normalization remained in the private StreamingVLA transform. The reference Gemma per-token RTC extension would need separate treatment only if RTC/non-SFP inference or training were added later; those are explicitly out of scope here.

Thank you for prompting this distinction. I will keep the private SFP normalization and simplify the installer to reuse the OpenPI installation path.

GhahahahaG: Implemented in `0b2f93c5`. `install_streamingvla_model` now only enforces the current `libero`-only scope and delegates to `install_openpi_model`, so it reuses the standard OpenPI dependency pins, assets, tokenizers bound, and Transformers overlay instead of duplicating the installation path.

The StreamingVLA-specific dataset quantile normalization remains isolated in `StreamingVLANormalize` / `StreamingVLAUnnormalize`; no OpenPI global transform is changed. The Reference-only per-token adaptive-RMSNorm extension is for RTC/non-SFP behavior and is not used by this training-only SFP implementation.

Validation after rebasing onto current `main`: Bash syntax and delegation/error-path harness passed; compile/import passed; the equivalent 29 existing model/data/metric/worker regressions passed; full pre-commit (Ruff, format, DCO) passed; protected OpenPI/OpenPI-RLinf/common FSDP/test paths remain zero diff.

GhahahahaG:
The installation follow-up is now implemented in commit `0b2f93c5`: `install_streamingvla_model` retains the LIBERO-only guard and delegates to `install_openpi_model`, so StreamingVLA now reuses the standard OpenPI dependency pins, assets/tokenizer setup, and Transformers overlay instead of duplicating them. The private StreamingVLA dataset normalization and `action_states` handling remain unchanged.

Validation on the rebased branch: install-script syntax and delegation harness passed; full pre-commit and import smoke passed; 29 equivalent existing registration/data/metric/worker regressions passed; protected OpenPI/OpenPI-RLinf/distributed/test directories remain unchanged relative to the new base. Both regenerated Read the Docs checks passed.

The installation review thread has been replied to and resolved. @andylin-hao @guozhen1997 @Lin-xs, could you please re-review the current head when convenient?

andylin-hao:
@GhahahahaG Nice work here. Your source comparison in the installer thread was useful, and it's what got me digging into this properly.

My main concern is that this adds a third top-level `SupportedModel` for what is really just π₀.₅. You already showed that `StreamingVLAPytorch` imports `openpi_rlinf`'s `gemma` and `siglip` rather than `transformers`. I went one step further and checked the parameters: they're identical to what `Pi0` builds. I copied a state dict across with `strict=True` and all 415 tensors matched, nothing left over on either side.

The cost of making it a separate model type is that none of `openpi_rlinf`'s machinery applies, so you end up rebuilding a lot of it: the preprocessing module, a second `_CONFIGS` registry, `StreamingVLAObservation`, copies of `posemb_sincos` and `make_attn_mask`, a separate installer. And rollout, eval and RL all have to raise `NotImplementedError` even though `task: eval` and `task: rl` already exist.

I think it fits better as an objective flag, the same way `use_rlt` works. `sft_forward` already switches between flow matching and the RLT objective based on `actor.model.openpi`, so `use_sfp` can sit right next to it, `compute_sfp_loss` next to `compute_loss`, and `pi05_libero_sfp` as one more entry in the openpi config registry.

Based on this, I put a version together: GhahahahaG/RLinf#1, opened against this branch. Your objective, `action_states` and the scale-only normalization are untouched. I checked that both implementations give bitwise identical losses on the same weights and inputs. It comes to 44 files, +1632/−2267.

Mind taking a look and merging it if it looks right to you? Happy to change anything. And if you'd rather keep it standalone, just say so and I'll write up both options for the maintainers to pick.

GhahahahaG:
@andylin-hao Thank you for putting together [the refactor](https://github.com/GhahahahaG/RLinf/pull/1) and for the careful parameter and fixed-input comparisons. I agree with your architectural direction: SFP fits better as an opt-in `use_sfp` objective of `openpi_rlinf` than as a third top-level `SupportedModel`. I am happy to adopt and merge your refactor into this PR.

I originally kept StreamingVLA isolated to avoid changing existing OpenPI/OpenPI-RLinf behavior. The opt-in design addresses that concern while removing substantial duplication, provided the non-SFP paths remain unchanged. Before calling the refactor training-equivalent, I will review the integration and rerun the real-GPU 100-step comparison from `pi05_libero_pytorch` with the same data/order, time/noise, and distributed settings. In particular, your description notes that the `train=True` image augmentation path differs, so the fixed-input `train=False` loss match alone does not establish matching training trajectories. I will also verify that SFP retains the correct cross-shard gradient norm and clipping while the general fix in RLinf/RLinf#1558 is still separate.

Thanks again for contributing this. The proposed structure is a better fit for RLinf, and I am comfortable moving forward with it once those training and existing-model regression checks pass.

andylin-hao:
> @andylin-hao Thank you for putting together [the refactor](https://github.com/GhahahahaG/RLinf/pull/1) and for the careful parameter and fixed-input comparisons. I agree with your architectural direction: SFP fits better as an opt-in `use_sfp` objective of `openpi_rlinf` than as a third top-level `SupportedModel`. I am happy to adopt and merge your refactor into this PR.
> 
> I originally kept StreamingVLA isolated to avoid changing existing OpenPI/OpenPI-RLinf behavior. The opt-in design addresses that concern while removing substantial duplication, provided the non-SFP paths remain unchanged. Before calling the refactor training-equivalent, I will review the integration and rerun the real-GPU 100-step comparison from `pi05_libero_pytorch` with the same data/order, time/noise, and distributed settings. In particular, your description notes that the `train=True` image augmentation path differs, so the fixed-input `train=False` loss match alone does not establish matching training trajectories. I will also verify that SFP retains the correct cross-shard gradient norm and clipping while the general fix in #1558 is still separate.
> 
> Thanks again for contributing this. The proposed structure is a better fit for RLinf, and I am comfortable moving forward with it once those training and existing-model regression checks pass.

Hi @GhahahahaG, if the refactor looks good to you, could you merge it?

andylin-hao (approved):

guozhen1997:
Hi @GhahahahaG, could you please review and merge this PR https://github.com/GhahahahaG/RLinf/pull/2 ?
SFP eval previously had no trajectory sampler, so use_sfp was rejected for any task that samples actions, including eval. This change adds a chunk-synchronous sampler aligned with StreamingVLA (prefix KV once, integrate from action_states, dt = 1/H) and routes Pi0Eval through it, with env-space action_states carried across chunks. Scale-only SfpNormalize/SfpUnnormalize keep the train/eval action scale consistent. RL, DAgger, DSRL, and RTC stay unsupported; a single LIBERO Spatial eval recipe is added.

GhahahahaG:
@guozhen1997 I have integrated your GhahahahaG/RLinf#2 implementation into this branch and applied the two follow-up fixes in b8b10ae3: optimizer-step-wide SFP time/noise sampling with deterministic micro-batch slicing, and explicit episode-start propagation/reset for recurrent SFP eval state. GitHub now marks PR #2 as merged. Validation on the fix commit: Ruff and pre-commit passed; targeted tests 18 passed; broader suite 75 passed, 8 skipped, with one unrelated existing VideoMetadata/Transformers compatibility failure. The EN/ZH docs checks have been retriggered. Could you please re-review the updated PR #1498? The branch is currently behind main; synchronizing the latest main is pending a GitHub CLI workflow-scope refresh because main modifies .github/workflows/agent-e2e-tests.yml.

GhahahahaG:
@guozhen1997 Follow-up: the branch has now been updated with the latest main. The current head is 78e8c6c1, with parents b8b10ae3 and d5ec6690. Post-update validation passed: git diff --check, Ruff lint/format, repository pre-commit, and the focused SFP/random-slicing/episode-reset suite (19 passed). Both English and Chinese Read the Docs checks are also successful. The PR is mergeable and is now blocked only on the requested review. Please review the current head when convenient.

GhahahahaG:
@guozhen1997 The branch is synced with upstream main at `b023deb7`, and the conflict resolution is now at `6d910652`.

The only semantic conflict came from main replacing the flat `EnvOutput` fields with `EnvTransition`. `episode_starts` now lives in `EnvTransition` and is carried through CPU/contiguous conversion, `to_dict`, split/merge, the `EnvOutput` compatibility property, and env eval/reset construction. The obsolete flat-schema test was removed while both the SFP and new upstream test coverage were retained.

Validation: SFP/schema target tests 19 passed; affected data/models suite 176 passed, 8 skipped; Ruff, format, compileall, `git diff --check`, pre-commit, DCO, and EN/ZH Read the Docs all passed. One unrelated VLM Trend `VideoMetadata` test also fails on a clean `upstream/main` worktree with this server's Transformers version, so it is not introduced by this PR.

Could you please review the current head?

GhahahahaG:
Resolved the latest upstream/main conflict in b31e20b7dc3565af331b0ea1ae97a6d417aeaaf0 (parents: 6d910652 + 184ba07f). The SFP additions now live entirely in the unified rlinf.models.embodiment.openpi namespace introduced by #1580; the removed openpi_rlinf model type/path is not reintroduced. Existing OpenPI configs continue to use the upstream official loader and normalization path, while use_sfp opts into the LIBERO action_states loader, scale-only normalization, SFP objective, and sampler.\n\nValidation on the resolved tree:\n- focused SFP/data/model regression: 20 passed\n- full test_models.py + test_data.py: 176 passed, 8 skipped; the sole failure is the pre-existing Transformers VideoMetadata(frames_indices=...) incompatibility already reproduced on clean main\n- compileall, Ruff lint/format, git diff --check, and pre-commit (including DCO): passed\n- local EN + ZH Sphinx builds: 0 warnings in both\n- no openpi_rlinf / OPENPI_RLINF references remain\n\nGitHub now reports the PR mergeable; DCO is green and both Read the Docs jobs have restarted. @guozhen1997, please re-review this unified-OpenPI version when convenient.

guozhen1997:
Hi @GhahahahaG, sorry for the late reply. I have re-reviewed the PR and applied the follow-up changes. Streaming Flow Policy now rejects a non-π₀.₅ configuration up front, and the eval integrator resets only on episode_starts. I also aligned the docker mirror link and clarified that action_states is a copy of the actions statistics in docs. Thank you for contributing this pr.

Could you review my changes and merge the pr https://github.com/GhahahahaG/RLinf/pull/3?

PR #1518, opened 2026-08-27
feat(envs): unify world models backend

ZJLi2013:
﻿### Description

PR 1~3 of the plan in #1511: 

1. `WorldModelBackend` is a Protocol — `open_session` / `generate` / `close_session` / `offload` / `onload` — with `WanBackend` and `OpenSoraBackend` as the in-process implementations. The session owns the condition window, so the env hands it over once instead of re-sending five 256×256 frames per env per chunk, and a remote backend gets the trajectory boundaries it would need to cache. Window semantics are unchanged.

2.  `WanEnv` and `OpenSoraEnv` shared an 850-line episode loop almost line for line. What is left of each is `_build_backend` and `_load_reward_model` over a shared `WorldModelEnv`.

3. **Auto reset restarted every slot when any episode ended** (bug reported #1516). With 16 envs, one estimated success pushed all 16 back to a freshly sampled episode while the other 15 still reported `done=False`, in this pr make `elapsed_steps` for per env-slot in `BaseWorldEnv`, `reset()` takes `env_idx`, and `_handle_auto_reset` passes the indices from `dones`.

### Motivation and Context

**The goal is to serve the world model independently of the env**, eventually from another engine. That needs an interface, and this PR is what it takes to have one. The auto-reset bug #1516 comes along because a session is keyed by trajectory, and "which trajectory is on this slot" had no answer while every termination restarted everything.

### How has this been tested?

2×MI300X (ROCm 6.4). Both backends ran end to end with the same shape: 16 envs × 16 steps, two chunks, placement `2-3`, only `env.train` swapped.

- **Unit, 18 passed**, all stubbing their heavy deps so they run on the plain openvla-oft venv without a GPU or a checkpoint. `test_wan_backend.py` (5) and `test_opensora_backend.py` (5) cover session bookkeeping and each window's contents; `test_world_model_autoreset.py` (8) covers per-slot `elapsed_steps`, a subset reset leaving the other slots untouched, the shared time axis keeping its length, and `_handle_auto_reset(dones=[F, T, F, T])` restarting exactly slots 1 and 3.
- **Wan**, `wan_libero_spatial_grpo_openvlaoft` with `RLinf-Wan-LIBERO-Spatial`: exit 0, Global Step 1/1, `env_interact_step` 202.2 s against 201.7–210.2 s over six earlier runs of the same setup.
[...]

Issue #1511, opened 2026-08-26
[Feature]: decouple world model inference from the env so WM backends are pluggable


ZJLi2013:
### The problem

Two problems, one cause: the world model is built inside the env object.

**Every new world model is a new copy of the env.** `world_model_wan_env.py` (854 lines) and `world_model_opensora_env.py` (917 lines) are the same method for method outside model loading and the single pipe call. This already cost a fix: #1126 added non-CUDA device support to `base_world_env.py` + `world_model_wan_env.py` only, so `OpenSoraEnv.chunk_step` still hardcodes `device_type="cuda"`.

**One world model replica per env rank.** Each env builds its own pipeline in `__init__`, so the DiT + VAE replica count is `env_world_size * stage_num` — set by env parallelism, not by the generation batch — and `chunk_step` calls the pipeline synchronously. Generation, episode bookkeeping and the policy are three profiles that scale independently, which is why `cluster.component_placement` already separates actor, rollout and env. The world model is the one component with no such choice.

### The design

One `WorldModelEnv` owns episode bookkeeping; a backend advances frames. The interface is session-shaped rather than a stateless call because without open/close a serving backend cannot know when a trajectory begins, so it cannot pool anything.

```mermaid
classDiagram
    direction TB

    class BaseWorldEnv {
        <<abstract>>
        +cfg
    }

    class WorldModelEnv {
        <<abstract>>
        +backend : WorldModelBackend
        +reward_model
        +current_obs : Tensor
        +elapsed_steps : int
        +dataset
        +reset_state_ids
        +task_descriptions
        +reset()
        +chunk_step(policy_output_action)
        +offload()
        +onload()
        +get_state()
        #_build_backend()*
        #_load_reward_model()*
        #_reward_instructions()
        #_infer_next_chunk_frames(actions)
        #_infer_next_chunk_rewards()
        #_wrap_obs()
        #_handle_auto_reset()
        #_calc_step_reward()
        #_record_metrics()
    }

    class WanEnv {
        #_build_backend() WanBackend
        #_load_reward_model() ResnetRewModel
        #_reward_instructions() list
    }

    class OpenSoraEnv {
        #_build_backend() OpenSoraBackend
        #_load_reward_model() ResnetRM
    }

    class WorldModelBackend {
        <<Protocol>>
        +chunk : int
        +condition_frame_length : int
        +image_size : tuple
        +open_session(env_ids, init_frames, init_actions, task_ids, seeds)
[...]

PR #1519, opened 2026-08-27
refactor(envs): keep the Wan condition window in the backend

ZJLi2013:
﻿### Description

**Depends on #1518** â€” stacked on that branch, review it first. PR 2 of the plan in #1511.

The condition window moves from `WanEnv` into `WanBackend`:

- `open_session` takes `init_actions` alongside `init_frames`, so a session starts from a complete window: the reference frame the trajectory conditions on, the last generated frames, and the actions that produced them.
- `generate(env_ids, actions)` no longer takes `condition`. The env sends only the new action chunk; the backend prepends its own action history, conditions on its own frames, and rolls both forward.
- `WanEnv` loses `image_queue` and `condition_action` entirely.

Window semantics are unchanged: slot 0 stays the reset-time reference frame (KIR), slots 1â€“4 are the last four generated frames, and `retain_action` still prepends the action history to the chunk.

### Motivation and Context

A remote backend cannot cache anything per trajectory while the env re-sends the whole window on every call â€” five 256Ã—256 frames per env per chunk over the wire, with no way for the server to reuse the previous encode. With the window behind the session, that becomes the backend's choice.

It also removes the only conversion on this path that existed purely to cross the interface: the pipeline emits PIL frames, the env stored them as `[-1, 1]` float tensors, and the backend converted them back to uint8 on the next call. That round trip is not lossless â€” `x/255*2-1` restored through `((v+1)/2*255).astype(np.uint8)` truncates, so 63 of 256 gray levels come back one lower (~1/4 of pixels on a random image). The frames the pipeline conditioned on were systematically a gray level dark; keeping its own output fixes that.

#1511 tracks the rest: `WorldModelEnv` dedup + `OpenSoraBackend` (PR 3), the `world_model.backend` config key (PR 4), `RemoteBackend` (PR 5).

### How has this been tested?

Env: 2Ã—MI300X (ROCm 6.4), `wan_libero_spatial_grpo_openvlaoft`, 16 envs Ã— 16 steps, `RLinf-Wan-LIBERO-Spatial`.

1. **End to end.** `env_interact_step = 209.5 s` against 206.5 / 208.3 / 209.5 s on the previous trees; exit 0, Global Step 1/1. A byte-identical gate is not available here by construction: dropping the round trip changes the condition frames handed to the pipeline, which is the point of the change.
2. **Unit.** `tests/unit_tests/test_world_model_backend.py` â€” the window keeps its reference frame, rolls in the pipeline's own frames without a round trip, and prepends the action history (5 passed).

### Types of changes

### Checklist

### DIFF
--- a/rlinf/envs/world_model/backend.py
+++ b/rlinf/envs/world_model/backend.py
@@ -1,0 +1,70 @@
+# Copyright 2026 The RLinf Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""The interface between a world-model environment and the thing that generates frames.
+
+The env owns episode semantics — session lifetime, rewards, auto reset, metrics — and a backend only
+advances frames. It is session-shaped so a backend that pools per-trajectory state knows when a
+trajectory begins and ends. Implementations live next to this module, one file each.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Protocol, Sequence, runtime_checkable
+
+import torch
+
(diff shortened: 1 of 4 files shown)


ZJLi2013:
Folded into #1518, which now carries both commits: this PR's base was `main`, so its diff already contained #1518 and split the same change across two reviews. Keeping the two commits there lets each gate stay attached to the change it validates - byte-identical pipeline inputs for the code motion, the measured gray-level fix for the window move.

Issue #1502, opened 2026-08-24
[Feature]: SFT PI0.5 FSDP sharding_strategy


CarlDegio:
### Feature description

It is recommended to support hybrid shard for high communication efficiency multi-node training.

### Before submitting a new issue...

- [x] Have you checked relevant issues, FAQs (https://rlinf.readthedocs.io/en/latest/rst_source/resources/faq.html), or asked the chatbot at the top right corner of the [documentation page](https://rlinf.readthedocs.io).

kakiuwang-ui:
I'd like to pick this up. Some findings from digging into why `hybrid_shard` does not currently work:

`sharding_strategy: "hybrid_shard"` is already documented in `examples/*/config/hybrid_engines/fsdp.yaml` and accepted by `get_sharding_strategy()`, but it cannot be used, because `create_device_mesh()` always builds a **1-D** mesh:

```python
# rlinf/hybrid_engines/fsdp/utils.py
def create_device_mesh(world_size):
    return init_device_mesh(
        Worker.torch_device_type, mesh_shape=(world_size,), mesh_dim_names=["fsdp"]
    )
```

PyTorch requires a 2-D mesh for hybrid strategies. Minimal repro (4 gloo ranks, CPU):

```
1-D mesh (create_device_mesh today): ValueError: Expected device_mesh to have ndim=2 but got 1
2-D mesh (ddp, fsdp):                OK  shard_pg=2  inter_pg=2
```

There is a second, related symptom in `FSDPModelManager.__init__`:

```python
self._dp_group = (
    self._device_mesh["ddp"].get_group()
    if "ddp" in self._device_mesh.mesh_dim_names
    else None
)
```

`"ddp"` is never a dim name of the 1-D mesh, so `_dp_group` is always `None` — the branch is dead. It is the hook the gradient-norm all-reduce in `FSDPStrategy.clip_grad_norm_` / `get_grad_norm` needs once the mesh becomes 2-D.

Proposed change:

[...]

PR #1522, opened 2026-08-28
feat(fsdp): add hybrid sharding and reorganize guides

zhexuanxu:
### Description

Enable FSDP hybrid sharding for OpenPI RLinf SFT and organize scaling documentation by workload.

- Build a two-dimensional `(ddp, fsdp)` device mesh for `hybrid_shard` while preserving existing process-group semantics for other strategies.
- Derive actor-to-inference shard offsets from the mesh and validate hybrid topology consistently for FSDP1 and FSDP2.
- Synchronize the one-rank-per-node case as replicated data parallelism instead of silently skipping replica gradient reduction.
- Allow sharded OpenPI RLinf SFT and call the SigLIP stem through its module so nested FSDP hooks run.
- Add a BEHAVIOR config switch between keyframe streaming and resumable indexed sampling; the indexed path checkpoints per-rank dataloader and RNG state.
- Move Dynamic Scheduling and 5D Parallelism into Agentic examples and add equivalent English and Chinese FSDP practical-tips pages.

### Motivation and Context

Classic FSDP previously received a one-dimensional mesh, so `hybrid_shard` could not express intra-node sharding with inter-node replication. BEHAVIOR keyframe streaming also has mutable worker-local cursors that cannot be restored reliably. Indexed distributed sampling provides a deterministic checkpoint boundary for OpenPI SFT, including data order, LR, and stochastic model state.

### How has this been tested?

- `pre-commit run --all-files` passed, including Ruff lint, Ruff format, and commit sign-off checks.
- `pytest -q tests/unit_tests/test_fsdp_device_mesh_timeout.py` passed all 9 tests.
- `build_docs.py` built both English and Chinese docs with zero warnings; changed pages have no markup or EN-ZH parity findings.
- A true two-node run used two A800 GPUs per physical host, `cluster.num_nodes=2`, a `(ddp=2, fsdp=2)` hybrid mesh, and actor ranks 0-3. Ray exposed only two GPUs on each host, and the remote host's other two GPUs stayed idle.
- The uninterrupted 200-step run used `global_batch_size=64`, `micro_batch_size=16`, 8 dataloader workers per rank, seed 42, `chunk_streaming_using_keyframe=false`, and a checkpoint interval of 100. Mean loss fell from 0.28665 at steps 1-20 to 0.09563 at steps 181-200, a 66.64% decrease, and ended at 0.0798.
- Resuming `global_step_100` started at 100/200 and completed through step 200. Across all 100 overlapping steps (101-200), the raw TensorBoard values for loss, learning rate, and gradient norm were each exactly equal to the uninterrupted run (100/100 matches; maximum absolute difference and MAE were 0 for all three metrics).
- The step-100 checkpoint contained four FSDP DCP shards, four per-rank dataloader snapshots at offset 100, and four per-rank Python, NumPy, CPU Torch, and CUDA RNG states. The resumed step-200 checkpoint again contained four shards and four dataloader snapshots, now at offset 200.
- Four frames sampled through indexed and streaming paths produced pixel-exact transformed images, identical actions, padding, and prompts. The two-rank sampler had zero overlap and covered all 429,928 selected frames.

The 20-step mean losses over the overlapping interval were:

| Steps | Uninterrupted | Resumed |
| --- | ---: | ---: |
| 101-120 | 0.171000 | 0.171000 |
| 121-140 | 0.147700 | 0.147700 |
| 141-160 | 0.123900 | 0.123900 |
| 161-180 | 0.112235 | 0.112235 |
| 181-200 | 0.095630 | 0.095630 |

The multi-node Ray cluster followed the documented node-rank setup and communicated over Ethernet with RDMA disabled. Both runs completed without traceback, OOM, NCCL errors, or Ray worker restarts.

- Uninterrupted log: `logs/20260914-real-2node-2x2-hybrid-gbs64-mbs16-200step-uninterrupted/run_embodiment.log`
- Resume log: `logs/20260914-real-2node-2x2-hybrid-gbs64-mbs16-200step-resume-from-100/run_embodiment.log`
- Resume source: the uninterrupted run's `checkpoints/global_step_100`

### Types of changes

### Checklist:


rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
```diff
@@ -96,10 +96,14 @@ def __init__(self, cfg: DictConfig, world_size: int, rank: int) -> None:
...
+        self.device = Worker.torch_platform.current_device()
+
+        sharding_strategy = str(self._cfg.fsdp_config.sharding_strategy)
+        self._device_mesh = create_device_mesh(world_size, sharding_strategy)
         self._dp_group = (
-            self._device_mesh["ddp"].get_group()
-            if "ddp" in self._device_mesh.mesh_dim_names
+            self._device_mesh["fsdp"].get_group()
+            if sharding_strategy.lower() == "hybrid_shard"
             else None
```
qurakchin: for fully_shard, do not use None

zhexuanxu: Fixed in a026b68b. `_dp_group` now always comes from the `fsdp` mesh dimension: `full_shard` uses the full actor process group, while `hybrid_shard` uses the intra-node shard group.

rlinf/hybrid_engines/fsdp/utils.py
```diff
@@ -62,7 +63,18 @@ class FSDPVersion(str, Enum):
     FSDP2 = "fsdp2"
 
 
-def create_device_mesh(world_size):
+def create_device_mesh(
+    world_size: int, sharding_strategy: str = "full_shard"
+) -> DeviceMesh:
+    """Create the device mesh for an FSDP sharding strategy."""
+    if sharding_strategy.lower() == "hybrid_shard":
+        shard_size = int(os.environ["NODE_LOCAL_WORLD_SIZE"])
```
sherlockcooper: A bare `os.environ[...]` subscript here throws an opaque `KeyError` when `NODE_LOCAL_WORLD_SIZE` isn't set (e.g. single-process, tests, actors not launched via the worker group). Consider aligning with the Megatron sibling at `token_dispatcher.py:258-272`, which uses `os.getenv(...)` with a CUDA fallback and a clear error.

rlinf/hybrid_engines/fsdp/utils.py
```diff
@@ -62,7 +63,18 @@ class FSDPVersion(str, Enum):
...
+        return init_device_mesh(
+            Worker.torch_device_type,
+            mesh_shape=(world_size // shard_size, shard_size),
```
sherlockcooper, line 92: When `world_size % shard_size != 0` this silently truncates and `init_device_mesh` fails opaquely. Suggest a pre-check: `assert world_size % shard_size == 0, f"hybrid_shard requires world_size ({world_size}) divisible by actor ranks per node ({shard_size})"`.

rlinf/hybrid_engines/fsdp/utils.py
```diff
@@ -62,7 +63,18 @@ class FSDPVersion(str, Enum):
     FSDP2 = "fsdp2"
 
 
-def create_device_mesh(world_size):
+def create_device_mesh(
+    world_size: int, sharding_strategy: str = "full_shard"
+) -> DeviceMesh:
+    """Create the device mesh for an FSDP sharding strategy."""
+    if sharding_strategy.lower() == "hybrid_shard":
```
sherlockcooper: `.lower()` here but `get_sharding_strategy` (~L820) does exact lowercase match without lowercasing, so `Hybrid_Shard` builds the 2D mesh then fails the enum lookup. Consider aligning — e.g. lowercase in `get_sharding_strategy` too.

rlinf/hybrid_engines/fsdp/utils.py
zhexuanxu: Fixed in 30e797f4. When `NODE_LOCAL_WORLD_SIZE` is absent, the code now falls back to the available accelerator count (bounded by `world_size`) and raises a clear `RuntimeError` if it cannot infer the value. I used `Worker.torch_platform` instead of CUDA directly to preserve RLinf's accelerator abstraction.

rlinf/hybrid_engines/fsdp/utils.py
zhexuanxu: Fixed in 30e797f4. `shard_size` is now checked for positivity and `world_size % shard_size == 0` before `init_device_mesh`, with the relevant values included in the `ValueError`.

rlinf/hybrid_engines/fsdp/utils.py
zhexuanxu: Fixed in 30e797f4 by keeping the existing strict lowercase contract: `create_device_mesh` now uses the same exact `hybrid_shard` match as `get_sharding_strategy`, so mixed-case input is no longer partially accepted.

sherlockcooper (approved):

kakiuwang-ui:
I duplicated this work in #1542 before spotting this PR (it doesn't reference #1502, so it didn't show up on that issue's timeline — my screening error). I've closed mine in favour of this one, which is approved and has real two-node validation that mine doesn't.

Three things I ran into while going over the same code that this PR doesn't currently cover. The first is the one I'd genuinely worry about:

**1. The actor↔inference weight sync assumes a single shard group.**

`FSDPStrategyBase.setup_actor_sync_inference_ranks` and its inference-side counterpart advertise each rank's shard as:

```python
shard_size = (full_tensor_size + actor_world_size - 1) // actor_world_size
global_start = actor_rank * shard_size
```

That holds only while a parameter is sharded once across the whole world. With the 2-D mesh, rank `r` holds shard `r % shard_size` and the shards repeat along the `ddp` dim, so the advertised offsets describe data the rank doesn't own — `load_from_actors_by_intersection` then copies the wrong region (or slices past the end of the received buffer). The docstring on that function already flags multi-dim sharding as future work.

`sync_model_to_rollout` is fine, since it goes through `DTensor.full_tensor()`. So this only bites configs that launch an FSDP inference group, not the SFT case in #1502 — but once `hybrid_shard` is selectable, nothing stops someone setting it there, and the failure is silent rather than loud. A `NotImplementedError` in the handshake plus a config-time check would cover it. Worth noting the handshake guard has to be paired with a config-time one: it sits inside a cross-group collective, so if only one side is configured `hybrid_shard` the other blocks in the paired send/recv.

**2. An unknown `sharding_strategy` silently falls back to `full_shard` on FSDP2.**

`get_sharding_strategy()` asserts on the name, but it's only reached on the FSDP1 path (`strategy/fsdp.py`), and `validate_fsdp_cfg` only defaults the key. With the new `if sharding_strategy == "hybrid_shard"` branch, `"hybrid-shard"` / `"HYBRID_SHARD"` take the 1-D path and train under `full_shard` without a word. Validating the name inside `create_device_mesh` catches it for every component and both backends.

**3. `shard_size == world_size` builds a `(1, world_size)` mesh.**

Reachable on a single-node run where `NODE_LOCAL_WORLD_SIZE` equals the world size: replicate degree 1, so it's `full_shard` with extra machinery rather than an error.

I also wrote 23 CPU-only unit tests over the mesh shape/dim-name pairing, the degree checks and the rejection paths (no GPU or distributed init needed) — the dim order especially seems worth pinning, since `("fsdp", "ddp")` would silently swap sharding and replication rather than fail.

Happy to open a small follow-up PR on top of this one with the guard, the validation and the tests, once this merges — or to just hand over the tests if you'd rather fold them in here. Whichever is less disruptive.

andylin-hao (approved):

PR #1525, opened 2026-08-28
feat(starvla): add Ascend support for StarVLA

GanDing:
### Description

Runs the starVLA execution paths on whichever accelerator RLinf resolved for the worker instead of assuming CUDA, and moves the action-space helpers to the starVLA v1.6 API.

- Every `torch.autocast("cuda", ...)` in the starVLA backbone and the OFT, adapter, FAST and flowmatching heads goes through `accelerator_autocast(dtype)` in `starvla/utils/accelerator.py`. It autocasts on `Worker.torch_device_type` and falls back to a no-op context when the worker has no accelerator.
- `FSDPModelManager` builds its AMP context on the same device type. With the hardcoded `"cuda"`, `fsdp_config.amp_autocast` did nothing on non-CUDA backends; it now applies there, so non-CUDA runs that already enabled it will see mixed precision take effect.
- `build_gaussian(mean, std)` builds the policy `Normal` in float32. The OFT, adapter and flowmatching heads use it.
- `utils/action_space.py` unnormalizes through `starVLA.model.tools.FrameworkTools`. An `unnorm_key` missing from the checkpoint's `norm_stats` now raises an error listing the available keys.
- `requirements/install.sh` pins starVLA to the `starVLA-v1.6` tag instead of `starVLA-1.2`. Like the gr00t install, it calls `maybe_build_decord_from_source` (a source build on aarch64, a no-op elsewhere) and drops `eva-decord`, which ships the same `decord` module and has no aarch64 wheel.
- The new CI job `embodied-starvla-libero-ascend-test` runs `libero_10_grpo_starvla` on the Ascend runner.

Behavior on CUDA is unchanged: the autocast contexts resolve to `"cuda"` there, and the heads already produced float32 means under the float32 autocast.

### Motivation and Context

Training a starVLA policy on Ascend NPU fails in the OFT action head with:

```
tensor mean not implemented for DT_BFLOAT16
```

`torch.autocast("cuda", ...)` is a silent no-op on a non-CUDA accelerator: it does not raise, it just does not apply. The float32 autocast around the head never undoes the backbone's bfloat16, so the Gaussian gets a bfloat16 `loc` and sampling reaches `torch.normal`, which has no bfloat16 kernel in CANN. Keying the autocast on the resolved device type fixes that, and `build_gaussian` keeps the distribution in float32 whether or not an upstream autocast fired.

starVLA v1.6 moved its model-agnostic action helpers off `baseframework` onto `starVLA.model.tools.FrameworkTools`, so `baseframework.unnormalize_actions` raises `AttributeError` against a current checkout. The old `starVLA-1.2` pin predates `FrameworkTools`, and an unpinned clone would follow the moving `starVLA_dev` branch, so the install pins the v1.6 tag.

### How has this been tested?

- Training on 4x Ascend 910C with `total_num_envs=80`, `rollout_epoch=2`; the curve is below.
- CUDA e2e `libero_10_grpo_starvla` on 4x A100, from a fresh `install.sh` build against `starVLA-v1.6`: one full global step (checkpoint load, rollout with `FrameworkTools` unnormalization, weight sync, actor update) completes with exit 0. With the smoke config's 8-step episodes nothing succeeds, so GRPO filters every sample and the losses are zero; the run checks the pipeline, not the learning curve.
- `pytest tests/unit_tests/test_models.py`: 35 passed, 1 skipped. The starVLA cases cover norm-stats source selection, the unknown-key error, chunk reshaping with the LIBERO gripper mapping, the worker-keyed autocast and the float32 Gaussian. Each one fails when its fix is reverted.
- `pre-commit run` passes on the changed files.

image

CI reuses the checkout at `/workspace/dataset/starVLA` through `STARVLA_PATH`. It stays on `starVLA-1.2` until this PR is ready to merge, so the starVLA e2e jobs here will fail to import `FrameworkTools` until that checkout moves to `starVLA-v1.6`. The Ascend job has not run yet: it needs `Qwen2.5-VL-OFT-LIBERO-4in1`, the `Qwen2.5-VL-3B-Instruct` base model its config points at, and the `decord` checkout staged under `/workspace/dataset/` on the Ascend runner.

### Types of changes

### Checklist:


andylin-hao (approved):

rlinf/models/embodiment/starvla/utils/accelerator.py
```diff
@@ -0,0 +1,67 @@
...
+            ``torch.float32`` to undo it around an action head.
+
+    Returns:
+        A ``torch.autocast`` on ``Worker.torch_device_type``, or a no-op context
+        when the worker has no accelerator.
+    """
+    device_type = Worker.torch_device_type
+    if device_type is None:
+        return nullcontext()
+    return torch.autocast(device_type, dtype=dtype)
```
sherlockcooper, line 47: **The float32 autocast is still a no-op off CUDA.** This helper centralizes the pattern but keeps the behaviour.

In `torch/amp/autocast_mode.py`, `"cuda"` is the only branch that skips dtype validation. `cpu`, `xpu`, `mps`, `hpu` and `xla` each restrict to `{float16, bfloat16}`, and a PrivateUse1 backend (`npu` via `torch_npu`, `musa`) is checked against `custom_device_mod.get_amp_supported_dtype()`, which for `torch_npu` is fp16/bf16. A float32 request warns and sets `enabled=False`.

So on 910C the four `accelerator_autocast(torch.float32)` sites are still inert, the OFT and adapter heads still run in bfloat16, and `build_gaussian` is what makes training pass. The Motivation section's "Keying the autocast on the resolved device type fixes that" holds for the bfloat16 contexts but not the float32 ones.

Suggested fix: stop expressing "force fp32" as an autocast dtype. Give the helper an explicit disable path, `torch.autocast(device_type, enabled=False)`, and cast the region's inputs with `.float()`. That is backend-independent and is what those four regions actually want.

Worth confirming on the 910C, since `get_amp_supported_dtype` is the deciding call:

```python
import torch, torch_npu
print(torch.npu.get_amp_supported_dtype())
print(torch.autocast("npu", dtype=torch.float32)._enabled)
```

tests/unit_tests/test_models.py
```diff
@@ -294,6 +296,122 @@ def test_value_model_does_not_rescale_gemma3_language_embeddings(monkeypatch):
...
+    np.testing.assert_array_equal(env_actions[0, :, 6], -1.0)
+    np.testing.assert_array_equal(env_actions[1, :, 6], 1.0)
+
+
+def test_starvla_autocast_targets_the_worker_accelerator(monkeypatch):
+    accelerator = _load_starvla_util("accelerator")
+    # CPU stands in for a non-CUDA accelerator such as an Ascend NPU.
+    monkeypatch.setattr(Worker, "torch_device_type", "cpu")
+
+    with accelerator.accelerator_autocast(torch.bfloat16):
```
sherlockcooper, line 387: This test would have caught the float32 problem, but it only covers bfloat16, which is the one dtype every backend accepts.

The same test with the dtype the four head sites actually request fails on this branch:

```python
def test_starvla_autocast_targets_the_worker_accelerator_in_float32(monkeypatch):
    accelerator = _load_starvla_util("accelerator")
    monkeypatch.setattr(Worker, "torch_device_type", "cpu")

    with accelerator.accelerator_autocast(torch.float32):
        assert torch.is_autocast_enabled("cpu")
```

```
E   AssertionError: assert False
warnings: ['In CPU autocast, but the target dtype is not supported. Disabling autocast.']
```

Worth parametrizing this case over both dtypes once the helper is fixed.

rlinf/models/embodiment/starvla/action_heads/flowmatching.py
```diff
@@ -657,11 +652,7 @@ def run_rollout_flowmatching(
     # model under torch.autocast("cuda", dtype=torch.float32).  Without this,
     # the ODE integration inherits bf16 from the backbone and accumulates
     # truncation errors over num_steps Euler steps.
-    fp32_ctx = (
-        torch.autocast("cuda", dtype=torch.float32)
-        if rollout_hidden.is_cuda
-        else nullcontext()
-    )
+    fp32_ctx = accelerator_autocast(torch.float32)
```
sherlockcooper, line 655: Following from the `accelerator_autocast` comment: because this context is inert on Ascend, the ODE state changes dtype from step to step depending on the data.

`actions_t` is seeded just below at `dtype=rollout_hidden.dtype`, bfloat16 on NPU. `build_gaussian` returns a float32 distribution, so `sampled_actions` is float32 while `mean_next` stays bfloat16. On a step where `active_step_mask.any()` holds, the `torch.where` a few lines down promotes `actions_t` to float32; on a step where it does not, `actions_t = mean_next` stays bfloat16. The integration dtype therefore flips across Euler steps according to which envs are active.

On CUDA the whole loop was float32, so this is new behaviour confined to non-CUDA backends, and the comment right above this line is precisely about not letting the ODE inherit bf16.

Suggested fix: seed `actions_t` in float32 and keep it there for the loop. This falls out of the autocast fix.

rlinf/models/embodiment/starvla/action_heads/oft.py
```diff
@@ -94,7 +95,7 @@ def _run_oft_backbone_and_head(
     )
     model = policy.starvla_model
     last_hidden = backbone_output["last_hidden"]
-    with torch.autocast("cuda", dtype=torch.float32):
+    with accelerator_autocast(torch.float32):
```
sherlockcooper, line 98: One of the four float32 sites that stay inert on Ascend, see the comment on `accelerator.py`. The head still runs in bfloat16 here; the crash is avoided downstream by `build_gaussian`, not by this context.

rlinf/models/embodiment/starvla/action_heads/adapter.py
```diff
@@ -243,7 +243,7 @@ def inject_query_hook(_module, _inputs, output):
         state_projected = model.proprio_projector(proprio=proprio)
 
     # 8) Predict actions from the fused multi-layer vision/query/state features.
-    with torch.autocast("cuda", dtype=torch.float32):
+    with accelerator_autocast(torch.float32):
```
sherlockcooper, line 246: Same as `oft.py`: this float32 context does not apply on a non-CUDA accelerator. See the comment on `accelerator.py`.

rlinf/models/embodiment/starvla/action_heads/flowmatching.py
```diff
@@ -489,11 +488,7 @@ def run_default_forward_flowmatching(
 
     # Match the rollout path which runs ODE integration under fp32 autocast
     # to avoid bf16 truncation errors accumulating over Euler steps.
-    fp32_ctx = (
-        torch.autocast("cuda", dtype=torch.float32)
-        if rollout_hidden.is_cuda
-        else nullcontext()
-    )
+    fp32_ctx = accelerator_autocast(torch.float32)
```
sherlockcooper, line 491: Same float32 context in the `default_forward` path. It replays a recorded chain rather than integrating, so there is no dtype-flip here, but the region still does not get the float32 it asks for on Ascend.

rlinf/models/embodiment/starvla/utils/action_space.py
```diff
@@ -186,12 +180,8 @@ def unnormalize_actions_for_env(
...
         )
 
-    try:
-        from starVLA.model.framework.base_framework import baseframework
-    except Exception as exc:
-        raise ModuleNotFoundError(
-            "starVLA is required for action unnormalization but is not importable."
-        ) from exc
+    # starVLA is installed only in the starvla venv.
+    from starVLA.model.tools import FrameworkTools
```
sherlockcooper, line 184: This import still lost its explanatory error. On `main` it was wrapped:

```python
raise ModuleNotFoundError(
    "starVLA is required for action unnormalization but is not importable."
) from exc
```

The comment added above documents the lazy import, but a user running without the starvla venv now gets a raw `ImportError` naming an internal module path instead of the message that tells them what to do. Restoring the `try` / `raise ... from exc` wrapper would fix it.

.github/workflows/embodied-e2e-tests.yml
```diff
@@ -1394,6 +1394,36 @@ jobs:
...
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Create Ascend embodied environment
+        run: |
+          unset UV_DEFAULT_INDEX
+          export UV_PATH=/workspace/dataset/.uv
+          export UV_CACHE_DIR=/workspace/dataset/.uv_cache
+          export UV_PYTHON_INSTALL_DIR=/workspace/dataset/.uv_python
+          export STARVLA_PATH=/workspace/dataset/starVLA
```
sherlockcooper, line 1409: The v1.6 bump currently relies on the runner's checkout having been moved by hand, as the description notes for the embodied runner.

`clone_or_reuse_repo` only *warns* on a ref mismatch and never updates an existing checkout, so that guarantee lives outside the repo. The next pin bump, a fresh runner, or this Ascend runner's own staged checkout can silently reuse the wrong ref and fail deep inside a rollout rather than at install time.

Putting the ref in the path, `STARVLA_PATH=/workspace/dataset/starVLA-v1.6` in both starVLA jobs, would make a pin bump force a fresh clone instead of relying on someone remembering. Promoting that mismatch in `clone_or_reuse_repo` from a warning to a hard error would work too.

sherlockcooper (commented):
Re-reviewed at `eb1f5c05`. The pin bump, the dead `get_action_stats` branch, the duplicated autocast blocks and the test coverage all look resolved. I checked that `starVLA-v1.6` really does carry `FrameworkTools.get_action_stats` / `unnormalize_actions` with the signatures the call sites expect, and `pytest tests/unit_tests/test_models.py` gives 35 passed, 1 skipped locally.

Findings are inline. The first one is load-bearing: the float32 autocast is still a no-op off CUDA, so `build_gaussian` is what makes Ascend training pass, not the autocast change.

One finding has no line to hang on. The docs checkboxes are still unticked although this PR adds a supported hardware/model combination and a CI job asserting it. `docs/source-en/rst_source/guides/ascend_cann.rst` is written entirely around OpenVLA-OFT, down to `source switch_env openvla-oft`, and `docs/source-en/rst_source/examples/embodied/starvla.rst` mentions only GPUs, so nothing tells a user that starVLA runs on Ascend or which install line to use. A short starVLA section in the Ascend guide carrying the new job's own invocation, plus a one-line pointer from the starVLA example page, would cover it. Both need the matching ZH edit under `docs/source-zh/`.

panhe1818:
@GanDing Hi, we have some comments above. Could you please take a look and fix those when you get a chance? Thanks.

rlinf/models/embodiment/starvla/action_heads/flowmatching.py
GanDing: The `actions_t` is seeded in float32, later `_predict_velocity` -> `head.action_encoder(actions_t, t_bucket_index)` will throw errors. It is because the head is loaded in bf16, previously CUDA can handle this type conversion.

One solution is changing the model precision to `fp32` in the yaml so the weight are loaded in `fp32`, but it will cause extra mem usage and computation cost.

rlinf/models/embodiment/starvla/utils/accelerator.py
GanDing: I confirmed on existing 910C env the `get_amp_supported_dtype` are only `[float16 and bfloat16]`. There maybe float32 in the newer version of `torch_npu`.

I think this PR is mainly to support StarVLA training on npu. This accuracy issue can be handled by another PR which will require more code refactoring (if we keep loading weight in `bf16`, please check the comment about `action_t`)

rlinf/models/embodiment/starvla/action_heads/flowmatching.py
sherlockcooper: No objection to deferring the precision work. One small thing for this PR: the comment just above this line still says the fp32 context is what keeps the ODE from inheriting bf16, which is now the opposite of what happens on Ascend. There's a shorter copy of the same claim at `:489-490`.

Something like:

```python
# On CUDA this runs the action model in float32, matching starVLA's
# QwenGR00T.predict_action. On backends whose autocast rejects float32
# (Ascend: get_amp_supported_dtype() is [fp16, bf16]) it falls through
# to the backbone dtype, so the Euler steps do accumulate truncation.
# TODO(agent): integrate in float32 independently of autocast.
```

Keeps the tradeoff visible in the code rather than only in this thread.

sherlockcooper (approved):

GanDing:
@guozhen1997 You can try upgrading CANN to 9.1.0 on the CI image, then you don't need the Qwen VL patch embedding. CANN 9.1.0 fixed that issue.

[9.1.0](https://www.hiascend.com/document/detail/zh/CANNCommunityEdition/910/softwareinst/releasenote/9.1.0/release-notes.md)

guozhen1997 (approved):

PR #1534, opened 2026-09-02
feat(embodied): recompute prev logprobs on the actor

ZJLi2013:
### Description

`algorithm.recompute_logprobs` now works for the embodied FSDP actor: `prev_logprobs` are rescored with the actor's own forward, in the same micro-batches the update loop will train on. Off by default, and it reports the rollout-to-actor gap it removes as `actor/rollout_train_logprob_gap`.

The key is already in every embodied config, but only GR00T ever recomputed, inside its own forward; OpenVLA / OpenVLA-OFT used the rollout's logprobs as `old_logprobs`. Other model types and `adv_type: opd` are asserted out.

### Motivation and Context

The rollout worker and the actor score the same action tokens on different paths — eval vs train mode, the actor's AMP context, FSDP-gathered weights, different batch shapes — so their logprobs differ on identical weights, and that difference lands in the PPO ratio.

With `lr=0` on LIBERO-spatial GRPO + OpenVLA-OFT, where the policy provably has not moved: `clip_fraction` 0.046 per token, `approx_kl` 0.0074, mean absolute logprob difference 0.064 nats. So ~4.6% of tokens sit outside the clip band before any policy movement — the trust region is spent on numerics. After recompute with the aligned split, `ratio` is 1.000 and all three read 0.

### How has this been tested?

4x MI300X (ROCm), LIBERO-spatial, OpenVLA-OFT, GRPO, `update_epoch=1`:

- `lr=0`, recompute off — the baseline readings above; recompute on — `ratio` 1.000, the three out-of-band metrics 0. Splitting the recompute any other way (flatten) only moves `clip_fraction` 0.319 to 0.271, which is what forced the aligned split.
- Recompute on, `lr=1e-5`, 20 steps: no collapse, `eval/success_once` 0.448 to 0.572 on the full 500-trajectory enumeration.
- Same knob on a second environment (Wan world model as env, `wan_libero_spatial_grpo_openvlaoft`, 20 steps): `ratio` 1.000 on the steps where every rank has surviving samples, gap 0.040 to 0.060, `eval/success_once` 0.566 on the same 500-trajectory enumeration. The batch shape there is half as long per episode, so this also covers a split the LIBERO runs do not exercise.

`clip_fraction` and `approx_kl` are quoted per token; the logged values are `action_dim`x larger ([#1471](https://github.com/RLinf/RLinf/issues/1471)).

### Additional information

Truncated importance sampling is deliberately not part of this PR. On a synchronous run the rollout weights are the training weights, so the residual gap has no direction and reweighting the advantages by it would add variance without correcting a bias. It becomes meaningful once the rollout engine differs from the training forward or the weights genuinely lag, and it needs this PR's actor-side logprobs to exist first.

### Types of changes

### Checklist

- [x] Code style follows the project (Ruff, Google-style docstring).
- [ ] No docs update yet — the config key is documented on the reasoning side only.
- [ ] No new test. Existing embodied e2e configs keep the key off, so they cover the unchanged path; the recompute path was verified on hardware as above, not in CI.


rlinf/workers/actor/embodied_fsdp_actor_worker.py
```diff
@@ -388,6 +388,91 @@ def compute_opd_teacher_logprobs(self) -> None:
...
+        the update loop's split, so each sample is scored in the micro-batch it
+        will be trained in.
+
+        Args:
+            batch_size_per_rank: Samples per optimizer step on this rank.
+
+        Returns:
+            Dict with ``actor/rollout_train_logprob_gap`` (mean absolute logprob difference).
+        """
+        assert SupportedModel(self.cfg.actor.model.model_type) in [
```
andylin-hao: It'd better to move the assertion check into config.py. Also, should we assert openvla/oft only, or assert not gr00t according to the comment?

rlinf/workers/actor/embodied_fsdp_actor_worker.py
```diff
@@ -388,6 +388,91 @@ def compute_opd_teacher_logprobs(self) -> None:
...
+            SupportedModel.OPENVLA,
+            SupportedModel.OPENVLA_OFT,
+        ], (
+            "algorithm.recompute_logprobs only supports OpenVLA models; GR00T "
+            "already recomputes inside its training forward."
+        )
+        assert self.cfg.algorithm.adv_type != "opd", (
```
andylin-hao: Same as above

andylin-hao (approved):

rlinf/workers/actor/embodied_fsdp_actor_worker.py
ZJLi2013: applied

Lin-xs (approved):

PR #1535, opened 2026-09-02
feat(metrics): report loss mask coverage

ZJLi2013:
### Description

`compute_rollout_metrics` now reports `loss_mask_fraction`, the share of the rollout batch that survives the loss mask, alongside the metrics that are already taken over that mask.

- New rollout metric `loss_mask_fraction`, reduced across ranks the same way as the existing ones. Nothing is renamed or removed, and runs without a loss mask are unaffected.

### Motivation and Context

Every reward, advantage and logprob metric in that function is a masked mean, and `filter_rewards` can drop whole groups when a group's rewards are uniform. A batch where most groups were dropped therefore produces metrics that look clean because little survived, and there is currently no way to tell that case apart from a healthy one.

The distinction changes what you tune. Low coverage means the effective batch shrank, so the gradient noise grew â€” the knobs are `total_num_envs`, `group_size` and the `filter_rewards` bounds. Without the coverage reading, an unstable run of this kind looks like too large a learning rate, and the learning rate is the wrong knob for it.

### How has this been tested?

4x MI300X (ROCm), OpenVLA-OFT, GRPO, `filter_rewards: [0.5, 4.5]`, 20 optimizer steps per run:

- LIBERO-spatial with the MuJoCo env: `loss_mask_fraction` sits at 0.33 across the run, matching `group_size: 8` with roughly a third of the groups surviving the filter.
- The same policy and knobs against the Wan world-model env: coverage swings between 0.092 and 0.455 with a mean of 0.27, i.e. some optimizer steps train on a tenth of the batch. That is invisible in the other metrics â€” this is the reading that motivated the patch.

`pytest tests/unit_tests/test_metric_utils.py` â€” 5 passed, including two new cases that pin the fraction for a synthetic mask and check it stays absent when there is no mask.

### Additional information

While reading these numbers it is worth knowing that `masked_mean` returns 0 when a chunk is fully masked out (`rlinf/utils/utils.py`), and that 0 is averaged into `actor/ratio` and friends, so on low coverage those metrics get diluted rather than reported as empty. Not addressed here.

### Types of changes

### Checklist

- [x] Code style follows the project (Ruff, Google-style docstring).
- [x] The change is metrics-only and does not alter training behaviour.
- [x] Unit tests added in `tests/unit_tests/test_metric_utils.py`.


ZJLi2013:
Both review points are addressed in a0f5d30:

- Reworded the comment above the metric into a full sentence (Report the share of the batch that did.).
- Added unit_tests/test_metric_utils.py: reports 5 passed.

andylin-hao (approved):

Lin-xs (approved):

panhe1818:
Hi, the CI pipeline is failing. Could you please take a look and fix this when you get a chance? Thanks. @ZJLi2013

ZJLi2013:
> Hi, the CI pipeline is failing. Could you please take a look and fix this when you get a chance? Thanks. @ZJLi2013

fixed, thanks again

PR #1542, opened 2026-09-04
feat(fsdp): support hybrid_shard via a 2-D device mesh

### LINKED ISSUE #1502: [Feature]: SFT PI0.5 FSDP sharding_strategy

kakiuwang-ui:
### Description

`sharding_strategy: "hybrid_shard"` is documented in every `examples/*/config/hybrid_engines/fsdp.yaml` and accepted by `get_sharding_strategy()`, but it could not be used: `create_device_mesh()` always built a 1-D mesh, and PyTorch requires a 2-D mesh for hybrid strategies.

```python
# before
def create_device_mesh(world_size):
    return init_device_mesh(
        Worker.torch_device_type, mesh_shape=(world_size,), mesh_dim_names=["fsdp"]
    )
```

Reproduced with 4 gloo ranks on CPU, driving `torch.distributed.fsdp._init_utils._init_process_group_state` with the mesh `create_device_mesh` produces:

```
1-D mesh (create_device_mesh today): ValueError: Expected device_mesh to have ndim=2 but got 1
2-D mesh (ddp, fsdp):                OK  shard_pg=2  inter_pg=2
```

The change:

1. **`resolve_fsdp_mesh()`** returns `mesh_shape=(world_size // hybrid_shard_size, hybrid_shard_size)` with `mesh_dim_names=("ddp", "fsdp")` for `hybrid_shard`, and the current 1-D `("fsdp",)` mesh for everything else. `create_device_mesh()` is now a thin wrapper over it.
2. **New `fsdp_config.hybrid_shard_size`**: ranks per shard group. Set it to the accelerators on one node to shard intra-node and replicate inter-node.
3. **`FSDPModelManager._dp_group`** now resolves to the shard (`"fsdp"`) group.
4. **Validation** for unknown strategy names, degenerate shard/replicate degrees, and `hybrid_shard` on runs that configure an FSDP inference group.
5. `hybrid_shard_size` documented in the four `hybrid_engines/fsdp.yaml` files and in the EN + ZH `fsdp_config` tables in `guides/agentic_config.rst`.

### Motivation and Context

Closes #1502 — @CarlDegio asked for hybrid shard for communication efficiency in multi-node training, in the context of PI0.5 SFT.

**Why the dim order is `("ddp", "fsdp")` and not the reverse.** It is part of the contract, not cosmetic. FSDP1's `_init_process_group_state_for_hybrid_shard` takes mesh dim 0 as `_inter_node_pg` and dim 1 as `process_group`; FSDP2's `fully_shard` reads a 2-D named mesh as `HSDPMeshInfo(shard_mesh_dim=1, replicate_mesh_dim=0)`. One mesh therefore gives both backends the same HSDP layout, and reversing it silently swaps shard and replicate.

**Why `hybrid_shard_size` is required rather than defaulted from the node-local rank count.** My first version defaulted to `LOCAL_WORLD_SIZE`. That is wrong twice over: `init_device_mesh` is collective, so ranks that disagree on the shape hang rather than error, and an uneven placement gives different per-node counts (a 4+4+8 layout over world 16 has both `(4,4)` and `(2,8)` validating on different ranks). FSDP worker groups also launch with `isolate_gpu=True`, which pins `LOCAL_WORLD_SIZE` to `1` in `Worker._setup_local_rank_world_size`, so the default would have resolved to `1` on every rank anyway. A configured value is identical on every rank by construction.

**Why `_dp_group` changed from `"ddp"` to `"fsdp"`.** The old lookup was dead code — the mesh was always 1-D `("fsdp",)`, so `"ddp" in mesh_dim_names` was never true and `_dp_group` was always `None`. It is the hook the grad-norm all-reduce in `FSDPStrategy.clip_grad_norm_` and `get_grad_norm` needs, and the shard group is the correct one: under HSDP a rank holds one shard of the gradient and the shards are replicated along `ddp`, so summing the p-norm over `ddp` or over WORLD over-counts it by the replicate degree.

**Why `hybrid_shard` is rejected for runs with an FSDP inference group.** The actor↔inference handshake in `setup_actor_sync_inference_ranks` advertises a rank's shard as `rank * ceil(numel / world_size)`, which only holds while a parameter is sharded once across the whole world; its own docstring notes multi-dim sharding as future work. Under HSDP a rank holds shard `rank % hybrid_shard_size` and the shards repeat along the replicate dim, so the advertised offsets would not match the data and weights would be synced from the wrong region. `sync_model_to_rollout` is unaffected — it goes through `DTensor.full_tensor()`, which is correct over a 2-D mesh.

### How has this been tested?

**Unit tests** — `tests/unit_tests/test_fsdp_hybrid_shard_mesh.py`, 23 tests, CPU-only, no GPU or distributed init needed. They cover the mesh shape and dim-name pairing, the shard/replicate degrees, every rejection path, and the two config guards.

To confirm the tests discriminate rather than decorate, I broke the implementation eight ways, re-ran, and checked each failure landed on the intended assertion before restoring:

| Deliberate break | Result |
| --- | --- |
| dim order reversed to `("fsdp", "ddp")` | 3 tests fail |
| degenerate shard/replicate guard removed | 2 fail |
| 2-D mesh returned for every strategy | 3 fail |
| weight-sync handshake guard removed | 1 fails |
| unset `hybrid_shard_size` accepted | 3 fail |
| strategy-name whitelist removed | 3 fail |
| weight-sync config gate always returns early | 3 fail |

**End-to-end against real torch** (4 gloo ranks on CPU, torch 2.11), feeding `resolve_fsdp_mesh`'s output to both backends:

```
resolve_fsdp_mesh -> shape=(2, 2) names=('ddp', 'fsdp')
FSDP1 shard pg size=2 replicate pg size=2
FSDP2 mesh info: HSDPMeshInfo shard_mesh_dim=1 replicate_mesh_dim=0
_dp_group (mesh['fsdp']) size=2  WORLD size=4
all_reduce over _dp_group summed the shard group only: 1.0 (WORLD would give 6.0)
```

**Regression** — ran the CPU unit-test suite file-by-file as CI does, on this branch and on `main`. Identical results; the only failures are pre-existing local environment noise (`test_auto_placement` and `test_auto_resume_checkpoint` error on a missing `torchdata`, `test_distributed_ray_log_collector` has one failure, three GPU-only files collect nothing) and reproduce unchanged on `main`. `ruff check --preview` and `ruff format --check` are clean.

**Not tested**: I have no multi-node GPU access, so a real HSDP training run is unverified. What is verified is that the mesh both backends receive is the HSDP mesh they expect, and that the grad-norm reduction spans the shard group.

### Notes for review

- **The weight-sync rejection is conservative, not exact.** Whether an FSDP inference group is actually launched depends on the entry script — for reasoning it is the placement mode plus `algorithm.recompute_logprobs` in `main_grpo.py`. `validate_fsdp_weight_sync_cfg` keys off the presence of an `inference.fsdp_config` section instead, because restating that launch condition in `config.py` would duplicate logic that can drift. It can therefore reject a run whose inference group would not have been created. If you would rather have the exact condition, say so and I will thread the placement mode through. SFT-style runs with no `inference` section — the case #1502 asks for — are unaffected.
- **One adjacent bug fixed, deliberately.** Passing `_dp_group` unconditionally also fixes a pre-existing FSDP2 defect: `get_grad_norm` wraps both reductions in `if dp_group is not None`, so `fsdp2` + `full_shard` previously performed no cross-rank reduction at all and every rank clipped on its own shard's partial norm. I folded it in because the alternative was leaving `hybrid_shard` and `full_shard` clipping on different bases in code this PR already rewrites. Happy to split it into its own PR if you would rather review it separately.
- **Adjacent, found but deliberately not fixed**: making the actor↔inference handshake HSDP-aware. It would need the shard layout in the exchanged metadata rather than the `rank * ceil(numel / world_size)` assumption. That is a bigger change with a wire-format implication, and I cannot test it without multi-GPU access. Guarded here instead; happy to open a follow-up issue.
- **`hybrid_shard_size` naming**: I asked in the issue whether you prefer this or verl's `fsdp_size`, which the surrounding code looks derived from, and got no answer. Trivial to rename.
- The runtime guard in `_assert_single_shard_group` is kept as a backstop for callers that bypass config validation, but with the config check in place it should never fire in a Hydra-launched run.

### Types of changes

### Checklist:

### DIFF
--- a/docs/source-en/rst_source/guides/agentic_config.rst
+++ b/docs/source-en/rst_source/guides/agentic_config.rst
@@ -607,7 +607,15 @@ Used when ``actor.training_backend`` is ``fsdp``.
      - FSDP strategy: ``fsdp`` or ``fsdp2`` (case-insensitive).
    * - ``actor.fsdp_config.sharding_strategy``
      - Sharding strategy: ``full_shard``, ``shard_grad_op``, ``hybrid_shard``, or
-       ``no_shard``.
+       ``no_shard``. ``hybrid_shard`` is rejected for runs that configure an FSDP
+       inference group, whose weight-sync metadata assumes a single shard group.
+   * - ``actor.fsdp_config.hybrid_shard_size``
+     - Ranks per shard group when ``sharding_strategy`` is ``hybrid_shard``:
+       parameters are sharded inside a group of this many ranks and replicated
+       across the remaining ranks, so set it to the accelerators on one node to
+       shard intra-node and replicate inter-node. Required by ``hybrid_shard``;
+       must divide the world size of the component it applies to and leave a
+       shard degree and a replicate degree of at least 2. Ignored otherwise.
    * - ``actor.fsdp_config.cpu_offload``
      - FSDP2: keep parameters on CPU, moving them to GPU only when needed.
    * - ``actor.fsdp_config.offload_pin_memory``
--- a/docs/source-zh/rst_source/guides/agentic_config.rst
+++ b/docs/source-zh/rst_source/guides/agentic_config.rst
@@ -597,6 +597,13 @@ actor
      - FSDP 策略：``fsdp`` 或 ``fsdp2``\ （不区分大小写）。
    * - ``actor.fsdp_config.sharding_strategy``
      - 分片策略：``full_shard``、``shard_grad_op``、``hybrid_shard`` 或 ``no_shard``。
+       配置了 FSDP inference 组的运行不支持 ``hybrid_shard``，因为其权重同步元数据
+       假设只有一个覆盖全部 rank 的分片组。
+   * - ``actor.fsdp_config.hybrid_shard_size``
+     - ``sharding_strategy`` 为 ``hybrid_shard`` 时每个分片组的 rank 数：参数在该数量的
+       rank 内分片，并在其余 rank 上复制。设为单节点的加速卡数即可实现节点内分片、
+       节点间复制。使用 ``hybrid_shard`` 时必须设置；它必须整除所属组件的 world size，
+       且分片度与复制度均不小于 2。其他分片策略会忽略该项。
    * - ``actor.fsdp_config.cpu_offload``
      - FSDP2：参数保留在 CPU，仅在需要时移到 GPU。
    * - ``actor.fsdp_config.offload_pin_memory``
--- a/examples/embodiment/config/hybrid_engines/fsdp.yaml
+++ b/examples/embodiment/config/hybrid_engines/fsdp.yaml
@@ -6,6 +6,13 @@ strategy: "fsdp"            # FSDP strategy: ["fsdp", "fsdp2"]
 # "no_shard" (no sharding)
 sharding_strategy: "full_shard"
 
+# Ranks per shard group when sharding_strategy is "hybrid_shard": parameters are
+# sharded inside a group of this many ranks and replicated across the remaining
+# ranks. -1 means one shard group per node (shard intra-node, replicate
+# inter-node). Must divide the actor world size and leave a replicate degree of
+# at least 2. Ignored by every other sharding strategy.
+hybrid_shard_size: -1
+
 gradient_checkpointing: False # if True, gradient checkpointing will be used for FSDP/FSDP2 training, which will save memory but increase computation time.
 cpu_offload: False            # whether to offload parameters and gradients to CPU when not in use, if True, actor's enable_offload should be True too.
 offload_pin_memory: False     # whether FSDP2's CPU offload policy should pin memory when cpu_offload is True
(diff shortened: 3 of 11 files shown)


kakiuwang-ui:
Closing as a duplicate of #1522, which implements the same 2-D `("ddp", "fsdp")` mesh, is already approved, and has two-node hardware validation this PR doesn't.

I missed it during screening: #1522 doesn't reference #1502, so it never appeared on that issue's timeline, and I didn't keyword-match it against open PR titles before starting. My mistake, and sorry for the extra review surface.

The parts of this PR that #1522 doesn't cover — a guard for the actor↔inference weight-sync path (which assumes a single shard group and would silently sync from the wrong region under HSDP), validation of unknown `sharding_strategy` names on the FSDP2 path, and 23 CPU-only unit tests over the mesh contract — I've written up in https://github.com/RLinf/RLinf/pull/1522#issuecomment-5536869571 and am happy to bring as a small follow-up on top of #1522.

PR #1544, opened 2026-09-06
fix(fsdp): honor RLINF_TIMEOUT for FSDP collectives

### LINKED ISSUE #996: [Feature]: FSDP Actor NCCL timeout is hardcoded to PyTorch default (30 min), not configurable
### Feature description

  **Problem**

  The FSDP Actor's default torch.distributed process group is implicitly initialized by   DeviceMesh._get_or_create_default_group(), which calls init_process_group() with no arguments, resulting in   PyTorch's default timeout of 1800 seconds (30 minutes).

  RLinf's MultiChannelProcessGroup has its own configurable timeout (RLINF_TIMEOUT, default 180 min), but   FSDP's internal collective operations (all_gather, reduce, etc.) use the default process group and are not   affected by RLINF_TIMEOUT.

  This causes false timeout kills in two scenarios:
  1. Long training steps (e.g. large gradient accumulation) where a rank's normal computation exceeds 30   minutes between collectives.
  2. Debugging a specific rank (e.g. attaching a debugger to rank 3), which pauses that rank and causes all   other ranks to hit the watchdog timeout.

  WorkNCCL(SeqNum=1878, OpType=_ALLGATHER_BASE, ..., Timeout(ms)=1800000) ran for 1800000 milliseconds before   timing out.

  **Root Cause**

  rlinf/hybrid_engines/fsdp/fsdp_model_manager.py calls create_device_mesh() without first explicitly  initializing torch.distributed. PyTorch's DeviceMesh.__init__ detects this and auto-calls   init_process_group() with the default timeout (1800s), which cannot be overridden externally.

  **Suggested Fix**

  Explicitly call torch.distributed.init_process_group(backend="nccl", timeout=...) before  create_device_mesh() in FSDPModelManager.__init__, with the timeout value configurable via environment   variable or config.

### Additional context

_No response_

- [x] Have you checked relevant issues, FAQs (https://rlinf.readthedocs.io/en/latest/rst_source/faq.html), or asked the chatbot at the top right corner of the [documentation page](https://rlinf.readthedocs.io).

kakiuwang-ui:
### Description

`FSDPModelManager.__init__` calls `create_device_mesh()` before anything has created the default `torch.distributed` process group. `init_device_mesh` then falls back to a bare `init_process_group()`, which installs PyTorch's built-in watchdog timeout of 30 minutes. For the 1-D mesh RLinf builds, the mesh dimension *is* the default group (`DeviceMesh._init_one_process_group` returns `_get_default_group()` when the dimension spans the whole world), so every FSDP all-gather and reduce-scatter runs on a 30-minute watchdog.

`RLINF_TIMEOUT` — which the scheduler applies to every other inter-worker group and which defaults to 180 minutes — never reaches them, and PyTorch's `default_pg_timeout` is a C++ constant with no environment override, so there is no way to raise it from outside the code.

This PR creates the default group inside `create_device_mesh()` with the scheduler's timeout, so the mesh reuses it:

- `Cluster.get_collective_timeout()` now owns the `RLINF_TIMEOUT` read. `MultiChannelProcessGroup.init()` was doing the same parse inline; both now share it, so the FSDP timeout and the inter-worker timeout cannot drift apart again — which is the exact failure this PR is fixing.
- `create_device_mesh()` initializes the default group when there is none, and logs a warning when one already exists. Without that warning, a user who follows the new FAQ entry, raises `RLINF_TIMEOUT`, and still dies at `Timeout(ms)=1800000` has nothing to go on.
- The shared helper rejects `0` and negative values. They parsed fine before and installed a watchdog that aborts the first collective.
- FAQ entry (EN + ZH) for the `Timeout(ms)=1800000` symptom.

### Motivation and Context

Closes #996.

Two reported ways to hit this, both of which are normal operation rather than a hang:

1. A step where one rank's compute between two collectives exceeds 30 minutes (large gradient accumulation, a slow checkpoint write).
2. Attaching a debugger to one rank — every other rank then dies on the watchdog.

### How has this been tested?

New unit test file `tests/unit_tests/test_fsdp_device_mesh_timeout.py` (9 tests). It builds a real single-rank mesh over gloo and reads the timeout back off the group's backend options, which is the same `_timeout` attribute `DeviceMesh` itself reads when forwarding a timeout to sub-groups.

The tests distinguish a correct implementation from a wrong one — each of these mutations was applied, run, and confirmed to fail on the expected assertion:

- Removing the `init_process_group` call → `test_mesh_group_uses_the_configured_timeout` and `test_mesh_group_defaults_above_the_torch_watchdog` fail with `timedelta(seconds=1800) != timedelta(seconds=5820)`.
- Calling `init_process_group()` without the `timeout` kwarg → the same two fail.
- Dropping the already-initialized warning → `test_existing_process_group_is_left_alone` fails.
- Removing the non-positive check → both parametrized cases of `test_collective_timeout_rejects_unusable_values` fail with `DID NOT RAISE`.

`test_torch_still_installs_the_short_timeout_on_its_own` pins the upstream behaviour that makes the workaround necessary, so if PyTorch ever honours a longer timeout for the implicit group, CI says so.

Full local run (macOS, torch 2.11, CPU): every `tests/unit_tests/test_*.py` file run individually the way `unit-tests.yml` does, plus `pytest --doctest-modules rlinf/scheduler`. Results are identical to clean `main` — the only failures are pre-existing local noise (`test_auto_placement` and `test_auto_resume_checkpoint` fail at collection on missing `torchdata`, `test_distributed_ray_log_collector` has 1 failure, and the 4 `rlinf/scheduler` doctests need a live Ray cluster). `ruff check --preview` and `ruff format --check` are clean.

Not verified on real multi-GPU/NCCL hardware — I do not have a multi-GPU box. The mechanism is backend-independent (the timeout is attached at group creation regardless of backend), but a maintainer confirming on NCCL would be welcome.

**Notes for review**

- **Why no explicit backend.** `init_process_group(timeout=...)` is called without a `backend`, so torch resolves the per-device backend exactly as it did before; only the timeout changes. Hardcoding `"nccl"` (as the issue suggests) would break the Ascend/XPU/MUSA paths that `AcceleratorUtil.get_ccl_backend` supports.
- **Reusing `RLINF_TIMEOUT` instead of a new config key.** `ClusterEnvVar.TIMEOUT` is documented as "timeout for all inter-worker communications" and already defaults to 180 in `Cluster.DEFAULT_SYS_ENV_VAR`; FSDP collectives are inter-worker communication. Happy to move it under `fsdp_config` instead if you would rather keep the two separate — the Megatron path does have its own `megatron.distributed_timeout_minutes`.
- **Adjacent gap left alone.** A 2-D mesh (`hybrid_shard`, #1522) creates its sub-dimension groups through `new_group(timeout=None)`, which falls back to the module-level default rather than the default group's timeout, so those sub-groups would still get 30 minutes. Fixing that needs a `backend_override` with `pg_options._timeout` per dimension and belongs with the hybrid-shard work rather than here. Say the word and I will open a follow-up.
- **Touching `multi_channel_pg.py`.** Only to route it through the shared helper. The behaviour change there is that `RLINF_TIMEOUT=0` or a negative value now raises instead of installing an instantly-aborting watchdog.

### Types of changes

### Checklist:


andylin-hao (approved):

panhe1818:
Hi, the CI pipeline is failing. Could you please take a look and fix this when you get a chance? Thanks. @kakiuwang-ui

kakiuwang-ui:
OK, I'll do that

kakiuwang-ui:
Thanks for the ping — fixed in ef5248b.

The failure was `unit-tests / unit-tests-ascend`, and it was my test rather than the fix. `test_torch_still_installs_the_short_timeout_on_its_own` asserted that the implicitly created default group carries 1800 seconds, which is the NCCL/Gloo default. Ascend HCCL installs 3636 seconds, so the assertion failed on a difference that is not a defect. The eight tests that actually cover the change all passed on Ascend, including the one asserting `RLINF_TIMEOUT` reaches the mesh group.

That test now asserts what holds on every backend: the implicit group does not get the timeout `RLINF_TIMEOUT` asked for, and whatever it does get is below RLinf's 180-minute default. No number is hardcoded any more.

This also turned up a wording problem I have corrected in the same commit — the docstring and both FAQ pages described the 30 minutes as if it were universal. They now name it as the NCCL/Gloo value and give the HCCL one alongside. Worth noting that the underlying bug is real on Ascend too: 3636 s is still well below the 180 minutes RLinf gives its own groups.

Also rebased onto @guozhen1997's merge of `main`, so the branch is current. Locally: `ruff check --preview` and `ruff format --check` clean, and the mutation checks still hold (removing the `init_process_group` call, or dropping its `timeout` kwarg, fails the two timeout tests).

guozhen1997 (approved):

PR #1549, opened 2026-09-08
feat(agent): agentic tokenizer refactor + glm search-r1

sherlockcooper:
### Description

Introduce a pluggable `AgenticTokenizer` abstraction for agent tokenization, with Qwen and GLM-4.7-Flash implementations, enabling search-r1 agent training on GLM models.

- **AgenticTokenizer** base class (`encode`/`decode`/`apply_chat_template`/`get_tool_response_ids`/`content_start_offset`) + `QwenAgenticTokenizer` + `GLMAgenticTokenizer` + `AgenticTokenizerType` enum + `get_agentic_tokenizer` factory.
- **search-r1 agent loop**: use `AgenticTokenizer` for token-level search result truncation + reasoning-content skip via `content_start_offset`.
- **Fix GLM reasoning-end token id**: `convert_tokens_to_ids` without the pipe character (GLM's actual token has no pipe); previously returned `None` → `content_start_offset` always 0 → reasoning skip ineffective.
- **AgenticTokenizer**: `trust_remote_code` passthrough (no longer hardcoded `True`) + edge-case handling.

### Motivation and Context

Search-r1 agent training requires model-specific tokenization (chat template, tool response ids, reasoning-content boundaries). The previous hard-coded Qwen tokenizer broke on GLM-4.7-Flash (different chat template, reasoning-end token, tool call format). This PR abstracts tokenization behind `AgenticTokenizer` with model-specific subclasses, enabling search-r1 on any supported model.

### How has this been tested?

- GLM-4.7-Flash search-r1 agent e2e: tokenizer init + reasoning-content skip (188/216 `content_start>0` after fix) + search truncation + generation + clean exit.
- ruff + ruff-format pass.

### Types of changes

### Checklist:

- [x] My code follows the code style of this project. (`ruff` + `ruff-format` pass on all changed files.)
- [x] All new and existing tests passed. (Manual e2e passed; CI runs via the `run-ci` label.)

### DIFF
--- a/rlinf/agents/searchr1/searchr1_agent_loop.py
+++ b/rlinf/agents/searchr1/searchr1_agent_loop.py
@@ -46,12 +46,9 @@ class Searchr1AgentLoopWorker(MultiAgentLoopWorker):
         assert self.toolcall_parser is not None, (
             "toolcall_parser must be set in searchr1"
         )
-
-        # Inserting tool info requires re-encode token_ids, so the recompute_logprobs must be true.
-        if self.cfg.runner.task_type != "reasoning_eval":
-            assert self.cfg.algorithm.recompute_logprobs, (
-                "search r1 must use recompute_logprobs"
-            )
+        # recompute_logprobs is optional (P2). Base class already sets
+        # return_logprobs = not recompute_logprobs, so when recompute=False,
+        # sglang returns logprobs and truncate_after_stop keeps alignment.
 
     async def pre_process_query(
         self, prompt_ids: list[int], answer: str
@@ -70,7 +67,7 @@ class Searchr1AgentLoopWorker(MultiAgentLoopWorker):
         self, generate_context: dict[str, Any], output: MultiAgentLoopOutput
     ) -> MultiAgentLoopOutput:
         # Compute reward from all LLM-generated tokens (excluding tool responses)
-        final_response_text = self.tokenizer.decode(
+        final_response_text = self.agentic_tokenizer.decode(
             generate_context["all_llm_response_ids"]
         )
         reward_score = compute_score(
@@ -83,7 +80,7 @@ class Searchr1AgentLoopWorker(MultiAgentLoopWorker):
         # Store trajectory-level info for eval
         output.extra_fields["llm_reward"] = reward_score
         output.extra_fields["response_text"] = final_response_text
-        output.extra_fields["prompt_text"] = self.tokenizer.decode(
+        output.extra_fields["prompt_text"] = self.agentic_tokenizer.decode(
             generate_context.get("problem_prompt_ids", [])
         )
         # Per-turn details: each turn's input and output text
@@ -91,8 +88,12 @@ class Searchr1AgentLoopWorker(MultiAgentLoopWorker):
         for single_turn_output in output.single_turn_outputs:
             turns.append(
                 {
-                    "input": self.tokenizer.decode(single_turn_output.prompt_ids),
-                    "output": self.tokenizer.decode(single_turn_output.response_ids),
+                    "input": self.agentic_tokenizer.decode(
+                        single_turn_output.prompt_ids
+                    ),
+                    "output": self.agentic_tokenizer.decode(
+                        single_turn_output.response_ids
+                    ),
                 }
             )
         output.extra_fields["turns"] = turns
@@ -120,24 +121,36 @@ class Searchr1AgentLoopWorker(MultiAgentLoopWorker):
             turn_prompt_ids, sampling_params={"max_new_tokens": max_resp_len}
         )
         llm_response_ids: list[int] = generate_result["output_ids"]
+        llm_response_logprobs = (
+            generate_result["logprobs"] if self.return_logprobs else None
+        )
 
         if len(llm_response_ids) > max_resp_len:
             llm_response_ids = llm_response_ids[:max_resp_len]
-        llm_response_text = self.tokenizer.decode(llm_response_ids)
+            if llm_response_logprobs is not None:
+                llm_response_logprobs = llm_response_logprobs[:max_resp_len]
 
-        # split </search> manually
-        if "</search>" in llm_response_text:
-            llm_response_text = llm_response_text.split("</search>")[0] + "</search>"
-            llm_response_ids = self.tokenizer.encode(llm_response_text)
+        # Truncate at </search> in token space (no re-encode, keeps logprobs aligned).
+        # content_start_offset skips the reasoning block
+        content_start = self.agentic_tokenizer.content_start_offset(llm_response_ids)
+        llm_response_ids, matched = self.agentic_tokenizer.truncate_after_stop(
+            response_ids=llm_response_ids,
+            stop_str="</search>",
+            start=content_start,
+        )
+        if llm_response_logprobs is not None:
+            llm_response_logprobs = llm_response_logprobs[: len(llm_response_ids)]
+        llm_response_text = self.agentic_tokenizer.decode(llm_response_ids)
 
         llm_output = AgentLoopOutput(
             prompt_ids=copy.deepcopy(turn_prompt_ids),
             response_ids=llm_response_ids,
+            response_logprobs=llm_response_logprobs,
         )
         generate_context["all_llm_response_ids"] += llm_response_ids
 
         if len(llm_response_ids) == max_resp_len:
-            return False, None, None, llm_output
+            return False, None, llm_response_text, llm_output
 
         return True, llm_response_ids, llm_response_text, llm_output
 
@@ -167,22 +180,24 @@ class Searchr1AgentLoopWorker(MultiAgentLoopWorker):
             message = {"role": "tool", "content": tool_response.text}
             tool_messages.append(message)
 
-        # Tokenize tool responses
-        tool_response_ids: list[int] = self.tokenizer.encode(
-            tool_messages[0]["content"], add_special_tokens=False
+        # Tokenize tool response via agentic tokenizer (text mode = plain
+        # encode, model-agnostic). Returns the full next-turn prompt.
+        prefix_with_llm = turn_prompt_ids + llm_response_ids
+        next_turn_prompt_ids = self.agentic_tokenizer.get_tool_response_ids(
+            prefix_with_llm, tool_messages, mode="text"
         )
+        appended_len = len(next_turn_prompt_ids) - len(prefix_with_llm)
         max_tool_resp_len = self.max_resp_len - (
             len(turn_prompt_ids) + len(llm_response_ids) - len(problem_prompt_ids)
         )
-        if len(tool_response_ids) > max_tool_resp_len:
+        if appended_len > max_tool_resp_len:
             return False, None
 
-        next_turn_prompt_ids = turn_prompt_ids + llm_response_ids + tool_response_ids
         if self.print_outputs:
             # add anything you want to print
             trace_prints.append(
                 {
-                    "prompt": self.tokenizer.decode(turn_prompt_ids),
+                    "prompt": self.agentic_tokenizer.decode(turn_prompt_ids),
                     "generate": llm_response_text,
                     "tool_resp": tool_messages,
                 }
--- a/rlinf/agents/tokenization/__init__.py
+++ b/rlinf/agents/tokenization/__init__.py
@@ -1,0 +1,31 @@
+# Copyright 2025 The RLinf Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Agentic tokenizer package: model-level tokenization for agent loops."""
+
+from rlinf.agents.tokenization.agentic_tokenizer import (
+    AgenticTokenizer,
+    AgenticTokenizerType,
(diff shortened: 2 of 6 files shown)


rlinf/agents/searchr1/searchr1_agent_loop.py
```diff
@@ -120,24 +121,36 @@ async def generate_llm_response(
...
+        # content_start_offset skips the reasoning block
+        content_start = self.agentic_tokenizer.content_start_offset(llm_response_ids)
+        llm_response_ids, matched = self.agentic_tokenizer.truncate_after_stop(
+            response_ids=llm_response_ids,
+            stop_str="</search>",
+            start=content_start,
+        )
+        if llm_response_logprobs is not None:
+            llm_response_logprobs = llm_response_logprobs[: len(llm_response_ids)]
+        llm_response_text = self.agentic_tokenizer.decode(llm_response_ids)
```
andylin-hao, line 143: **Searches written inside the reasoning block still get executed.**

The reasoning block is skipped when we look for the cut point, but the text handed to the tool-call parser is decoded from the whole response, reasoning included. The parser grabs every `<search>...</search>` it finds, so a query the model only thought about gets run as a real search. On main this couldn't happen because everything after the first `</search>` was dropped.

Suggested fix: decode only the content part (from `content_start` on) for the text that goes to the parser.

rlinf/agents/tokenization/agentic_tokenizer.py
```diff
@@ -0,0 +1,250 @@
...
+        mode="text" encodes the tool response as plain text (search-r1 style,
+        model-agnostic). mode="native" uses the model-native chat-template
+        boundary merge (subclass-specific).
+        """
+        if mode == "native":
+            return self.build_native_tool_response(prefix_ids, tool_messages)
+        return list(prefix_ids) + self.build_text_tool_response(tool_messages)
+
+    def build_text_tool_response(self, tool_messages: list[dict]) -> list[int]:
+        content = "".join(m["content"] for m in tool_messages)
```
andylin-hao, line 80: **All tool results are now concatenated; main only used the first one.**

Together with the reasoning issue above, one turn can now inject several search results into the next prompt. That changes what the model sees during training and isn't mentioned in the PR description.

Suggested fix: keep only the first tool message as before, or state the change explicitly and add a test for it.

rlinf/agents/searchr1/searchr1_agent_loop.py
```diff
@@ -46,12 +46,9 @@ def __init__(
         assert self.toolcall_parser is not None, (
             "toolcall_parser must be set in searchr1"
         )
-
-        # Inserting tool info requires re-encode token_ids, so the recompute_logprobs must be true.
-        if self.cfg.runner.task_type != "reasoning_eval":
-            assert self.cfg.algorithm.recompute_logprobs, (
-                "search r1 must use recompute_logprobs"
-            )
+        # recompute_logprobs is optional (P2). Base class already sets
```
andylin-hao, line 49: **This turns on a training path that hasn't been tested.**

Dropping the assert means search-r1 can now train with the rollout's own logprobs instead of recomputing them. The PR description doesn't say which setting the GLM run used, and no test covers `recompute_logprobs: False` for search-r1.

Suggested fix: either keep the assert and move this to a separate PR, or add a search-r1 e2e config with `recompute_logprobs: False`. Please also replace "(P2)" with a plain explanation, since readers won't know what it refers to.

rlinf/agents/searchr1/searchr1_agent_loop.py
```diff
@@ -120,24 +121,36 @@ async def generate_llm_response(
             turn_prompt_ids, sampling_params={"max_new_tokens": max_resp_len}
         )
         llm_response_ids: list[int] = generate_result["output_ids"]
+        llm_response_logprobs = (
```
andylin-hao, line 124: **This crashes if the rollout backend doesn't return logprobs.**

The sglang worker only adds the `logprobs` key when it was started with logprobs enabled, and the online router returns `None`. In either case this line, or the length check in `get_rollout_result`, fails with an unhelpful KeyError or TypeError.

Suggested fix: check that logprobs are present and raise a clear error that names the config field to set.

rlinf/workers/agent/agent_loop.py
```diff
@@ -106,7 +106,13 @@ def __init__(
                 "agent loop worker must be MultiAgentLoopWorker if is_dynamic_rollout_batch is True"
             )
 
-        self.tokenizer = AutoTokenizer.from_pretrained(cfg.rollout.model.model_path)
+        self.agentic_tokenizer = get_agentic_tokenizer(
+            cfg.rollout.model.model_path,
+            cfg.agentloop.get("agentic_tokenizer", "default"),
+            cfg.agentloop.get("chat_template_kwargs", None),
+            cfg.rollout.model.get("trust_remote_code", False),
```
andylin-hao, line 113: **This config key doesn't exist, so GLM can never get `trust_remote_code=True`.**

No config on main defines `rollout.model.trust_remote_code`; agent configs use `actor.tokenizer.trust_remote_code`. Because this lookup always yields `False` and passes it explicitly, the `True` default in `GLMAgenticTokenizer.__init__` never takes effect.

Suggested fix: read the existing `actor.tokenizer.trust_remote_code` key and drop the unused default from the GLM subclass.

rlinf/agents/tokenization/agentic_tokenizer.py
```diff
@@ -0,0 +1,250 @@
...
+        """Convert token string to id, raising if not in vocab."""
+        tid = self.tokenizer.convert_tokens_to_ids(token_str)
+        if tid is None or tid == self.tokenizer.unk_token_id:
+            raise ValueError(f"{token_str!r} not in {type(self).__name__} vocab")
+        return tid
+
+    def encode(self, text: str, add_special_tokens: bool = True) -> list[int]:
+        return self.tokenizer.encode(text, add_special_tokens=add_special_tokens)
+
+    def decode(self, ids: list[int], skip_special_tokens: bool = True) -> str:
```
andylin-hao, line 56: **Decoding now strips special tokens, which main didn't do.**

On main, `decode` kept special tokens. For example, Qwen2.5 used to decode `<answer>x</answer><|im_end|>` with the `<|im_end|>` included; with this change it's gone. That affects the reward text, the tool parser input, and the eval logs. The docstring above says the new path is "bit-identical" to the old one, which isn't true. The switch to `hf_tokenizer` also rewrites the pad token.

Suggested fix: default `skip_special_tokens` to `False` to match main, or drop the "bit-identical" claim and list the behavior change in the PR description.

rlinf/agents/tokenization/agentic_tokenizer.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+        return 0
+
+    def truncate_after_stop(
+        self, response_ids: list[int], stop_str: str, start: int
+    ) -> tuple[list[int], bool]:
+        """Truncate response_ids at the first token where stop_str appears in
+        the content region (response_ids[start:]).
+
+        Returns (truncated_ids, matched). Decoded text length is monotonic in
```
andylin-hao, line 103: **Truncation depends on an assumption that isn't guaranteed.**

The binary search assumes decoded text only grows as tokens are added. If that ever fails, the assert below crashes the rollout. I tried Chinese, emoji, and CJK extension characters with the Qwen2.5 tokenizer and all of them worked, so this is a hardening suggestion, not a confirmed bug.

Suggested fix: binary-search on `stop_str in decode(ids[start:mid])` directly. That condition is always monotonic, and it's exactly what the assert checks.

rlinf/agents/tokenization/agentic_tokenizer.py
```diff
@@ -0,0 +1,250 @@
...
+        # clear_thinking=False keeps reasoning blocks in the token stream.
+        merged = {"clear_thinking": False, **(chat_template_kwargs or {})}
+        super().__init__(model_path, merged, trust_remote_code=trust_remote_code)
+        # Resolve special-token ids at load time; fail loudly if missing.
+        self.observation_token_id = self.require_token_id("<|observation|>")
+        self.user_token_id = self.require_token_id("<|user|>")
+        self.ambiguous_token_ids = {self.observation_token_id, self.user_token_id}
+        self.reasoning_end_token_id = self.require_token_id("</think>")
+
+    def render_incremental(self, tool_messages: list[dict]) -> list[int]:
```
andylin-hao, line 164: **About half of this class has no caller in this PR.**

Native mode, `render_incremental`, the observation and user token ids, `stop_token_ids_extra`, the parser name attributes, and `apply_chat_template` are all unused. Code that never runs is easy to get wrong without anyone noticing. For example, `render_incremental` ignores `self.chat_template_kwargs`, so the `clear_thinking=False` set in `__init__` doesn't apply here.

Suggested fix: keep only what search-r1 uses (decode, truncation, content offset, text tool response). Add native mode together with the agent loop that needs it.

andylin-hao (commented):
Thanks for this PR. Truncating in token space instead of decoding and re-encoding is a real improvement, and it keeps logprobs aligned. Inline comments are ordered roughly by severity. The first three items are the ones that should be fixed before merge:

1. Search queries inside the reasoning block are still executed.
2. Turning off `recompute_logprobs` for search-r1 isn't covered by any test.
3. `trust_remote_code` is read from a config key that doesn't exist.

A few items that don't belong to a specific line:

- **Tests:** `truncate_after_stop` (multibyte text, a stop string inside the reasoning block), the GLM `content_start_offset`, and the factory's error path can all be unit-tested with a small local tokenizer.
- **Docs and example:** the title says GLM search-r1, but there's no GLM config under `examples/agent/searchr1/config/`, and the new keys (`agentloop.agentic_tokenizer`, `data.apply_chat_template_kwargs` for search-r1) aren't documented. Please add the config and update the search-r1 example page in both EN and ZH.
- **Commits:** the second commit's author is `RLinf Dev <dev@rlinf.local>`, but it's signed off by you, which will likely fail DCO. Please re-author it.
- **Rebase:** the branch is based on an older main, so please rebase before the next round.

rlinf/agents/tokenization/agentic_tokenizer.py
```diff
@@ -0,0 +1,250 @@
...
+        return self.tokenizer.encode(text, add_special_tokens=add_special_tokens)
+
+    def decode(self, ids: list[int], skip_special_tokens: bool = True) -> str:
+        return self.tokenizer.decode(ids, skip_special_tokens=skip_special_tokens)
+
+    def apply_chat_template(self, messages: list[dict], **kwargs: Any):
+        merged = {**self.chat_template_kwargs, **kwargs}
+        return self.tokenizer.apply_chat_template(messages, **merged)
+
+    def get_tool_response_ids(
```
andylin-hao, line 63: **This shares a name with an existing worker method but returns something different.**

`AgentLoopWorker.get_tool_response_ids(tool_messages)` already exists (wideseek_r1 uses it) and returns only the new tokens. This method returns the whole next-turn prompt. Having two methods with the same name and different results is a trap for the next person, and `render_incremental` duplicates the worker's template-diff logic.

Suggested fix: either move the worker's version in here and switch wideseek_r1 to it, or give this method a different name, such as `append_tool_response`.

rlinf/workers/agent/agent_loop.py
```diff
@@ -106,7 +106,13 @@ def __init__(
                 "agent loop worker must be MultiAgentLoopWorker if is_dynamic_rollout_batch is True"
             )
 
-        self.tokenizer = AutoTokenizer.from_pretrained(cfg.rollout.model.model_path)
+        self.agentic_tokenizer = get_agentic_tokenizer(
+            cfg.rollout.model.model_path,
+            cfg.agentloop.get("agentic_tokenizer", "default"),
+            cfg.agentloop.get("chat_template_kwargs", None),
```
andylin-hao, line 112: **Now there are two config keys for chat-template arguments.**

The dataset reads `data.apply_chat_template_kwargs`, and the tokenizer reads `agentloop.chat_template_kwargs`, but search-r1 only honors the first. Someone turning off GLM thinking could easily set the key that does nothing.

Suggested fix: use only `data.apply_chat_template_kwargs`, which rstar2 already uses.

rlinf/agents/tokenization/agentic_tokenizer.py
```diff
@@ -0,0 +1,250 @@
...
+        assert stop_str in self.decode(result), (
+            f"truncate_after_stop landed before {stop_str!r}"
+        )
+        return result, True
+
+
+class QwenAgenticTokenizer(AgenticTokenizer):
+    """Qwen2.5/3 reference subclass. Inherits base text-mode path."""
+
+    reasoning_parser = "qwen3"
```
andylin-hao, line 130: **Qwen3 thinking models don't get the reasoning skip.**

This class sets `reasoning_parser = "qwen3"`, but `content_start_offset` still returns 0, so a Qwen3 model in thinking mode keeps the old behavior.

Suggested fix: remove the attribute, or implement the offset using `</think>` in the base class so every model that emits it gets the skip.

rlinf/runners/agent_runner.py
```diff
@@ -130,6 +130,7 @@ def init_rollout_workers(self):
             if (
                 self.cfg.actor.training_backend == "megatron"
                 and self.cfg.actor.megatron.use_hf_ckpt
+                and not getattr(self.cfg.actor.megatron, "mbridge", False)
```
andylin-hao, line 133: **This mbridge fix is unrelated to the PR, and only one of three runners gets it.**

`reasoning_runner.py` already has the same check. `agentlightning_runner.py` and `coding_online_rl_runner.py` still convert HF checkpoints to Megatron when mbridge is enabled.

Suggested fix: move this to a separate `fix(runner)` PR that updates all three runners.

rlinf/agents/tokenization/__init__.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2025 The RLinf Authors.
```
andylin-hao, line 1: New files should use the current year: `Copyright 2026 The RLinf Authors.` (Same for `agentic_tokenizer.py`.)

PR #1551, opened 2026-09-08
feat: support new Dagger in Robotwin

renji555:
fix:support new Dagger in Robotwin


examples/embodiment/config/robotwin_adjust_bottle_dagger_openpi.yaml
```diff
@@ -48,6 +48,19 @@ algorithm:
     beta_decay: 0.99
     # False enables classic DAgger: student-acted data is re-labeled by expert.
     only_save_expert: False
+    # Store one observation/action pair per environment step. The rolling
+    # LeRobot dataset reconstructs action chunks for OpenPI training.
+    online_lerobot:
+      enabled: True
+      only_success: False
```
zanghz21, line 55: I think only_success should be true?

panhe1818:
Hi, @renji555, we have one comment above. Could you please take a look when you get a chance? Thanks.

examples/embodiment/config/robotwin_adjust_bottle_dagger_openpi.yaml
renji555: Yeah, I have already fixed it

zanghz21 (approved):

renji555:
This is the corresponding picture, first picture is async, second is sync
截屏2026-09-16 20 58 48
7821789399537_ pic

rlinf/envs/sim/robotwin/robotwin_env.py
```diff
@@ -329,13 +329,14 @@ def chunk_step(self, chunk_actions):
         obs_list = []
         infos_list = []
 
-        raw_obs, step_reward, terminations, truncations, info_list = self.venv.step(
-            chunk_actions
+        raw_obs_list, step_reward, terminations, truncations, info_list = (
+            self.venv.step_with_full_obs(chunk_actions)
```
Lin-xs, line 333: Is `step_with_full_obs` available in the RoboTwin that RLinf uses now?

renji555: Yeah, it's available and the pr is: https://github.com/RoboTwin-Platform/RoboTwin/pull/499

Lin-xs: Got it. So PR #1551 depends on the RoboTwin PR and should be merged after it.

renji555: Yeah, you can check the RoboTwin PR first

Lin-xs (approved):

andylin-hao (commented):
Here are some comments

rlinf/workers/env/env_worker.py
```diff
@@ -1125,9 +1151,14 @@ async def _run_interact_once(
                     actions = self._recv_actions(input_channel, stage_id)
                     self.smooth_intervene.remember_actions(stage_id, actions)
 
+                    curr_obs = env_outputs[stage_id].obs
                     env_output, env_info, chunk_step_data = self.env_interact_step(
                         actions, stage_id
                     )
+                    if chunk_step_data["valid_action_mask"] is not None:
```
andylin-hao, line 1158: This breaks an existing test. `test_env_closes_reward_stream_with_final_chunk` in `tests/unit_tests/test_data.py` passes on the merge base and fails on this branch with `KeyError: 'valid_action_mask'`, because its stubbed payload has no such key. Any other producer of `chunk_step_data` that doesn't set the key would fail the same way. Using `.get("valid_action_mask")` fixes the crash, though moving this logic out of the loop (next comment) would remove the lookup altogether.

rlinf/workers/env/env_worker.py
```diff
@@ -1125,9 +1151,14 @@ async def _run_interact_once(
...
+                        chunk_step_data["obs_list"] = [
+                            curr_obs, *chunk_step_data["obs_list"][:-1]
```
andylin-hao, line 1160: With this change, `obs_list` holds pre-action observations when a mask is present and post-action observations otherwise. The accumulator's `pending_observation` path still assumes the second meaning. Nothing in the schema says so: `LeRobotChunk.valid_action_mask` is documented as "executed action slots", and `has_final_observation` is quietly switched off in the masked branch. The next env that starts reporting `executed_action_count` will get its observation/action pairing changed without anyone noticing.

I'd rather make the convention explicit. Either have the env return pre-action observations itself, or add a documented field on the chunk that says which convention `obs_list` uses. At minimum, the field comment on `valid_action_mask` should state that its presence means `obs_list` holds pre-action observations. Rewriting the payload in the middle of the rollout loop is also easy to miss when reading the worker.

rlinf/workers/env/env_worker.py
```diff
@@ -394,6 +394,27 @@ async def _maybe_wait_env_delay(self, stage_id: int) -> None:
             return
         await env.wait_delay()
 
+    @staticmethod
+    def _valid_action_mask_from_infos(
+        infos: Any,
+        chunk_size: int,
+    ) -> torch.Tensor | None:
+        executed_counts = infos.pop("executed_action_count", None)
```
andylin-hao, line 402: A few issues with this helper:

- It `pop`s from the env's `infos` on every train step, even when online LeRobot is disabled, so it changes what everything downstream of `env_output.env_infos` sees.
- It crashes if `infos` is `None`.
- The call site passes `chunk_size=len(obs_list)`, which gives the number of keys if an env returns a single obs dict instead of a list.

I'd only parse the mask when `self.enable_online_lerobot` is set, and take the width from `self.model_cfg.num_action_chunks`. A short docstring on what `executed_action_count` means and where it may appear (top level or `final_info`) would help the next env author.

rlinf/envs/sim/robotwin/robotwin_env.py
andylin-hao, line 333: This makes every RoboTwin run render an observation after each action, not just DAgger runs. On the RoboTwin side, `step_with_full_obs` calls `gen_sparse_reward_data(collect_obs=True)`, which calls `get_obs()` per action. PPO/GRPO on RoboTwin never uses the intermediate frames, but would now pay for C−1 extra renders per chunk.

Could you call `step_with_full_obs` only when online LeRobot collection is enabled and keep `venv.step` otherwise? Please also post env throughput with and without the change. One more thing: anyone with a RoboTwin checkout older than 2026-09-08 will hit an `AttributeError` here, so the RoboTwin docs (EN and ZH) should tell users to pull the latest `RLinf_support`.

rlinf/data/schema/embodied_types.py
```diff
@@ -1940,13 +1979,17 @@ def merge_steps(values: list[list[Any]]) -> list[Any]:
...
 
-    return {
+    merged = {
         "chunk_actions": merge_batch_values([value["chunk_actions"] for value in data]),
         "obs_list": merge_steps([value["obs_list"] for value in data]),
         "terminations": merge_batch_values([value["terminations"] for value in data]),
         "truncations": merge_batch_values([value["truncations"] for value in data]),
         "infos_list": merge_steps([value["infos_list"] for value in data]),
+        "valid_action_mask": merge_batch_values(
+            [value["valid_action_mask"] for value in data]
```
andylin-hao, line 1989: `split_episode_data` reads the mask with `data.get(...)`, but this uses `value["valid_action_mask"]`, so merging payloads that don't carry the key raises `KeyError`. Both sides should treat a missing key the same way. Separately, the `chunks = ...; return chunks` and `merged = ...; return merged` rewrites only add a temporary variable. Returning the expressions directly, as before, keeps the diff smaller.

rlinf/data/schema/embodied_types.py
```diff
@@ -1231,16 +1237,33 @@ def step_count(self) -> int:
...
-        terminated = self._flag_at(self.terminations, step_index, env_index)
-        truncated = self._flag_at(self.truncations, step_index, env_index)
+        if self.valid_action_mask is None:
+            terminated = self._flag_at(self.terminations, step_index, env_index)
+            truncated = self._flag_at(self.truncations, step_index, env_index)
+        else:
+            valid_length = int(self.valid_action_mask[env_index].sum())
+            is_last_valid = step_index == valid_length - 1
+            terminated = is_last_valid and bool(
+                np.asarray(self.terminations[env_index]).any()
```
andylin-hao, line 1256: `self.terminations[env_index]` only means "this env's flags across the chunk" when terminations are `[B, C]`. The field comment allows `[B]` or a scalar too. With `[B]` this reads a single flag, which happens to work, but a scalar fails on indexing. Since LIBERO and RoboTwin both produce `[B, C]`, the simplest fix is to check in `from_data` that terminations and truncations are `[B, C]` whenever a mask is given.

rlinf/models/embodiment/openpi/openpi_action_model.py
```diff
@@ -693,6 +701,10 @@ def prepare_lerobot_sft_batch(self, batch):
             obs_dict["observation/extra_view_image"] = torch.stack(
                 obs_dict["observation/extra_view_image"], dim=1
             )
+        if merge_wrist:
+            obs_dict["observation/wrist_image"] = torch.stack(
+                [batch[key] for key in _wrist_merge_keys], dim=1
+            ).permute(0, 1, 3, 4, 2)
```
andylin-hao, line 707: The merged wrist images get permuted from CHW to HWC here, but the `extra_view_image` merge just above and the single `image`/`wrist_image` keys aren't permuted. All of them then go through the same `input_transform(..., transpose=False)`. Either the replay buffer stores wrist images in a different layout from the other cameras, which would be worth a comment explaining why, or one of these paths feeds the model the wrong layout. Please make the layouts consistent, and add a test that runs a two-wrist batch through `prepare_lerobot_sft_batch`. This code has also moved to `tasks/dagger.py` on main, as noted in the summary.

rlinf/data/schema/embodied_types.py
renji555: OK, I will add a check to avoid it.

rlinf/data/schema/embodied_types.py
renji555: OK, I have returned the expression

rlinf/envs/sim/robotwin/robotwin_env.py
renji555: I will check use lerobot or not to determine choose venv.step or venv.step_with_full_obs

rlinf/workers/env/env_worker.py
renji555: OK, I will use get instead of pop to avoid downstream can't see.And I will check infos. Finally I will use num_action_chunks to calculate chunk_size. And all these changes will only be used when enable_online_lerobot is True

rlinf/workers/env/env_worker.py
renji555: OK

rlinf/workers/env/env_worker.py
renji555: I have moved pre-action observation alignment into env_interact_step and documenting that masked chunks use action-aligned pre-action observations, while EnvOutput.obs remains the final post-action observation.

PR #1556, opened 2026-09-10
fix(envs): add optional methods to ManiskillOffloadEnv

### LINKED ISSUE #1555: ManiSkill offload env crashes at the first interaction step: "Method 'wait_delay' not found"
### Environment

- RLinf main @ `f1b2ee15` (verified still present on current main `4e07a1dd`; none of the involved files changed in between)
- Config: `examples/embodiment/config/maniskill_ppo_openvlaoft_quickstart.yaml` as published, only `model_path` localized, single node
- `env.train.enable_offload: true` — the published default
- Python 3.11, PyTorch 2.x, 2x NVIDIA A100

### What happens

Training dies at the start of the first rollout epoch. The env worker awaits `wait_delay` on the offload env, the call is forwarded to the simulator process, and the simulator's dispatch rejects it:

```python
  File "rlinf/workers/env/env_worker.py", line 1127, in _run_interact_once
    await self._maybe_wait_env_delay(stage_id)
  File "rlinf/workers/env/env_worker.py", line 483, in _maybe_wait_env_delay
    await env.wait_delay()
  File "rlinf/envs/sim/maniskill/maniskill_offload_env.py", line 454, in method_proxy
    return self._rpc(name, args=args, kwargs=kwargs)
RuntimeError: Method 'wait_delay' not found
```

### Root cause

`ManiskillOffloadEnv` does not implement env APIs itself. `__getattr__` resolves any public name to an RPC `method_proxy`, and its `get_wrapper_attr` — added so wrapper lookups are not forwarded to the worker process — still falls back to `self.__getattr__(name)` for names it does not recognize, handing out a proxy. As a result `get_env_attr` never returns `None` for this env, and the existence guard in `_maybe_wait_env_delay`

```python
if get_env_attr(env, "wait_delay") is None:
    return
await env.wait_delay()
```

never triggers. The worker awaits the proxy, `_maniskill_worker_main`'s dispatch has no `wait_delay` method, and the returned error kills the job. The guard works for in-process gym envs because a missing attribute raises `AttributeError`; a catch-all proxy cannot produce that answer.

A minimal reproducer needs neither a GPU nor a simulator process:

```python
from rlinf.envs.sim.maniskill.maniskill_offload_env import ManiskillOffloadEnv
from rlinf.envs.utils import get_env_attr

env = object.__new__(ManiskillOffloadEnv)  # no worker process started
for name in ("wait_delay", "insert_delay_metrics", "get_hold_actions", "not_a_real_method"):
    print(name, get_env_attr(env, name))  # every name resolves to a callable
```

### Scope

- The same conflict exists at the `insert_delay_metrics` probe on the interaction path.
- `rlinf/utils/env_helpers/smooth_intervene.py` probes `get_hold_actions`, which only real-robot envs implement today — latent, reachable only with smooth-intervene configs.
- Deterministic, not a race.
- The e2e suite does not cover this path: `maniskill_async_ppo_openvlaoft.yaml` and the `delay_sampler` e2e both set `enable_offload: false`, so CI never runs ManiSkill VLA against the offload proxy. The published quickstart does.

### Fix

Two directions:

- (a) Implement the optional protocol methods locally on `ManiskillOffloadEnv`: `async def wait_delay()` as a no-op (delays are sampled by `InsertDelay` wrappers, never by the sim process) and `insert_delay_metrics()` returning an empty tensor (`compute_delay_stats` already handles empty input). 12 lines, no shared-code changes.
- (b) Make existence probes proxy-proof, e.g. by deciding on wrapper type rather than attribute presence.

Happy to send a PR for (a) — it is what we are running; the quickstart then trains stably past the point where it previously crashed.

puzhiyuan:
### Description

`ManiskillOffloadEnv` resolves every unknown public attribute to an RPC method proxy, so existence probes such as `get_env_attr(env, "wait_delay") is None` never return `None` for it. The env worker's interaction path then awaits `wait_delay` on the simulator process, whose dispatch has no such method, and training dies at the first rollout epoch with `Method 'wait_delay' not found`.

This change implements the two optional protocol methods locally on the proxy:

- `async def wait_delay()`: a no-op. Delays are sampled by `InsertDelay` wrappers around the env, never by the simulator process, so there is nothing to wait here.
- `insert_delay_metrics()`: returns an empty float32 tensor; `compute_delay_stats` already handles empty input and reports zeros.

Existence probes now resolve to real methods, the same semantics every in-process gym env already provides. A unit test pins the contract.

### Motivation and Context

Fixes #1555. Any ManiSkill run with `enable_offload: true` (the published quickstart default) crashed at the first interaction step. The issue has the full analysis, a no-GPU reproducer, and why the e2e suite does not cover this combination (`maniskill_async_ppo_openvlaoft.yaml` sets `enable_offload: false`).

### How has this been tested?

- New unit test `tests/unit_tests/test_envs.py::test_optional_protocol_methods_resolve_locally` (auto-skips without `mani_skill`); runs green locally together with `test_utils.py` (16 passed). The ray-based cluster/placement suites need more local resources and are left to CI.
- A minimal reproducer on a clean checkout confirmed `get_env_attr` resolved arbitrary names — including a fabricated one — to RPC proxies before the change.
- End-to-end: the published `maniskill_ppo_openvlaoft_quickstart` on 2x A100 trains stably past the previous crash point (9+ rollout epochs, ~169 s/step, rewards/success/`time/interact_delay` logged).
- `ruff check` and `ruff format --check` pass on both changed files.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/envs/sim/maniskill/maniskill_offload_env.py
+++ b/rlinf/envs/sim/maniskill/maniskill_offload_env.py
@@ -430,6 +430,18 @@ class ManiskillOffloadEnv(EnvOffloadMixin):
             return getattr(self, name)
         return self.__getattr__(name)
 
+    async def wait_delay(self):
+        """No-op: delays are sampled by ``InsertDelay`` wrappers, never here.
+
+        Defined locally so existence probes like ``get_env_attr`` resolve to a
+        real method instead of an RPC proxy for a method the worker process
+        does not have.
+        """
+
+    def insert_delay_metrics(self):
+        """Report no recorded delays; samples live in ``InsertDelay`` wrappers."""
+        return torch.tensor([], dtype=torch.float32)
+
     def _force_shutdown(self):
         if self.command_queue is not None:
             self.command_queue.close()
--- a/tests/unit_tests/test_envs.py
+++ b/tests/unit_tests/test_envs.py
@@ -1,0 +1,48 @@
+# Copyright 2026 The RLinf Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Environment attribute contracts seen by the env worker."""
+
+import asyncio
+
+import pytest
+
+pytest.importorskip("mani_skill")
+
+import torch  # noqa: E402
+
+from rlinf.envs.sim.maniskill.maniskill_offload_env import (  # noqa: E402
+    ManiskillOffloadEnv,
+)
+from rlinf.envs.utils import get_env_attr  # noqa: E402
+
+
+def _bare_offload_env() -> ManiskillOffloadEnv:
+    """Build the proxy without starting its simulator worker process."""
+    return object.__new__(ManiskillOffloadEnv)
+
+
+def test_optional_protocol_methods_resolve_locally():
+    """get_env_attr probes must see the local protocol methods, not RPC proxies."""
+    env = _bare_offload_env()
+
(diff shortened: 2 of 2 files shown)


andylin-hao (approved):

guozhen1997 (approved):

zhanghengnb:
taishuaile

PR #1558, opened 2026-09-11
fix(fsdp): reduce the gradient norm over every shard

andylin-hao:
### Description

`create_device_mesh` builds a 1-D mesh whose only dimension is named `fsdp`, so the `"ddp" in mesh_dim_names` guard on `_dp_group` never matched and the group stayed `None` for every FSDP and FSDP2 run. `get_grad_norm` skipped its `all_reduce` as a result and returned the norm of whichever shard the local rank held, which `clip_grad_norm_` then clipped against.

The selection now lives in a named `gradient_reduction_group(device_mesh)` rather than inline in the manager, because it carries a constraint that is easy to get wrong: the group has to span the sharding dimension and nothing else. Hybrid sharding replicates across `ddp`, so reducing over that dimension would count each shard once per replica.

### Motivation and Context

The reported norm was roughly `1/sqrt(world_size)` of the real one and shrank as the job grew, so gradient clipping silently loosened with scale. A two-GPU measurement from #1498 puts numbers on it: without the reduction `grad_norm` came out 50.6862 against a reference DDP value of 71.9728 (≈ 1/√2); with it, 72.0690.

**This changes training behaviour for every existing FSDP config.** Clipping now bites at the configured threshold, so runs that were effectively unclipped will now be clipped even though nothing in their YAML moved. That is the intended correction, but it is worth knowing before rerunning a baseline.

Overlaps with #1522, which reaches the same `_dp_group = mesh["fsdp"].get_group()` while adding the 2-D hybrid mesh. The two agree on the value; whichever lands second should take the other's version of the line.

### How has this been tested?

- `tests/unit_tests/test_fsdp_device_mesh_timeout.py` gains a case pinning which mesh dimension the group comes from, including the loud failure when no sharding dimension exists — silently reducing over the wrong one is how this went unnoticed.
- Full file: 10 passed. Note this file is order-sensitive on a busy machine; its real `init_process_group` tests race for ephemeral ports and fail intermittently on `main` as well.
- The cross-rank reduction itself needs more than one rank and is **not** covered by a unit test. A 2-rank gloo spawn test does cover it but costs ~60s and polluted the other tests in the file, so it was dropped in favour of the two-GPU measurement above.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
+++ b/rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
@@ -43,6 +43,7 @@ from rlinf.hybrid_engines.fsdp.strategy.base import FSDPStrategyBase
 from rlinf.hybrid_engines.fsdp.utils import (
     create_device_mesh,
     get_lr_scheduler,
+    gradient_reduction_group,
 )
 from rlinf.models.tokenization.hf import hf_tokenizer
 from rlinf.scheduler import Worker
@@ -97,11 +98,7 @@ class FSDPModelManager:
             self.tokenizer = hf_tokenizer(cfg.tokenizer.tokenizer_model)
 
         self._device_mesh = create_device_mesh(world_size)
-        self._dp_group = (
-            self._device_mesh["ddp"].get_group()
-            if "ddp" in self._device_mesh.mesh_dim_names
-            else None
-        )
+        self._dp_group = gradient_reduction_group(self._device_mesh)
 
         self._strategy = FSDPStrategyBase.create(
             self._cfg, world_size, self._dp_group, self._logger
--- a/rlinf/hybrid_engines/fsdp/utils.py
+++ b/rlinf/hybrid_engines/fsdp/utils.py
@@ -97,6 +97,23 @@ def create_device_mesh(world_size: int) -> DeviceMesh:
     )
 
 
+def gradient_reduction_group(device_mesh: DeviceMesh) -> torch.distributed.ProcessGroup:
+    """Return the group a gradient's shards are spread over.
+
+    :func:`get_grad_norm` sums per-rank shard norms over this group, so it has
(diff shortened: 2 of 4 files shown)


rlinf/hybrid_engines/fsdp/fsdp_model_manager.py
```diff
@@ -97,11 +98,7 @@ def __init__(self, cfg: DictConfig, world_size: int, rank: int) -> None:
             self.tokenizer = hf_tokenizer(cfg.tokenizer.tokenizer_model)
 
         self._device_mesh = create_device_mesh(world_size)
-        self._dp_group = (
-            self._device_mesh["ddp"].get_group()
-            if "ddp" in self._device_mesh.mesh_dim_names
-            else None
-        )
+        self._dp_group = gradient_reduction_group(self._device_mesh)
```
chenchaoxu7575: One edge case I’d like to confirm: is ignored_module_classes intended to contain only frozen parameters? 
If an ignored parameter is trainable, its replicated gradient will be counted once per rank. For example, with two ranks, sharded gradients of 3 and 4 plus a replicated gradient of 12, this gives 17.69 `sqrt((3² + 12²) + (4² + 12²))` instead of the expected 13 `sqrt(3² + 4² + 12²)`.
I see that FSDP1 handles sharded and non-sharded contributions separately. Should we follow this?
https://github.com/RLinf/RLinf/blob/db66ac56d1aa4a9c8441c4026e4212b21811970d/rlinf/hybrid_engines/fsdp/strategy/fsdp.py#L400-L452

myshrry:
I'm running into a similar issue. Could you please let me know why this PR was closed? Has there been an official fix for it? Thank you! @andylin-hao

guozhen1997 (approved):

Lin-xs (approved):

PR #1561, opened 2026-09-11
feat(robotics): add arm moves and robots built from config

andylin-hao:
### Description

Adds the robot-layer operations a robot-independent task will need, moves each real-world env's config-to-robot mapping onto its robot class, and pins the schema every registered task shows its policy.

- `Arm.move_to(pose, duration=, rate_hz=, clear_errors=)` travels to a tool pose through evenly spaced `tcp_pose` commands (linear position, slerp rotation). It is `FrankaEnv._interpolate_move` moved onto the arm, so a remotely hosted arm now runs the whole move on its own node.
- `Arm.DOF` states the joint count before anything connects: Franka 7, Piper 6, SO-101 5, GimArm 6.
- `MethodEndEffector.open()` / `close()` go through the host, and `is_open` reads a host state field when one is named. GimArm's gripper used to report open whatever it held.
- `Robot.from_config(config, cameras=, env_idx=, node_rank=, worker_rank=)` on Franka, Piper, SO-101 and GimArm replaces the mapping each env wrote by hand in `_setup_hardware`.
- `normalize` and `quat_slerp` move from `rlinf.envs.real.utils.pose` to `rlinf.robotics.pose`, so the robot layer does not import from `envs`.
- `realworld_gim_arm_peg_insertion.yaml` gets back the `env_type: real` line #1481 dropped; without it the GimArm mock e2e stopped at config validation.
- Docs (EN and ZH): `concepts/robotics.rst` covers the new arm and gripper operations; `extending/new_robot.rst` shows `from_config()` where it used to tell readers to write that translation in shared setup code.

### Motivation and Context

First of the steps that let a real-world task (peg insertion, bin relocation, reach) run on any robot meeting its requirements instead of subclassing one robot's env. Later steps move each env onto a shared task env; the schema table added here is what they must leave unchanged, row by row, unless a change is named.

The one behaviour difference: `FrankaEnv` no longer reads its cameras before an interpolated move, only the arm. The read after the move is unchanged.

### How has this been tested?

- `pytest tests/unit_tests/test_robotics.py`: 282 passed. New: `move_to` command count, pacing, endpoint and slerp quarter-point; refusal on joint-only arms; `DOF`; gripper open/close/`is_open` on the GimArm driver against its fake SDK; `from_config` forwarding for all four robots with and without `controller_node_rank`.
- `pytest tests/unit_tests/test_real_env.py`: 225 passed. New: `test_a_task_keeps_the_schema_its_policies_were_trained_on` for all 12 registered ids, generated from `main` and compared equal to this branch; `test_franka_step_moves_the_arm_by_the_scaled_clipped_delta` on fakes.
- Each new test was checked against a one-line mutation that breaks it (no fault clear, nlerp instead of slerp, dropped config field, wrong controller node, no `open_field`, no `Quat2EulerWrapper`, no clip, no scale).
- Mock e2e: `realworld_mock_sac_cnn`, `piper_mock_sac_mlp_reach`, `so101_mock_sac_mlp_reach`, `gim_arm_mock_sac_cnn` all reach step 2/2.
- `build_docs.py`: EN and ZH build with 0 warnings. `pre-commit run` clean.

The GimArm mock e2e job has been green on `main` without training anything: `Cluster._shutdown_ray_at_exit` calls `os._exit(0)` at exit unless `sys.excepthook` saw a failure, and Hydra catches the exception and calls `sys.exit(1)` before the hook can see it. That masking is left for a separate PR because fixing it may turn other jobs red.

### Types of changes

### Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/source-en/rst_source/concepts/robotics.rst
+++ b/docs/source-en/rst_source/concepts/robotics.rst
@@ -56,7 +56,10 @@ the hardware. The following example keeps those phases visible in one place:
 ``Robot``. Construction records hardware arguments and placement, but it does
 not import a vendor SDK or open a device. ``describe()`` reads that declaration,
 so it can report paths, nodes, and connection ownership before the hardware is
-powered or reachable.
+powered or reachable. An environment starts from the scheduler's hardware config
+instead of builder arguments, and calls ``from_config()`` on the registered
+robot class. It forwards each config field to ``build()`` and places parts on
+the environment's node unless the config names another.
 
 Setup code then selects the capabilities it needs. ``child("arm", Arm)`` returns
 the part at ``arm`` and checks that it implements ``Arm``; the expected class is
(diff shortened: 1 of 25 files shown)


andylin-hao:
Closing: this will land as part of one PR covering every phase of the real-world task decoupling.

PR #867, opened 2026-03-13
feat: add embodied reward worker and ResNet model

guozhen1997:
### Description

1. Add embodied reward worker
2. Add reward inference in rollout
3. Support Resnet as reward model
4. Support Resnet reward training

### How has this been tested?

ResNet reward training
image

PPO training with trained reward model inference
image

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update (Document-only update)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1563, opened 2026-09-11
fix(config): enforce group_size>1 for the ReinForce++ baseline

kakiuwang-ui:
### Description

Fix the group-baseline guard in `validate_cfg` so it actually fires for the ReinForce++ baseline.

- Replace the dead `adv_type == "reinpp_baseline"` check (a value no code ever produces) with a helper `adv_requires_group_baseline(adv_type, use_reinpp_baseline)`.
- The helper returns `True` for `grpo`, `grpo_dynamic`, and `reinpp` **when** `use_reinpp_baseline` is set, and lower-cases `adv_type` to match `get_adv_and_returns`' case-insensitive dispatch.
- Give the `group_size > 1` assertion an explanatory message.

### Motivation and Context

`validate_cfg` guards group-based advantage estimators with `assert group_size > 1`, but it tested `cfg.algorithm.adv_type in ("grpo", "grpo_dynamic", "reinpp_baseline")`. `reinpp_baseline` is never a produced `adv_type`: ReinForce++ registers as `reinpp` (`rlinf/algorithms/advantages.py:302`) and selects its baseline through the separate boolean `use_reinpp_baseline`. The guard therefore never fired for the ReinForce++ baseline.

The consequence is a silent failure, not an error. With `adv_type: reinpp`, `use_reinpp_baseline: true`, `group_size: 1`, config validation passes and `compute_reinpp_advantages` runs `grouped_rewards -= grouped_rewards.mean(dim=1)` over a group of one — subtracting each reward from itself and zeroing every advantage, so the policy trains on no signal. GRPO with `group_size: 1` is rejected up front; the ReinForce++ baseline should be too.

A second, related gap surfaced in review: `get_adv_and_returns` dispatches on `adv_type.lower()`, so a differently-cased name (`GRPO`) would skip the guard yet still reach the group-based estimator. The helper now normalizes case as well.

The dead string dates back to #867; `reinpp` and `use_reinpp_baseline` were added later, so the two drifted apart. No shipped config sets `use_reinpp_baseline: true`, so no existing run changes behavior — this only rejects a misconfiguration that used to fail silently.

### How has this been tested?

New unit test `tests/unit_tests/test_config_adv_group_baseline.py` (9 cases) pins the predicate: grpo/grpo_dynamic always require a group, reinpp requires one only in baseline mode, plain reinpp/gae/raw/opd are exempt, and the check is case-insensitive.

```
$ PYTHONPATH=$PWD python -m pytest tests/unit_tests/test_config_adv_group_baseline.py -q
9 passed
$ ruff check rlinf/config.py tests/unit_tests/test_config_adv_group_baseline.py
All checks passed!
```

Both fixes were mutation-tested: reverting the reinpp-baseline branch fails `test_reinpp_requires_a_group_only_in_baseline_mode`, and removing the `.lower()` fails `test_guard_matches_case_insensitive_dispatch`.

`grpo_video` with `advantage_mode: frame` and `group_size: 1` has the same singleton-group problem (its `std(dim=-1)` over one element is NaN), but whether it needs a group depends on `advantage_mode`, so it is intentionally left out of this guard and noted for a separate change.

### Types of changes

### Checklist:

### DIFF
--- a/rlinf/config.py
+++ b/rlinf/config.py
@@ -1518,6 +1518,29 @@ def validate_coding_online_rl_cfg(cfg: DictConfig) -> DictConfig:
     return cfg
 
 
+def adv_requires_group_baseline(
+    adv_type: str, use_reinpp_baseline: bool = False
+) -> bool:
+    """Whether an advantage estimator draws its baseline from within-group
+    statistics and therefore requires ``algorithm.group_size > 1``.
+
+    GRPO and GRPO-dynamic always normalize each reward against its group, so a
+    group of one leaves nothing to compare against. ReinForce++ only subtracts a
+    per-group mean when its baseline mode is enabled (``use_reinpp_baseline``);
+    plain ReinForce++ normalizes over the whole batch and is exempt.
+
+    ``adv_type`` is lower-cased to match how :func:`get_adv_and_returns`
+    dispatches, so a differently-cased name cannot slip past the guard and still
+    reach the group-based estimator.
+    """
+    adv_type = adv_type.lower()
+    if adv_type in ("grpo", "grpo_dynamic"):
+        return True
+    if adv_type == "reinpp":
+        return bool(use_reinpp_baseline)
+    return False
+
+
 def validate_cfg(cfg: DictConfig) -> DictConfig:
     OmegaConf.set_struct(cfg, True)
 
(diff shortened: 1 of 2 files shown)


tests/unit_tests/test_config_adv_group_baseline.py
```diff
@@ -0,0 +1,57 @@
+# Copyright 2025 The RLinf Authors.
```
Lin-xs: Please delete this unit test.

rlinf/config.py
```diff
@@ -1518,6 +1518,29 @@ def validate_coding_online_rl_cfg(cfg: DictConfig) -> DictConfig:
...
+    GRPO and GRPO-dynamic always normalize each reward against its group, so a
+    group of one leaves nothing to compare against. ReinForce++ only subtracts a
+    per-group mean when its baseline mode is enabled (``use_reinpp_baseline``);
+    plain ReinForce++ normalizes over the whole batch and is exempt.
+
+    ``adv_type`` is lower-cased to match how :func:`get_adv_and_returns`
+    dispatches, so a differently-cased name cannot slip past the guard and still
+    reach the group-based estimator.
+    """
+    adv_type = adv_type.lower()
```
Lin-xs, line 1539: Some configs set `adv_type` to `None` (e.g. `examples/embodiment/config/d4rl_iql_mujoco.yaml`). Could you add an early guard at the beginning of this function to handle that case?

tests/unit_tests/test_config_adv_group_baseline.py
kakiuwang-ui: Deleted in 479080d.

rlinf/config.py
kakiuwang-ui: Confirmed — thanks. Four configs set `adv_type: null` and all run with `task_type: offline` and `only_eval: false`, so they reach the guard and raise `AttributeError` on the `.lower()` call: `examples/embodiment/config/d4rl_iql_{mujoco,antmaze,kitchen_adroit}.yaml` and `tests/e2e_tests/embodied/d4rl_iql_mujoco.yaml`.

Added the early guard you suggested — an unset `adv_type` selects no estimator, so no group is required. Widened the annotation to `Optional[str]` and noted the case in the docstring. Verified all four configs now return `False` where they previously raised. 479080d

Lin-xs (approved):

andylin-hao (approved):

